PR #342, opened 2014-04-29
Added support for overriding file browse dialogs by implementing the cef...

PlehXP:
Example usage...

```
public class SilentDialogHandler : IDialogHandler
{
    public static string FilePath;
    public bool OnOpenFile(IWebBrowser browser, string title, string default_file_name, List<string> accept_types, out List<string> result)
    {
        result = new List<string>() { FilePath };
        return true;
    }
}
```

Then make sure you set it on your view

```
webView.DialogHandler = new SilentDialogHandler();
```

then before you click a browse button, make sure you set the static FilePath field on the SilentDialogHandler class.

### DIFF
--- a/CefSharp.Core/Internals/ClientAdapter.cpp
+++ b/CefSharp.Core/Internals/ClientAdapter.cpp
@@ -408,5 +408,31 @@ namespace CefSharp
             // Unknown dialog type, so we return "not handled".
             return false;
         }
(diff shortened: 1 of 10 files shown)


ukushu:
Can someone give me the sample code of usage?
I cannot understand how to use the feature. =(

PR #380, opened 2014-06-02
Implement CefDialogHandler based on #342

amaitland:
PR based on #342 apply polish before final merge review.

### DIFF
--- a/CefSharp/IDialogHandler.cs
+++ b/CefSharp/IDialogHandler.cs
@@ -1,8 +1,31 @@
 ﻿using System.Collections.Generic;
 namespace CefSharp
 {
+    /// <summary>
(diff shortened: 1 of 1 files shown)


ukushu:
Looks like this feature was deleted, as at the moment method parameters is:

`public bool OnFileDialog(IWebBrowser browserControl, IBrowser browser, CefFileDialogMode mode, string title, string defaultFilePath, List<string> acceptFilters, int selectedAcceptFilter, IFileDialogCallback callback)`

But I need to have a way to select some file without display of OpenFileDialog. How can I realize that?

jankurianski:
@ukushu please ask questions on https://gitter.im/cefsharp/CefSharp

PR #422, opened 2014-07-15
Modified project and property files to support changing the target platform toolset

taylorjonl:
Changes were made to the project/property files to support building in VS2013 but targeting another version of the VC runtime, e.g. VS2010.  This is controlled through the project properties GUI in Visual Studio for the CefSharp.Core project.

platform-toolset

If no platform toolset is set it will use the version that matches the current version of Visual Studio.


amaitland:
This doesn't appear to play nice with `VS2012`, haven't had a chance to try it in `VS2013`.

It keeps wanting to default to `v100`

Also doesn't support the `BrowserSubProcess` (which is easy enough to fix).

@taylorjonl Any chance you could provide some insight on this?

amaitland:
I couldn't make this work properly in `VS2012`. If we can get this working in `VS2012` and `VS2013` then more than happy to merge.

Closing for now.

PR #619, opened 2014-11-18
WPF DPI auto zoom (again)

bjarteskogoy:
Messed up the other branch. Have to learn more Git ...
Fixed some issues, but haven't made it an attached property yet.


bjarteskogoy:
The code in this pull request does what it's supposed to, but it suffers under the issue that ZoomLevel isn't applied correctly in WPF. Easy to reproduce in the WPF control by zooming in (right click menu) -> nothing happens, reload the page and ZoomLevel is applied.

amaitland:
I've added a `CalculateZoomLevel` method and exposed `SetZoomLevel`, so anyone can implement zooming in their own code.

As stated I'd prefer to go with an `AttachedProperty` for this.

bjarteskogoy:
Good. Anyway, it looks like there is a bug in cef making it impossible to be sure you have the right zoom on the first page load: https://code.google.com/p/chromiumembedded/issues/detail?id=1467&sort=-id

jornh:
Thanks @bjarteskogoy added to #738 ... seems there are several `upstream` issues registered for it.

xperiandri:
So what about zoom on high DPI? Does it work?

amaitland:
`High DPI` under `WPF` in general needs quite a bit of work, there is an outstanding PR, see #937 which is basically stalled in it's progress. 

If it's a feature your interested in and have some time then feel free to get involved.

PR #679, opened 2014-12-11
Using Cef Native AutoLock

amaitland:
Remove AutoLock.h and revert to using Cef Native AutoLock. This change was made previously and had problems compiling under x64, I believe that has been resolved now.


amaitland:
One day I might get around to looking into this further. For now I'm going to close and we have a working lock implementation.

PR #702, opened 2014-12-19
Fix tabbing in and out of the browser (windows forms)

sylvain-hamel:
## Tl;dr; 
- Tabbing in and out of the browser control does not work
- Browser steals focus upon loading a page
- Entering or leaving the browser (using the mouse) does not deactivate/activate previous and next control correctly which cause them miss important focus/activation related events.
- This PR fixes all this

### Test setup:
- Create a form that contains (in this order, with this Tab order) a textbox, a go button, a browser and another textbox.
- In the go button handler, load a URL
- Handle textbox1_Enter and textbox1_Leave to change the textbox's back color to yellow and back to white

### Test case 1
- Start the app
  - *Result: Textbox1 has the focus and is yellow*
- Click go button
  - *Result: Browser loads, focus goes to button 1 and textbox1 becomes white*
- Press tab
  - *Result: focus enters the browser at its current element (based on 1st hyperlink in the page or wherever it was set by javascript)*
- Press tab a few times to focus each focusable element in the page until you reach the last one
- Press tab again
  - *Result: focus should leave the browser and go to textbox2*

### Test case 2
- Start the app
- Click in the browser
- Activate another application
- Click inside textbox1
  - *Result: textbox1 is yellow*

## Issues:
The above test cases don't work at all. Focus is stolen upon loading of a page, we are unable to get into and get out of the browser using the Tab key. Sometimes the textbox1 stays white even if it has the focus because the browser corrupted the state of which control is the active control.

## Fix: 
This change has two parts:
- Remove `IFocusHandler` and the related code. 
    - `IFocusHandler` cannot be used as is to handle focus related events due to the limitations of how cef behaves. This issue has to be resolved at the core of `CefSharp.WinForms.ChromiumWebBrowser`. 
    - Once this issue is resolved in the core, `IFocusHandler` no longer provides any useful extension point. Host applications are now able to use the built-in, and well known Windows Forms mechanisms such as the `Focused` event and the `Focus()` method to interact with the browser control.
- Fix the bugs above-mentioned.

## How to test the changes:
- We added the `TabulationDemoFrom` class to tests the above test cases. This form is a bit more complex than the form that is described in the context section above because it uses nested `IContainerControls` (the form itself and a `UserControl`) to make sure the code to activate the next control work well in this context.
- Edit `CefSharp.WinForms.Example\Program.cs` and replace `new BrowserForm` by `new TabulationDemoForm`
- Run
- Test tabulation

## Limitations:
We did not change anything in the context of WPF. The WPF side should behave exactly as it used to (I don't know if there are focus or tabulation related issues in WPF). A ` TabulationDemoFrom` sample in WPF could be written to see if the WPF side has the same issues and if they should be fixed of not.


amaitland:
Closing in favor of #774 

If there's still outstanding issues then more than happy to discuss further :+1:

PR #707, opened 2014-12-22
2171 Cleanup/Improvements

amaitland:
- Refactor/Simplify `OSR` Rendering (hopefully one day it can be abstracted out into a pluggable implementation)
- Improved comments so code can be better understood.
- Cleanup of `BrowserSettings`, just create a new instance and use it (also call dispose as it's an unmanaged wrapper).

Driving force behind this PR is the OSR rendering was overly complex and difficult to understand (it's actually fairly simple what's going on, the code was just all over the place)


amaitland:
I'm fairly happy with the `OSR` rendering rewrite and other code cleanups.

At some later point I'll look at abstracting the rendering into a pluggable implementation.

In the interest of progress merging this now (best to get major changes integrated early in the release cycle).

PR #710, opened 2014-12-23
Enhancement - FocusHandler Take3

amaitland:
Again based on #702 

- based on `master` branch
- Re-Implemented `IFocusHandler` allowing user to override if required
- Get BrowerHandle from Cef
- Code cleanup


sylvain-hamel:
Hi @amaitland, can you please give me an update on the status of this PR and #702?

amaitland:
@sylvain-hamel Have you had a chance to take this PR for a spin? Your test cases appear to be working for me. If you need to make any further changes feel free to merge my commits into #702 and make changes there.

Secondly I think it's **VERY** important that we add adequate documentation around this as we are working with an undocumented feature which may well change in the future. The inner `Window` is `Chrome_RenderWidgetHostHWND`, which provides a bit of information. See https://groups.google.com/a/chromium.org/forum/#!topic/chromium-dev/hmBh5YHjOFY

Long term I think we should create an issue and see about getting `OnGotFocus` fixed in upstream `CEF`

sylvain-hamel:
@amaitland 

> Have you had a chance to take this PR for a spin? Your test cases appear to be working for me.

No not yet. I was waiting for the issues raised during the code review to be fixed. I looks like you did not put back the `childWindows.Last();` *hack* in your branch yet. The test case provided by @Termit1975 requires that. Can you add it back and then I can re-test the whole thing based off your branch?

> Secondly I think it's VERY important that we add adequate documentation...

I agree. What would you have in mind exactly? We can possibly find the window by its name. That would make the code more understandable because the window name is meaningful. Something like:

```c#
    // The window that gets the focus is a child window of the browser and is called
    // Chrome_RenderWidgetHostHWND, see See https://groups.google.com/a/chromium.org/forum/#!topic/chromium-dev/hmBh5YHjOFY
    var cefWindowHandle = SomeHelper.GetSubWindowsByName(
        webBrowser.Handle, "Chrome_RenderWidgetHostHWND");      
```

amaitland:
> No not yet. I was waiting for the issues raised during the code review to be fixed. I looks like you did not put back the childWindows.Last(); hack in your branch yet.

I've tried on two different machines and it appears to be working properly with obtaining the first child of the browser handle. 
https://github.com/amaitland/CefSharp/blob/enhancement/focus-handler3/CefSharp.Core/ManagedCefBrowserAdapter.h#L102

> The test case provided by @Termit1975 requires that. Can you add it back and then I can re-test the whole thing based off your branch?

I don't really have any more time to commit to this, new project starting at work. Greatly appreciated if you could finish this off.

> That would make the code more understandable because the window name is meaningful. Something like:

Sounds reasonable :+1: I'd prefer to keep this sort of code in the `c++` side of things rather than have `DllImports`.

amaitland:
Closing in favor of #774

PR #730, opened 2015-01-06
change c++ file copyright symbol (©) encoding to work on non-english system

binsys:
change c++ file copyright symbol (©) encoding from 0xa9(cp1252) to 0xc2 0xa9(cp65001 without bom) ,

now work on non-english system.


amaitland:
Hi @binsys Thanks for the contribution :+1:

A few comments:

- Description says this PR changes the copyright, no mention about plugins and I can see there are **two commits**. If this is intentional, please split that out into a separate PR, speeds up the review process.
- If we're changing the `c++` file encodings then I believe we should standardize on the same encoding used by the `c#` files. So basically changing them to `65001` (UTF-8 with signature).

PR #731, opened 2015-01-07
change c++ file copyright symbol (©) encoding to work on non-english system

binsys:
change c++ file copyright symbol (©) encoding from 0xa9(cp1252) to 0xc2 0xa9(cp65001 with bom) ,
now work on non-english system.
Note:* .cs files encoded correct, do not need changes.

### DIFF
--- a/CefSharp.BrowserSubprocess.Core/CefAppUnmanagedWrapper.cpp
+++ b/CefSharp.BrowserSubprocess.Core/CefAppUnmanagedWrapper.cpp
@@ -1,4 +1,4 @@
-// Copyright Â© 2010-2014 The CefSharp Project. All rights reserved.
+// Copyright © 2010-2014 The CefSharp Project. All rights reserved.
 //
 // Use of this source code is governed by a BSD-style license that can be found in the LICENSE file.
 #pragma once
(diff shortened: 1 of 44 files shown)


amaitland:
Should they be with or without BOM? I think the default for `c#` is with signature.

binsys:
@amaitland Yes, you are right, I missing the BOM.sorry.

amaitland:
Thanks :+1: I'll try to merge this next week

amaitland:
@binsys Merging this now. I've looked at a random sample and everything looks ok :+1: 

It maybe a little bit of an uphill battle as it's likely new files added will be `cp1252` encoded, guess we'll see how we go.

jornh:
> It maybe a little bit of an uphill battle as it's likely new files added will be `cp1252` encoded, guess we'll see how we go.

Maybe it can be added as some kind of a linting step to the build, together with spaces vs tabs? Let's see how much of a stumbling block it will be?

amaitland:
> Maybe it can be added as some kind of a linting step to the build, together with spaces vs tabs? Let's see how much of a stumbling block it will be?

Some sort of pre-commit hook that does some housekeeping would be handy :+1:

binsys:
@amaitland  
I recommend replacing © to (C) or Copyright , remove any special encoding characters.

amaitland:
@binsys How much of a problem does the `Copyright` symbol cause when using a `non-english` system?

Also were you looking to add the plugin functions?

Issue #171, opened 2013-09-16
Drag and drop in WPF


mwisnicki:
Right now it's not supported.

See upstream [bug](http://code.google.com/p/chromiumembedded/issues/detail?id=1032) - it also applies to CEF1.

Bodekaer:
Did anyone look into this?
We'll need this for an upcoming release this month, so I could give it a go if no one is already on it.

amaitland:
@Bodekaer Go for it :+1:

Bodekaer:
Done :)

jornh:
Awesome-sovs! (Whish I had time to take it for a spin ... but sadly not today :cry:)

Bodekaer:
Here :) It works really nicely. And opens up so many new possibilities here, which is amazing.

@myrnazhou, FYI we have drag drop now. Might be useful for native image drag drop at some point (low priority though).

amaitland:
Drag and Drop support has been implemented with #756, it's not 100% feature complete. I'm happy that it covers your typical usage scenario so I'm going to close this.

Great work @Bodekaer :+1:

Bodekaer:
Thanks a bunch. Happy to help.

Yeah, the primary thing that is missing is the ability to drag e.g. image/text *out of* the browser.
You can drag anything into the browser atm. E.g. dropping files/images *into* the browser.

If anyone wants to implement the former, please just let me know and I can help direct to the relevant code and documentation files.

ralphmayr:
Hi @Bodekaer! Could you provide some pointers to what would be necessary to get drag and drop *out of* and *within* the browser working? (So that for example [http://html5demos.com/drag](http://html5demos.com/drag) would work in CefShrp.Wpf)
Thanks!

Bodekaer:
Hey @ralphmayr,
Yes definitely. Would very much like to see this added as well.
I'd recommend checking out the "drag" related issues in CEF3, and then find the relevant revision which adds support for this:
https://code.google.com/p/chromiumembedded/issues/list?can=1&q=drag&colspec=ID+Type+Status+Priority+Milestone+Owner+Summary&cells=tiles

E.g. this issue might provide the relevant input (although related to CEF1 I believe):
https://code.google.com/p/chromiumembedded/issues/detail?id=140&can=1&q=drag

With that, you'll probably see that we need to implement a few C++ methods to support the CEF3 API for dragging within/out of the browser. 
Once you find out what/where to add this, I can help with further directions if needed. 
[...]

PR #732, opened 2015-01-07
Added Drag Drop support to allow dragging objects into the browser

Bodekaer:
Supports files, links, text, html etc.
Currently only supported in WPF (would be easy to add WinForm support though, by just copy-pasting the WPF ChromiumBrowser code)

FYI #171

### DIFF
--- a/CefSharp.Core/Internals/CefDragDataWrapper.h
+++ b/CefSharp.Core/Internals/CefDragDataWrapper.h
@@ -41,6 +41,24 @@ namespace CefSharp
             virtual property bool IsFragment;
             virtual property bool IsLink;
 
+            ///
+            // Create a new CefDragData object.
+            ///
+            /*--cef()--*/
+            static CefDragDataWrapper^ Create(){
+                CefRefPtr<CefDragData> cefDragData = CefDragData::Create();
+                return gcnew CefDragDataWrapper(cefDragData);
+            }
+
+
+            virtual property CefRefPtr<CefDragData>* _internalDragData
+            {
+                CefRefPtr<CefDragData>* get() 
+                { 
+                    return new CefRefPtr<CefDragData>(_wrappedDragData.get()); 
+                }
+            }
+
             //TODO: Vector is a pointer, so can potentially be updated (items may be possibly removed)
             virtual property IList<String^>^ FileNames
             {
--- a/CefSharp.Wpf/ChromiumWebBrowser.cs
+++ b/CefSharp.Wpf/ChromiumWebBrowser.cs
@@ -7,7 +7,9 @@ using CefSharp.Internals;
 using Microsoft.Win32.SafeHandles;
 using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Threading.Tasks;
+using System.Linq;
 using System.Windows;
 using System.Windows.Controls;
 using System.Windows.Controls.Primitives;
@@ -422,6 +424,12 @@ namespace CefSharp.Wpf
             GotKeyboardFocus += OnGotKeyboardFocus;
             LostKeyboardFocus += OnLostKeyboardFocus;
 
+            // Drag Drop events
+            DragEnter += OnDragEnter;
+            DragOver += OnDragOver;
+            DragLeave += OnDragLeave;
+            Drop += OnDrop;
+
             IsVisibleChanged += OnIsVisibleChanged;
 
             ToolTip = toolTip = new ToolTip();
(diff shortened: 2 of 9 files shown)


CefSharp/CefDragOperationsMask.cs
```diff
@@ -0,0 +1,24 @@
...
+
+using System;
+
+namespace CefSharp
+{
+    /// <summary>
+    /// Drag operation mask types
+    /// </summary>
+    [Flags]
+    public enum CefDragOperationsMask
```
amaitland: There's already a `CefDragOperationsMask` enum see https://github.com/cefsharp/CefSharp/blob/cef/2171/CefSharp/DragOperationsMask.cs

Just a note most enums in `Cef` are `uint`, whilst you can cast between the two, the value for `Every` is different being `uint.MaxValue`

CefSharp/MouseEvent.cs
```diff
@@ -0,0 +1,17 @@
...
+//
+// Use of this source code is governed by a BSD-style license that can be found in the LICENSE file.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+
+namespace CefSharp
+{
+    public class MouseEvent : IMouseEvent
```
amaitland: Do we need the `IMouseEvent` interface? Can probably make this a struct as well (There's a few other classes in the code base that probably should be structs, one day I'll get around to changing them)

CefSharp.Wpf/ChromiumWebBrowser.cs
```diff
@@ -459,6 +467,151 @@ public ChromiumWebBrowser()
             PresentationSource.AddSourceChangedHandler(this, PresentationSourceChangedHandler);
         }
 
+        void OnDrop(object sender, DragEventArgs e)
```
amaitland: Can you please add the `private` keyword to these methods for consistency :+1:

CefSharp.Wpf/ChromiumWebBrowser.cs
```diff
@@ -422,6 +424,12 @@ public ChromiumWebBrowser()
             GotKeyboardFocus += OnGotKeyboardFocus;
             LostKeyboardFocus += OnLostKeyboardFocus;
 
+            // Drag Drop events
+            DragEnter += OnDragEnter;
```
amaitland: I see these event handlers being `hooked`, are they `unhooked` in dispose?

CefSharp.Core/Internals/CefDragDataWrapper.h
```diff
@@ -41,6 +41,24 @@ namespace CefSharp
...
+            // Create a new CefDragData object.
+            ///
+            /*--cef()--*/
+            static CefDragDataWrapper^ Create(){
+                CefRefPtr<CefDragData> cefDragData = CefDragData::Create();
+                return gcnew CefDragDataWrapper(cefDragData);
+            }
+
+
+            virtual property CefRefPtr<CefDragData>* _internalDragData
```
amaitland: Not really a fan of naming a property starting with a `_`, calling it `InternalDragData` should be fine.

amaitland:
@Bodekaer Looks pretty solid overall :+1: I've commented inline where appropriate

In general there's some cleanup that still needs to happen
- Opening braces (`{`) on new lines
- Cleanup whitespace

> Currently only supported in WPF (would be easy to add WinForm support though, by just copy-pasting the WPF ChromiumBrowser code)

FYI `WinForms` version already has drag and drop built in, I'm pretty sure those methods are only required for `OSR`

CefSharp/MouseEvent.cs
amaitland: Don't worry about the `struct` idea, we'll just keep things consistent for now.

Bodekaer: Thanks for all the great feedback Alex!
I'll run through it asap and update the PR.

On 1/8/2015 10:39:25 AM, Alex Maitland <notifications@github.com> wrote:
In CefSharp/MouseEvent.cs [https://github.com/cefsharp/CefSharp/pull/732#discussion-diff-22642543]:
> @@ -0,0 +1,17 @@ > +﻿// Copyright © 2010-2014 The CefSharp Authors. All rights reserved. > +// > +// Use of this source code is governed by a BSD-style license that can be found in the LICENSE file. > + > +using System; > +using System.Collections.Generic; > +using System.IO; > + > +namespace CefSharp > +{ > + public class MouseEvent : IMouseEvent
Don't worry about the struct idea, we'll just keep things consistent for now.

amaitland:
@Bodekaer When you have a sec can you retarget this against `master`? (I think you should just be able to edit the target branch above). I've merged `cef/2171` into `master` :smile:

Bodekaer:
Yep definitely. I just created a new `PR` against `master` here #756.
I cannot edit the comparison of this PR, so we can just close it.

Just FYI, I've been hyper busy lately, so I haven't yet cleaned up the code based on your feedback.

amaitland:
@Bodekaer Thanks!

PR #759, opened 2015-01-22
Fix OffScreen's Address field

jankurianski:
This fix ensures the Address field returns the correct value after construction (instead of null).  The Address field setter is now private as callers should use Load() to change url.

```C#
var browser = new CefSharp.OffScreen.ChromiumWebBrowser("https://www.google.com/");
var a = browser.Address; // now returns "https://www.google.com/" instead of null
browser.Address = "about:blank"; // now generates compiler error instead of doing nothing
```


amaitland:
:+1:

PR #744, opened 2015-01-15
Fixed #743: Forward state of Ctrl, Alt and Shift keys 

ralphmayr:
Fixed #743: Forward state of Ctrl, Alt and Shift keys to the managedCefBrowserAdapter

PR #782, opened 2015-02-02
Fixed #743: Forward state of Ctrl, Alt and Shift keys

### LINKED ISSUE #743: Ctrl+Shift does not work in CefSharp WPF
The state of the Ctrl (and Alt and Shift) keys is not forwarded to the web app running in CefSharp in WPF. (Basically the same problem as we had it in #542 with mouse events)

ralphmayr:
Fix for #743, Re-targeted the PR #744 to the current head.


CefSharp.Wpf/ChromiumWebBrowser.cs
```diff
@@ -626,8 +626,13 @@ private IntPtr SourceHook(IntPtr hWnd, int message, IntPtr wParam, IntPtr lParam
                         return IntPtr.Zero;
                     }
 
-                    handled = managedCefBrowserAdapter.SendKeyEvent(message, wParam.ToInt32(), lParam);
-                    
+                    const CefEventFlags modifiers = 0;
+
+                    if (managedCefBrowserAdapter.SendKeyEvent(message, wParam.ToInt32(), (IntPtr) AddKeyboardModifiers(modifiers)))
```
amaitland: As `lparam` is representing the native key code I think this is likely to break the keyboard handling.

https://github.com/cefsharp/CefSharp/blob/master/CefSharp.Core/ManagedCefBrowserAdapter.h#L191

Is the Modifier required in the `SourceHook`? Thought it would have been enough to just pass it though on mouse click?

Issue #785, opened 2015-02-04
Creating Extra Tab, Waiting, Closing Tab in Winforms.Example throws null-deref exception


rassilon:
NOTE: I reproduced this with CefSharp/master and NuGet pulling down whatever it's currently configured to pull down for cef.dll purposes.

The underlying cause is that WinForm's ChromiumWebBrowser::Dispose(bool disposing) method wasn't conditionally disposing BrowserSettings and BrowserTabUserControl in WinForms.Example was disposing the ChromiumWebBrowser twice. 

Once via default WinForm control disposing because the browser was reachable via this.browserPanel.Controls. 

The other was calling the Dispose() method specifically in the ChromiumWebBrowser Dispose event listening function: BrowserTabUserControlDisposed.

This is of course fairly straight forward to fix, but when I went to do so, I noticed that bunch of IDisposable pattern fixes needed to occur as well. PR to follow.

FYI,
Bill

amaitland:
> PR to follow.

Great!

Please make sure you read `Contributing.md` in particular https://github.com/cefsharp/CefSharp/blob/master/CONTRIBUTING.md#coding-style

amaitland:
Closing with merge of #786

PR #786, opened 2015-02-04
Cleanup/Implement Dispose Pattern

rassilon:
See #785 for details, but this fixes the obvious issue in #785 as well as some other problematic looking IDisposable patterns in the other ChromiumWebBrowser flavors.

### DIFF
--- a/CefSharp.OffScreen/ChromiumWebBrowser.cs
+++ b/CefSharp.OffScreen/ChromiumWebBrowser.cs
@@ -58,26 +58,60 @@ namespace CefSharp.OffScreen
             managedCefBrowserAdapter.CreateOffscreenBrowser(IntPtr.Zero, browserSettings ?? new BrowserSettings(), address);
         }
 
+        ~ChromiumWebBrowser()
+        {
+            Dispose(false);
+        }
+
         public void Dispose()
         {
+            Dispose(true);
+            GC.SuppressFinalize(this);
+        }
+
+        protected virtual void Dispose(bool disposing)
+        {
+            // Don't reference handlers any longer:
             ResourceHandler = null;
+            JsDialogHandler = null;
+            DialogHandler = null;
+            DownloadHandler = null;
+            KeyboardHandler = null;
+            LifeSpanHandler = null;
+            MenuHandler = null;
+            FocusHandler = null;
+            RequestHandler = null;
+            DragHandler = null;
+            GeolocationHandler = null;
+
+            // Don't reference event listenets any longer:
+            LoadError = null;
+            FrameLoadStart = null;
+            FrameLoadEnd = null;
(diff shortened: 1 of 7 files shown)


amaitland:
> I'm not sure if this is necessary or not in managed C++, but it certainly looked weird for both the Dipose and finalization method to both try and delete _browserSettings.

Should be perfectly safe to delete a `NULL` object. 

https://voidnish.wordpress.com/2006/06/07/you-can-delete-a-null-pointer/

CefSharp.OffScreen/ChromiumWebBrowser.cs
```diff
@@ -58,26 +58,60 @@ public ChromiumWebBrowser(string address, BrowserSettings browserSettings = null
...
+            DownloadHandler = null;
+            KeyboardHandler = null;
+            LifeSpanHandler = null;
+            MenuHandler = null;
+            FocusHandler = null;
+            RequestHandler = null;
+            DragHandler = null;
+            GeolocationHandler = null;
+
+            // Don't reference event listenets any longer:
```
jankurianski: listenets -> listeners

rassilon:
But _browsersSettings isn't set to NULL. Thus my paranoia.

amaitland:
> But _browsersSettings isn't set to NULL. Thus my paranoia.

Usually we set unmanaged objects to `NULL` and looking at browser settings closer I see that it's a `struct`, so I'm pretty sure it would be safe to call `delete`. Happy for the extra checking though :+1:

amaitland:
Overall, looks like a great contribution thanks! :+1: 

(The `BrowserSettings` dispose was actually a chance I made just recently, was done in a hurry)

CefSharp.Wpf/ChromiumWebBrowser.cs
```diff
@@ -606,9 +627,6 @@ private string ReadUrlFromDragDropData(IDataObject data, string urlDataFormatNam
             return url.TrimEnd('\0');
         }
 
-
```
amaitland: Stray whitespace that can ideally be removed.

rassilon: Are you suggesting I put the white space back? (I did remove that extraneous white space, yes.)

amaitland: Ignore that, was on the phone, so clearly wasn't paying enough attention! lol

amaitland:
FYI I've updated the title to make it a little more descriptive.

For reference you can include the issue number in one of your commit message and the issue will auto close on merge.
https://help.github.com/articles/closing-issues-via-commit-messages/

amaitland:
I've taken this for a spin and looks solid :+1:

Issue #752, opened 2015-01-18
Add Readme.txt to Nuget Packages


amaitland:
Add a `Readme.txt` file to the `nuget` packages as per http://docs.nuget.org/docs/creating-packages/creating-and-publishing-a-package

Include 
- Basic troubleshooting info
- The all important `AnyCpu` info
- The close Visual Studio after install if you don't see the references trick
- Links to FAQ, Troubleshooting Guide, Google Group, Github page
- Excerpt from Contributing.md around creating an issue

amaitland:
Anyone else have anything they'd like to add/remove?

amaitland:
Added the basic `Readme.txt` with commit https://github.com/cefsharp/CefSharp/commit/fb3415b5bd29b5198dba8db70d69c11a2fb3b034

Only problem is that it currently only shows the readme when installing the package clean, not when upgrading.

amaitland:
Additions to the readme are most welcome :+1:

amaitland:
Alternatively should we open a Url?

jankurianski:
> Alternatively should we open a Url?

I like your idea here, because a web page is more likely to be read and kept open as a reference while users take their first steps (unlike readme.txt which goes away after installation).  But how would it work?  Can you open a web browser?

amaitland:
I haven't tried it yet, apparently you can use powershell via an install script that runs every time. Looks easy enough to implement. Unsure if it opens a new browser instance, or just a browser inside of `VS`

http://stackoverflow.com/questions/15401521/how-to-open-readme-url-for-nuget-package

rassilon:
The SO answer that says to use $DTE...Navigate would be the VS browser. The "start url" suggestion would launch the default browser outside of VS.

Bill

amaitland:
I think for the `-pre` I'll just keep the `Readme.txt` file. Long term I'd like to change over.

Was more looking for any additional comments on the content https://github.com/cefsharp/CefSharp/blob/master/NuGet/Readme.txt

PR #787, opened 2015-02-04
#672: Notify CEF when our owning form moves (WM_MOVE or WM_MOVING) just ...

rassilon:
WM_MOVE and WM_MOVING only get sent to parent windows, so add a MovingListener that attaches to the control's parent form and sends the notifications to CEF.

Fyi,
Bill

### DIFF
--- a/CefSharp.WinForms/ChromiumWebBrowser.cs
+++ b/CefSharp.WinForms/ChromiumWebBrowser.cs
@@ -114,7 +114,7 @@ namespace CefSharp.WinForms
             {
                 this.BeginInvoke((MethodInvoker)delegate 
                 {
-                    movingListener = new MovingListener(this.FindForm());
+                    movingListener = new MovingListener(this);
                     movingListener.Moving += movingListener_Moving;
                 });
             }
--- a/CefSharp.WinForms/ChromiumWebBrowser.cs
+++ b/CefSharp.WinForms/ChromiumWebBrowser.cs
@@ -14,7 +14,7 @@ namespace CefSharp.WinForms
     public class ChromiumWebBrowser : Control, IWebBrowserInternal, IWinFormsWebBrowser
     {
         private ManagedCefBrowserAdapter managedCefBrowserAdapter;
-        private MovingListener movingListener;
+        private ParentFormMessageInterceptor parentFormMessageInterceptor;
 
         public BrowserSettings BrowserSettings { get; set; }
         public string Title { get; set; }
@@ -107,13 +107,13 @@ namespace CefSharp.WinForms
             IsBrowserInitialized = true;
 
             // By the time this callback gets called, this control
-            // is most likely hooked into a parent Form of some sort. 
-            // (Which is what MovingListener relies on.)
-            // Ensure the MovingListener construction occurs on the WinForms UI thread:
+            // is most likely hooked into a browser Form of some sort. 
+            // (Which is what ParentFormMessageInterceptor relies on.)
+            // Ensure the ParentFormMessageInterceptor construction occurs on the WinForms UI thread:
             this.InvokeOnUiThreadIfRequired(() =>
             {
-                movingListener = new MovingListener(this);
-                movingListener.Moving += MovingListenerMoving;
+                parentFormMessageInterceptor = new ParentFormMessageInterceptor(this);
+                parentFormMessageInterceptor.Moving += MovingListenerMoving;
             });
 
             ResizeBrowser();
--- a/CefSharp.WinForms/ChromiumWebBrowser.cs
+++ b/CefSharp.WinForms/ChromiumWebBrowser.cs
@@ -94,7 +94,7 @@ namespace CefSharp.WinForms
             base.Dispose(disposing);
         }
 
-        void movingListener_Moving(object sender, EventArgs e)
+        private void MovingListenerMoving(object sender, EventArgs e)
         {
             if (IsBrowserInitialized)
             {
@@ -109,15 +109,12 @@ namespace CefSharp.WinForms
             // By the time this callback gets called, this control
             // is most likely hooked into a parent Form of some sort. 
             // (Which is what MovingListener relies on.)
-            // Ensure the MovingListener contruction occurs on the WinForms UI thread:
-            if (InvokeRequired)
+            // Ensure the MovingListener construction occurs on the WinForms UI thread:
+            this.InvokeOnUiThreadIfRequired(() =>
             {
-                this.BeginInvoke((MethodInvoker)delegate 
-                {
-                    movingListener = new MovingListener(this);
-                    movingListener.Moving += movingListener_Moving;
-                });
-            }
+                movingListener = new MovingListener(this);
+                movingListener.Moving += MovingListenerMoving;
+            });
 
             ResizeBrowser();
 
--- a/CefSharp.WinForms/ChromiumWebBrowser.cs
+++ b/CefSharp.WinForms/ChromiumWebBrowser.cs
@@ -14,6 +14,7 @@ namespace CefSharp.WinForms
     public class ChromiumWebBrowser : Control, IWebBrowserInternal, IWinFormsWebBrowser
     {
         private ManagedCefBrowserAdapter managedCefBrowserAdapter;
+        private MovingListener movingListener;
 
         public BrowserSettings BrowserSettings { get; set; }
         public string Title { get; set; }
@@ -93,9 +94,30 @@ namespace CefSharp.WinForms
             base.Dispose(disposing);
         }
 
+        void movingListener_Moving(object sender, EventArgs e)
+        {
+            if (IsBrowserInitialized)
+            {
+                NotifyMoveOrResizeStarted();
+            }
+        }
+
         void IWebBrowserInternal.OnInitialized()
         {
             IsBrowserInitialized = true;
+
+            // By the time this callback gets called, this control
+            // is most likely hooked into a parent Form of some sort. 
+            // (Which is what MovingListener relies on.)
+            // Ensure the MovingListener contruction occurs on the WinForms UI thread:
+            if (InvokeRequired)
+            {
+                this.BeginInvoke((MethodInvoker)delegate 
+                {
+                    movingListener = new MovingListener(this.FindForm());
+                    movingListener.Moving += movingListener_Moving;
+                });
+            }
 
             ResizeBrowser();
 
--- a/CefSharp.WinForms/Internals/MovingListener.cs
+++ b/CefSharp.WinForms/Internals/MovingListener.cs
@@ -1,4 +1,8 @@
-﻿using System;
+﻿// Copyright © 2014 The CefSharp Authors. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style license that can be found in the LICENSE file.
+
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
--- a/CefSharp.WinForms/Internals/MovingListener.cs
+++ b/CefSharp.WinForms/Internals/MovingListener.cs
@@ -1,0 +1,61 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Windows.Forms;
+
+namespace CefSharp.WinForms.Internals
+{
+    internal class MovingListener : NativeWindow
+    {
+        public event EventHandler<EventArgs> Moving;
+
+        private Form ParentForm { get; set; }
+
+        public MovingListener(Form parent)
+        {
+            ParentForm = parent;
+            if (ParentForm.IsHandleCreated)
+            {
+                OnHandleCreated(ParentForm, null);
+            }
+            else
+            {
+                ParentForm.HandleCreated += OnHandleCreated;
+            }
+            ParentForm.HandleDestroyed += OnHandleDestroyed;
+        }
+
+        private void OnHandleCreated(object sender, EventArgs e)
+        {
+            AssignHandle(((Form)sender).Handle);
+        }
+
+        private void OnHandleDestroyed(object sender, EventArgs e)
+        {
+            ReleaseHandle();
+        }
+        protected override void WndProc(ref Message m)
+        {
+            // Listen for operating system messages 
+            switch (m.Msg)
+            {
+            // WM_MOVE
+            case 0x3:
+            // WM_MOVING:
+            case 0x0216:
+                OnMoving();
+                break;
+            }
+            base.WndProc(ref m);
+        }
+
+        protected virtual void OnMoving()
+        {
+            if (Moving != null)
+            {
+                Moving(this, null);
+            }
+        }
+    }
+}
--- a/CefSharp.WinForms/Internals/MovingListener.cs
+++ b/CefSharp.WinForms/Internals/MovingListener.cs
@@ -2,10 +2,10 @@
 //
 // Use of this source code is governed by a BSD-style license that can be found in the LICENSE file.
 
+using CefSharp.Internals;
 using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Text;
+using System.Drawing;
+using System.Runtime.InteropServices;
 using System.Windows.Forms;
 
 namespace CefSharp.WinForms.Internals
@@ -14,25 +14,58 @@ namespace CefSharp.WinForms.Internals
     {
         public event EventHandler<EventArgs> Moving;
 
+        private bool isMoving = false;
+        private Rectangle movingRectangle;
+
+        private Control ParentControl { get; set; }
+
         private Form ParentForm { get; set; }
 
-        public MovingListener(Form parent)
+        public MovingListener(Control parent)
         {
-            ParentForm = parent;
-            if (ParentForm.IsHandleCreated)
+            ParentControl = parent;
+            // Get notified if our parent window changes:
+            ParentControl.ParentChanged += parent_ParentChanged;
+            // Find the parent form to subclass to monitor WM_MOVE/WM_MOVING
+            RefindParentForm();
+        }
+
+        public void RefindParentForm()
+        {
+            parent_ParentChanged(ParentControl, null);
+        }
+
+        private void parent_ParentChanged(object sender, EventArgs e)
+        {
+            Control control = (Control)sender;
+            Form oldForm = ParentForm;
+            Form newForm = control.FindForm();
+            if (oldForm == null
+                || newForm == null
+                || oldForm.Handle != newForm.Handle
+                )
             {
-                OnHandleCreated(ParentForm, null);
+                if (this.Handle != IntPtr.Zero)
+                {
+                    ReleaseHandle();
+                }
+                if (oldForm != null)
+                {
+                    oldForm.HandleCreated -= OnHandleCreated;
+                    oldForm.HandleDestroyed -= OnHandleDestroyed;
+                }
+                ParentForm = newForm;
+                if (newForm != null)
+                {
+                    OnHandleCreated(newForm, null);
+                }
             }
-            else
-            {
-                ParentForm.HandleCreated += OnHandleCreated;
-            }
-            ParentForm.HandleDestroyed += OnHandleDestroyed;
         }
 
         private void OnHandleCreated(object sender, EventArgs e)
         {
             AssignHandle(((Form)sender).Handle);
+            Kernel32.OutputDebugString(String.Format("Attaching '{1:x}' to form '{0:x}'\r\n", ParentForm.Handle, ParentControl.Handle));
         }
 
         private void OnHandleDestroyed(object sender, EventArgs e)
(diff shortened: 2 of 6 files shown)


rassilon:
I tested this with TooltipTest.html from Cef.Example where the select drop down was in the wrong location before the change, but after the change, it seems to be fixed for me.

amaitland:
@rassilon You have been busy! Let's focus on getting #786 merged first.

For reference anyone needed a quick fix simply needs to override `OnMove` and call the relevant function. Example related in the `WinForms Example`

```c#
protected override void OnMove(EventArgs e)
{
	base.OnMove(e);

	var control = GetCurrentTabControl();
	if (control != null)
	{
		var browser = (ChromiumWebBrowser)control.Browser;
		browser.NotifyMoveOrResizeStarted();
	}
}
```

amaitland:
There are a few other factors at play as well that I'll mention briefly. There's another `Focus` related bug that will potentially need a `NativeWindow` to hook the relevant focus events (slightly different in that the proposed solution hooks the inner most `Cef` child window).

Other than the standard code review that needs to happen there's other considerations at play, so I'm just giving you a heads up that it may be a while before this is merged.

rassilon:
I'm not sure what you mean by the `Focus` related bug. i.e. I know there have been Focus related bugs, but I'm not sure what is left to do for WinForms (which is my focus atm).

I'm additionally, not sure how that might be related to this issue.

The nice thing about my approach here, is that you don't need to care in the Form's `OnMove` handler where in the control hierarchy the `WinForms` browser control is located.

Additionally, `OnMove` doesn't get called for **`WM_MOVING`** messages, only **`WM_MOVE`** messages.  

I did do some additional work on this today. Here's my current commit message for the my forthcoming commit (after some additional testing):

Much improved MovingListener that:
* Tries to handle any possible cases where the `ChromiumWebBrowser.Parent` gets changed. (`WinForms.Examples` doesn't currently need this, btw). (i.e. if you removed the control from one tab and moved it to another, instead of just creating a new control)
* Expose `RefindParentForm` as public so that the calling control has control over the call to `.FindForm()` if it becomes necessary.
* Only indicate movement when there is actual x/y movement and we're not going from Maximized to non-Maximized since this logic ends up being handled elsewhere.
* Add some debugging logic so that other folks can experiment with this scary looking change.
* Cave after realizing that every `ChromiumWebBrowser` needs to be notified of any kind of movement. (and not JUST the visible tabs from `WinForm.Examples`)
  This means we **DO NOT** check `ParentControl.Visible` on purpose, because the problem would still occur. 

This was tested by creating two tabs in `Winforms.Example`, point both at `TooltipTest.html` (from the custom scheme resources) and move/resizing around and testing both tabs after the move/resize finished.

Fyi,
Bill

rassilon:
I finished my testing for now. 

Tested via:
1. Open up two TooltipTest.html tabs.
2. Un-maximise the window.
3. Check select drop-down rendering for each tab.
4. Move window horizontally and check each tab.
5. Move window vertically and check each tab.
7. Move window diagonally and check each tab.
8. Minimize, Restore to normal, and check each tab.
9. Make window smaller by moving top edge down and check each tab.
10. Make window smaller by moving bottom edge up and check each tab.
12. Make window larger by moving left edge left and check each tab.
**NOTE:** This actually causes a **`WM_MOVE`** to occur, btw.
13. Make window smaller by moving left edge right and check each tab.
14. Maximize and check each tab.

Tests I still need to run:
* Have a SELECT drop down float right so that moving the right edge in can produce a useful test result.

If anyone has some handy HTML to do that, I'd be mightily obliged.

Fyi,
Bill

amaitland:
> I'm not sure what you mean by the Focus related bug. i.e. I know there have been Focus related bugs, but I'm not sure what is left to do for WinForms (which is my focus atm).

If your interested then the related ones are #702 #774

> I'm additionally, not sure how that might be related to this issue.

Partly code duplication, potentially both require a `NativeWindow` to hook a handle. As a reusable piece of code it's better to have one that's generic enough to be used by both. At it's core, it simply needs to forward message to `ChromiumWebBrowser`

> Additionally, OnMove doesn't get called for WM_MOVING messages, only WM_MOVE messages.

My point was that this is fixable in user land as it stands currently. For most I imagine `WM_MOVE` would be enough, overriding `WndProc` on the users form and implementing both is feasible.

Appreciate the effort you've put in, and I'm hopeful that we can merge this soon, just need to take a step back, look at the bigger picture.

CefSharp.WinForms/ChromiumWebBrowser.cs
```diff
@@ -93,10 +94,31 @@ protected override void Dispose(bool disposing)
...
+
         void IWebBrowserInternal.OnInitialized()
         {
             IsBrowserInitialized = true;
 
+            // By the time this callback gets called, this control
+            // is most likely hooked into a parent Form of some sort. 
+            // (Which is what MovingListener relies on.)
+            // Ensure the MovingListener contruction occurs on the WinForms UI thread:
+            if (InvokeRequired)
```
amaitland: There's an extension method `InvokeOnUiThreadIfRequired`, prefer to use that for consistency/brevity.

CefSharp.WinForms/ChromiumWebBrowser.cs
```diff
@@ -93,10 +94,31 @@ protected override void Dispose(bool disposing)
             base.Dispose(disposing);
         }
 
+        void movingListener_Moving(object sender, EventArgs e)
```
amaitland: - Should be private
- Also naming (Start with a capital, remove the `_`)

CefSharp.WinForms/Internals/MovingListener.cs
```diff
@@ -0,0 +1,155 @@
...
+            // Find the parent form to subclass to monitor WM_MOVE/WM_MOVING
+            RefindParentForm();
+        }
+
+        public void RefindParentForm()
+        {
+            parent_ParentChanged(ParentControl, null);
+        }
+
+        private void parent_ParentChanged(object sender, EventArgs e)
```
amaitland: Naming (Start with a capital, remove the `_`)

CefSharp.WinForms/Internals/MovingListener.cs
```diff
@@ -0,0 +1,155 @@
...
+        {
+            AssignHandle(((Form)sender).Handle);
+            Kernel32.OutputDebugString(String.Format("Attaching '{1:x}' to form '{0:x}'\r\n", ParentForm.Handle, ParentControl.Handle));
+        }
+
+        private void OnHandleDestroyed(object sender, EventArgs e)
+        {
+            ReleaseHandle();
+        }
+        protected override void WndProc(ref Message m)
```
amaitland: Missing space.

CefSharp.WinForms/Internals/MovingListener.cs
```diff
@@ -0,0 +1,155 @@
...
+            {
+                Kernel32.OutputDebugString(String.Format("Skipped OnMoving from control '{1:x}' to form '{0:x}'\r\n", ParentForm.Handle, ParentControl.Handle));
+            }
+            this.DefWndProc(ref m);
+        }
+
+        protected virtual void OnMoving()
+        {
+            isMoving = true;
+            if (Moving != null)
```
amaitland: It shouldn't matter in this case though it's best practice to take a reference to the event handler before using.

See https://github.com/cefsharp/CefSharp/blob/master/CefSharp.WinForms/ChromiumWebBrowser.cs#L128 for example

CefSharp.WinForms/Internals/MovingListener.cs
```diff
@@ -0,0 +1,155 @@
...
+            {
+                // .Left & .Right are negative when the window
+                // is transitioning from maximized to normal.
+                // If we are transitioning, the form will also receive
+                // a WM_SIZE which can deal with the move/size combo itself.
+                if (ParentForm.Left > 0
+                    && ParentForm.Right > 0)
+                {
+                    OnMoving();
+                    Kernel32.OutputDebugString(String.Format("Called OnMoving from control '{1:x}' to form '{0:x}'\r\n", ParentForm.Handle, ParentControl.Handle));
```
amaitland: What's the plan for these debug statements long term?

CefSharp.WinForms/Internals/MovingListener.cs
```diff
@@ -0,0 +1,155 @@
...
+
+            // Negative initial values keeps the compiler quiet:
+            int x = -1;
+            int y = -1;
+
+            // Listen for operating system messages 
+            switch (m.Msg)
+            {
+            // WM_MOVING:
+            case 0x0216:
```
amaitland: Prefer if these were constants

CefSharp.WinForms/Internals/MovingListener.cs
```diff
@@ -0,0 +1,155 @@
...
+            // * Not currently moving (on the UI thread only of course)
+            // * The current WindowState is Normal.
+            //      This check is to simplify the effort here.
+            //      Other logic already handles the maximize/minimize
+            //      cases just fine.
+            // You might consider checking ParentControl.Visible and
+            // not notifying our parent control if the parent control isn't visible.
+            // However, if you do that, the non-Visible CEF tab will still
+            // have any SELECT drop downs rendering where they shouldn't.
+            if (care
```
amaitland: Are all these extra checks necessary? Basic upstream usage scenario is https://bitbucket.org/chromiumembedded/branches-2171-cef3/src/cfa746b389c9b1f77746071bd0ee76412899cb49/tests/cefclient/cefclient_win.cpp?at=master#cl-651

CefSharp.WinForms/Internals/MovingListener.cs
```diff
@@ -0,0 +1,155 @@
...
+
+using CefSharp.Internals;
+using System;
+using System.Drawing;
+using System.Runtime.InteropServices;
+using System.Windows.Forms;
+
+namespace CefSharp.WinForms.Internals
+{
+    internal class MovingListener : NativeWindow
```
amaitland: Overall I'd probably prefer to have a simple `NativeWindow` that forwards messages to `ChromiumWebBrowser`, so it can be used for other scenarios.

CefSharp.WinForms/Internals/MovingListener.cs
rassilon: Remove them in the next commit after the code review fixes you're suggesting. They're basically there so you can more easily check out/test the changes.

CefSharp.WinForms/Internals/MovingListener.cs
rassilon: Hrm. ParentForm/ParentControl created might not be needed any longer. I'll give that a try.

The window state/actual movement check probably should stay.

The note about NOT checking .Visible on purpose probably needs to stay.

amaitland:
Out of curiosity have you aware if a parent change works? `CEF` can be very pick about such things.

CefSharp.WinForms/Internals/MovingListener.cs
```diff
@@ -0,0 +1,155 @@
...
+        private bool isMoving = false;
+        private Rectangle movingRectangle;
+
+        private Control ParentControl { get; set; }
+
+        private Form ParentForm { get; set; }
+
+        public MovingListener(Control parent)
+        {
+            ParentControl = parent;
```
amaitland: Isn't this really the `ChromiumWebBrowser` now? I think the naming could be clearer

CefSharp.WinForms/Internals/MovingListener.cs
```diff
@@ -0,0 +1,155 @@
...
+            // Get notified if our parent window changes:
```
amaitland: Parent Window should be Parent Control?

CefSharp.WinForms/Internals/MovingListener.cs
rassilon: Of `parent/ParentControl` you mean? i.e. rename `ParentControl` to `BrowserControl` and the type from Control to `ChromiumWebBrowser`? or something?

I have no objections if that's what you're asking. Let me know and I'll fold it in.

rassilon:
re: ParentChange: Nope, no idea if CEF likes it or not. 

The idea was to start with something useful enough for other WinForms controls and move from there.

I can put a warning doc comment about it not having been tested with CEF yet if you like.

Bill

CefSharp.WinForms/Internals/MovingListener.cs
rassilon: I'm not sure what you're saying here. (probably because I don't have enough context)

Are you asking me to change the class name from `MovingListener` to `BrowserParentFormWindowsMessageListener` or something similar so the name is a bit more generic?

If so, that's fine by me, let me have a name suggestion. :)

CefSharp.WinForms/ChromiumWebBrowser.cs
rassilon: Done. (In next commit)

CefSharp.WinForms/ChromiumWebBrowser.cs
rassilon: Done. (In next commit)

CefSharp.WinForms/Internals/MovingListener.cs
rassilon: Done. (In next commit)

CefSharp.WinForms/Internals/MovingListener.cs
rassilon: Code Lens strikes again.
Done. (In next commit)

CefSharp.WinForms/Internals/MovingListener.cs
rassilon: Done. (In next commit)

CefSharp.WinForms/Internals/MovingListener.cs
rassilon: Done. (In next commit)

CefSharp.WinForms/Internals/MovingListener.cs
rassilon: I in the end decided I needed to keep the `ParentControl.IsHandleCreated` check on the grounds that the `WndProc` isn't getting messages from `ParentControl`, it's getting the messages from `ParentForm` and WinForm handles aren't created immediately. 

I'm not sure if inserting a `ParentControl.CreateControl()` somewhere would be the thing to do or not. I'm not a WinForms expert. Thoughts?

Bill

CefSharp.WinForms/Internals/MovingListener.cs
rassilon: Depending on the naming discussion elsewhere, I'll fix the comment as appropriate.

Thanks,
Bill

CefSharp.WinForms/Internals/MovingListener.cs
amaitland: Ok cool :+1:

rassilon:
> If your interested then the related ones are #702 #774

I'll check those out next. I don't think it'll matter for my use of CEFSharp (WinForms full screen, just the browser control, nothing else), but I could give it a go.

> Partly code duplication, potentially both require a `NativeWindow` to hook a handle. 
> As a reusable piece of code it's better to have one that's generic enough to be used by both. 
> At it's core, it simply needs to forward message to `ChromiumWebBrowser`

Ah, after reading those two issues I see what you mean. `NativeWindow` is designed to only hook one **`HWNDs`** `WndProc` at a time.

If a `NativeWindow` subclass is necessary for the focus issue, it'd probably be in a separate subclass. (The logic would probably be slightly different, esp. if we need to hook a HWND inside CEF/Chromium.)

> My point was that this is fixable in user land as it stands currently. 
> For most I imagine WM_MOVE would be enough, overriding WndProc on the users form and 
> implementing both is feasible.

Oh, completely. No argument there. It'd just be awesome if this is one of the details you don't have to worry about. :)

> Appreciate the effort you've put in, and I'm hopeful that we can merge this soon, 
> just need to take a step back, look at the bigger picture.

Oh, not a problem at all. I've had code submissions to dead projects just rot on the vine, you're doing an amazing job keeping on top of this stuff.

Thanks again,
Bill

amaitland:
> re: ParentChange: Nope, no idea if CEF likes it or not.

Perhaps we need to check that? 

> The idea was to start with something useful enough for other WinForms controls and move from there.

As the class is `internal` (which I think it should be in the context of `CefSharp`), I'm thinking it would be better to only support scenarios that `CefSharp` is capable of. Thoughts?

> If a NativeWindow subclass is necessary for the focus issue, it'd probably be in a separate subclass. (The logic would probably be slightly different, esp. if we need to hook a HWND inside CEF/Chromium.)

My thought was to have a simple `NativeWindow` that had one responsibly, hook an assigned `HWND` and forward the messages via an `event/delegate`. Thoughts?

> Oh, completely. No argument there. It'd just be awesome if this is one of the details you don't have to worry about. :)

Your additional information about `WM_MOVING` and `WM_MOVE` add great value :+1: 

> Oh, not a problem at all. I've had code submissions to dead projects just rot on the vine, you're doing an amazing job keeping on top of this stuff.

From my experience here it's rare to find someone willing to contribute their time/code so for those that do I think it's important that we get things moving `ASAP` :smile: In general **everything** is open for discussion, so keep the ideas coming :+1:

amaitland:
I've been very reluctant to implement the changes proposed by #702 as it relies on an undocumented `CEF` `HWMD`. It is entirely possibly to replicate the code in user land (same as this issue can be fixed by the user), though I do appreciate that it's ideal for the framework to handle common scenarios where possible. (Just a bit of background if you do look over #702 in detail).

rassilon:
> > re: ParentChange: Nope, no idea if CEF likes it or not.
>
> Perhaps we need to check that?
>

Sadly, yes. Last thing to do of course. :) I'd have to add a button to
a tabbed browser UI thingy to try it out. (i.e swap the tabs the
browsers belong to) or something.

> > The idea was to start with something useful enough for other WinForms controls and move from there.
>
> As the class is internal (which I think it should be in the context of CefSharp), I'm thinking it would be better to only support scenarios that CefSharp is capable of. Thoughts?
>

Meh. If it's cheap and doesn't get in the way it shouldn't matter too
much. I don't have any particular problem with removing this kind of
stuff from CefSharp though, especially due to the reason we can't
check the `.Visible` property before notifying the browser.

> > If a NativeWindow subclass is necessary for the focus issue, it'd probably be in a separate subclass. (The logic would probably be slightly different, esp. if we need to hook a HWND inside CEF/Chromium.)
>
> My thought was to have a simple NativeWindow that had one responsibly, hook an assigned HWND and forward the messages via an event/delegate. Thoughts?
>

Oh, hrm. Sort of a `NativeWindowPubSub` kind of thing that would map
subclassed `WndProcs` to a set of event listeners?
That would just move the same logic elsewhere, and it might be more
difficult to coordinate there if there was only one NativeWindow that
forwarded to others. Not sure I see a big enough win to bother doing
it that way atm due to the reason explained next:

As a reminder, UI operations on Windows can (but not necessarily all
of them) involve the OS kernel calling back into user mode `WndProcs`
and that uses up kernel stack space.
If the kernel stack space runs out, then further `WndProcs` won't get
called. So this kind of stuff needs to pay its pound of flesh for
cost/benefit.

Lots of folks encountered this issue when 64bit OSs become more widely
installed. Deeply nested HWND containments that have complicated
layouts could not have some of the deeply nested relayout logic
invoked due to this issue.

Thanks again,
Bill

rassilon:
> I've been very reluctant to implement the changes proposed by #702 as it relies on an undocumented CEF HWMD. It is entirely possibly to replicate the code in user land (same as this issue can be fixed by the user), though I do appreciate that it's ideal for the framework to handle common scenarios where possible. (Just a bit of background if you do look over #702 in detail).
>

Don't know enough yet to comment, but I certainly understand your trepidation.

amaitland:
As the `Focus` issue may take some time, it maybe worth focusing on this PR for now, getting it ready for merged.

So my thoughts currently are:

- Rename `MovingListener` to `ParentFormMessageInterceptor` (Or something similar)
- Change the constructor of `ParentFormMessageInterceptor` to accept the `ChromiumWebBrowser` type, rather than `Control`, thus removing the need for the `EventHandler` and additional casts.
- Rename `ParentControl` property to `Browser` (just for clarity).
- Remove any additional unneeded complexity (would require checking to see if the parent can be changed). I think this one is fairly important from a long term maintenance point of view. Simplify down to only what is required to solve the problem at hand.
- Change to using constants for `WM_` values
- Add braces around `case` blocks (code consistency, everything has braces in `CefSharp`)
- Have `ParentFormMessageInterceptor` implement `IDisposable` and call `ReleaseHandle();` to cleanup code, plus unhook any events related to the parent form.
- Rename the `care` variable to something a little more descriptive, just something that you can glance at and know a little more about what's going on.
- Finally remove the debugging when everyone's happy.

(I'm not particularly attached to the name `ParentFormMessageInterceptor`, just prefer something that conveys the maximum amount of information in relation to the classes intent).

Thoughts/comments/Improvements/Objections? Happy to contribute some time to implementing the proposed changes.

rassilon:
> * Rename MovingListener to ParentFormMessageInterceptor (Or something similar)
Done.

> * Change the constructor of ParentFormMessageInterceptor to accept the ChromiumWebBrowser type, rather than Control, thus removing the need for the EventHandler and additional casts.
Done.

> * Rename ParentControl property to Browser (just for clarity).
Done.

> * Remove any additional unneeded complexity (would require checking to see if the parent can be changed). I think this one is fairly important from a long term maintenance point of view. Simplify down to only what is required to solve the problem at hand.

Parent's can be changed. I just tested a silly "switch browser controls around" menu item for the first two tabs of BrowserForm.cs  And it worked properly. I'm not really sure how useful such a thing is. I don't have a good feeling for how expensive each additional ChromeWebBrowser control is after the first one (assuming the contents are simple).

> * Change to using constants for WM_ values
Done. Added a NativeMethods class in CefShar.WinForms for them to live in.

> * Add braces around case blocks (code consistency, everything has braces in CefSharp)
Done.

> * Have ParentFormMessageInterceptor implement IDisposable and call ReleaseHandle(); to cleanup code, plus unhook any events related to the parent form.
Hrm. I started doing this, but I have this nasty suspicion that this needs to happen on the owning UI thread.  Some more thought is required here, I think. Unless we say, don't do that for now. (which is fine by me)

> * Rename the care variable to something a little more descriptive, just something that you can glance at and know a little more about what's going on.
Done. I renamed it to `isMovingMessage` .

Additionally, I think it might be wise for this class to handle more than one `ChromiumWebBrowser` control listening to the same parent form. That way cleanup ordering of unsubclassing is less likely to cause grief. (But that could be a happy later addition of course.)

Bill

rassilon:
Here's that menu handler I used for parent changing goo for reference:

        private void SwitchBrowserControlsAroundToolStripMenuItemClick(object sender, EventArgs e)
        {
            if (browserTabControl.TabCount > 0)
            {
                // A silly construction to test the ability of ParentFormMessageInterceptor to have
                // the browser's parent control changed.
                BrowserTabUserControl tab1 = (BrowserTabUserControl)browserTabControl.TabPages[0].Controls[0];
                BrowserTabUserControl tab2 = (BrowserTabUserControl)browserTabControl.TabPages[1].Controls[0];
                var browser1 = tab1.Browser;
                var browser2 = tab2.Browser;
                // NULL out tab2 so that a browser won't simultaneously exist in two Controls collections.
                tab2.Browser = null;
                tab1.Browser = browser2;
                tab2.Browser = browser1;
            }
        }

This code obviously didn't re-wire events tied to the address textbox, but you get the idea and it did indeed work correctly.

rassilon:
Oh, I just added a `style="float: right"` `select` to `TooltipTest.html` and this set of changes behaved just fine with adjusting the right edge smaller and larger.

Fyi,
Bill

rassilon:
> IDisposable

After thinking this through some more, maybe the reason NativeWindow isn't IDisposable itself is due to this issue of needing to ReleaseHandle() on the UI thread for the HWND.

Raymond Chen's latest blog discusses a very closely related problem and about what threads to release UI objects at: http://blogs.msdn.com/b/oldnewthing/archive/2015/02/12/10592892.aspx
As you can see by reading his blog, things can get messy fast.

Thoughts?
Bill

amaitland:
I'll have a read and get back to you ASAP, thanks :+1:

amaitland:
Ok, so finally got a change to read the article, Is there a problem with marshaling the `NativeWindow` destructor call onto the `UI` thread?

> Additionally, I think it might be wise for this class to handle more than one ChromiumWebBrowser control listening to the same parent form. That way cleanup ordering of unsubclassing is less likely to cause grief. (But that could be a happy later addition of course.)

I think for a tabbed environment this is probably important. Also `disposing` of the `NativeWindow` is much less likely to leak memory of it's bound to a lifecycle of a `Window`, rather than that of an individual browser.

How much work do you think is involved?

rassilon:
> Ok, so finally got a change to read the article, Is there a problem with marshaling the NativeWindow destructor call onto the UI thread?

C# destructor's are always invoked on the GC thread, (you probably meant `Dispose`) but you could issue a BeginInvoke() to push a method callback onto the UI thread from wherever Dispose got called. However, that method might never get called. (The UI message pump might no longer be running.) 
(Which is why using an Invoke() would be bad.)

> handling more than one ChromiumWebBrowser per parent form

I think is important as well. I don't think its too difficult, but it is certainly more complex than other approaches and probably needs to involve a thread safe static mapping of web browsers to parent form handles.

amaitland:
> (you probably meant Dispose)

You are correct, I did mean `Dispose`, my mistake.

> I think is important as well.

Important enough to make more changes before merging this? Would you prefer to see this merged straight away?

rassilon:
> Important enough to make more changes before merging this? Would you prefer to see this merged straight away?

Probably not important enough to not merge this in as is, but I'd certainly like to get it fixed before the next release of course.

Merging this PR will make this next step easier to review since some of the code will be in already of course.

I got derailed the last two weeks by needing to write code to play MP3 files via NAudio (.Net wrapping of core audio), since CEF can't play them by default without a MP3 decoder license. (and simultaneously make corp lawyers happy of course) :(

So, this is still on my radar, it just got bumped down the priority list a little bit. :(

Fyi,
Bill

amaitland:
Are you happy to ditch the event handler and call the notify method directly? Apart from that happy to merge now.

rassilon:
Notifying control directly now, no need to invoke handler. Kept `OnMoving` to keep it away from the WndProc mess.

amaitland:
I committed a few other changes (build was failing). Mostly just resharper inspired cleanups :+1:

amaitland:
I'll probably release another `-pre` package shortly.

rassilon:
Doh, thanks for fixing up my typos.

amaitland:
No probs.

I'd like to resolve issue #752 before releasing the next `-pre` as it's partially implemented now. Any thoughts or comments to add there?

PR #813, opened 2015-02-15
WPF Rendering - Implement WritableBitmap and Minor Optimisations

amaitland:
Added enough abstraction to rendering so a few new scenarios are possible:

- Support `WritableBitmap` as an alternative implementation
- Implement a .Net 4.5 version of `InteropBitmap` that supports `DirtyRects` (currently would need to be supported in a different assembly).

Also includes
- Code cleanup: removed some extra duplication that I'd not gotten around to fixing as of the last changes.
- More tweaks: some values were being calculated for each frame unnecessarily, so this should squeeze a little more performance out of the rendering.


amaitland:
I'm going to merge this, then look at releasing another `-pre` package after some other issues are resolved.

Overall I think we need a little more testing on a few areas before releasing `39` officially.

PR #827, opened 2015-02-20
Fixed memory leak issues:

dmrzn:
1. Unsubscription from CleanupElement.Unloaded event
2. Unsubscription from tooltipTimer.Tick event
3. Removing source hook

### DIFF
--- a/CefSharp.Wpf/ChromiumWebBrowser.cs
+++ b/CefSharp.Wpf/ChromiumWebBrowser.cs
@@ -362,6 +362,16 @@ namespace CefSharp.Wpf
 
                 IsVisibleChanged -= OnIsVisibleChanged;
 
+                if (tooltipTimer != null)
+                {
+                    tooltipTimer.Tick -= OnTooltipTimerTick;
+                }
+
+                if (CleanupElement != null)
+                {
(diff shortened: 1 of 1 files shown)


CefSharp.Wpf/ChromiumWebBrowser.cs
```diff
@@ -651,12 +663,20 @@ private void PresentationSourceChangedHandler(object sender, SourceChangedEventA
             {
                 if (source != null && sourceHook != null)
```
amaitland: Duplicate `null` checks by the looks of it.

amaitland:
@dmrzn In general looks solid :+1: Thanks! Shouldn't take long before this is merged.

amaitland:
@dmrzn The tooltip timer is a bit of a mess overall, so I've cleaned it up quite a bit in #831 

Haven't checked for memory leaks, perhaps you could take a look if you have time?

The timer is now implemented in an extension method see https://github.com/cefsharp/CefSharp/pull/831/files#diff-80281b988d81b59e8a6f3df0bde09f2dR17

amaitland:
Soon as the duplicate null check has been resolved I'm happy to merge this in the short term, then focus in improving the tooltip.

PR #831, opened 2015-02-23
WPF - Cleanup Tooltip Timer

amaitland:
- The tooltip timer code is messy, this extracts the functionality into an extension method for simplicity.
- I've also extracted `UiThreadRunAsync` into an extension method, part of my overall plan to reduce the amount of code in the `ChromiumWebBrowser` class where possible.


amaitland:
Bumping this till after `39`, it's not critical and would require more testing. Prefer to get `39` out the door.

amaitland:
Need to retarget this to the `2272` branch,

PR #832, opened 2015-02-23
Expose TaskScheduler

amaitland:
Exposes the `TaskScheduler` that's used in the `BrowserSubProcess`. There are quite a number of `Cef` calls that can only happen on a specific thread. Ideally the cookie handling code would be rewritten to properly use tasks with this change.

Not looking to directly merge this now, just putting it out there incase anyone wishes to expand on the idea further.


amaitland:
Outstanding:
- Merge in `master`
- Add license disclaimer.

PR #834, opened 2015-02-24
Expose Cross-Origin Whitelist from Cef.XXX as static methods.

rassilon:
Custom schemes are much easier to deal with if you can use XHR with them, so on the theory that these APIs work, expose them from CefSharp.

See CEF issues 246 and 950 for more details.

Let me know if these methods should go somewhere else.

Fyi,
Bill

### DIFF
--- a/CefSharp.Core/Cef.h
+++ b/CefSharp.Core/Cef.h
@@ -7,6 +7,7 @@
 #include <msclr/lock.h>
 #include <include/cef_version.h>
 #include <include/cef_runnable.h>
+#include <include/cef_origin_whitelist.h>
 
 #include "Internals/CefSharpApp.h"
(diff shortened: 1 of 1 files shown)


amaitland:
Few changes required before merging this one

- Add xml documentation
- Rename variables to be camcelCase without underscores

rassilon:
Done. If this could be part of the -pre03 on Nuget that'd be very cool. It's so much easier to build my project with it that way. :)

Bill

amaitland:
> Done. If this could be part of the -pre03 on Nuget that'd be very cool. It's so much easier to build my project with it that way. :)

Thanks! More than happy for this to be in `-pre03` :+1:

amaitland:
`-pre03` is on `nuget`

Issue #836, opened 2015-02-24
WPF - Implement CefScreenInfo


amaitland:
Look at implementing `CefRenderHandler.GetScreenInfo` and specifying `device_scale_factor`. Hopefully be able to get `CEF` to do all the scaling and remove the custom WPF `Transform`.

http://magpcss.org/ceforum/apidocs3/projects/%28default%29/CefRenderHandler.html#GetScreenInfo%28CefRefPtr%3CCefBrowser%3E,CefScreenInfo&%29

http://magpcss.org/ceforum/apidocs3/projects/%28default%29/_cef_screen_info_t.html#device_scale_factor

```c++
//CefRenderHandler
virtual bool GetScreenInfo(CefRefPtr< CefBrowser > browser, CefScreenInfo& screen_info) OVERRIDE
{
	
};
```

Really just adding this as a note to self.

bjarteskogoy:
I'm looking into this one. A quick test seem to prove that this works.

amaitland:
More info
https://code.google.com/p/chromiumembedded/issues/detail?id=946#c6

bjarteskogoy:
I wonder if a part of the problem is the dpiSize referred to here:
gfx::Size ToPixelSize(gfx::Size dipSize, float scale) {
  return gfx::ToCeiledSize(gfx::ScaleSize(dipSize, scale));
}

PR #842, opened 2015-02-25
Use device scale factor to support higher DPI on WPF

bjarteskogoy:
https://github.com/cefsharp/CefSharp/issues/836

### DIFF
--- a/CefSharp.Wpf/ChromiumWebBrowser.cs
+++ b/CefSharp.Wpf/ChromiumWebBrowser.cs
@@ -655,6 +655,7 @@ namespace CefSharp.Wpf
                 if (source != null)
                 {
                     matrix = source.CompositionTarget.TransformToDevice;
+                    _deviceScaleFactor = (float)matrix.M11;
                     sourceHook = SourceHook;
                     source.AddHook(sourceHook);
                 }
@@ -809,6 +810,12 @@ namespace CefSharp.Wpf
                 throw new Exception("BitmapFactory cannot be null");
             }
             return BitmapFactory.CreateBitmap(isPopup);
+        }
+
+        float _deviceScaleFactor = 1.0F;
+        float IRenderWebBrowser.GetScreenInfoScaleFactor()
+        {
+            return _deviceScaleFactor;
         }
 
         void IRenderWebBrowser.InvokeRenderAsync(BitmapInfo bitmapInfo)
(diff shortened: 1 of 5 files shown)


amaitland:
Is it possible to remove the `ScaleTransform` now?

CefSharp.Wpf/ChromiumWebBrowser.cs
```diff
@@ -811,6 +812,12 @@ public BitmapInfo CreateBitmapInfo(bool isPopup)
             return BitmapFactory.CreateBitmap(isPopup);
         }
 
+        float _deviceScaleFactor = 1.0F;
```
amaitland: Underscores only in `c++` code - consistency.
Variables should also be declared in the one place for consistency.

amaitland:
Additional info on scaling in this thread https://code.google.com/p/chromiumembedded/issues/detail?id=946#c6

bjarteskogoy:
Actually it's the combination of the ScaleTransform and device scale factor that makes it work properly. Without the transform and the device scale factor the wpf browser content would be the correct size on screen. WPF will upscale the image to comply with the current DPI setting, but this would just be an upscaling of the bitmap image -> very blurry. The scale transform keeps the pixel size from cef, which leads to a too small image when you have a higher DPI than 100%. The device scale factor "zooms" the rendered content from cef up to the correct size on screen.

amaitland:
That's based on the transformed dimensions. I still think there's some investigation required.

amaitland:
We should also expose this

http://magpcss.org/ceforum/apidocs3/projects/(default)/CefBrowserHost.html#NotifyScreenInfoChanged()

bjarteskogoy:
The NotifyScreenInfoChanged should then probably be triggered in this method: https://github.com/cefsharp/CefSharp/blob/master/CefSharp.Wpf/ChromiumWebBrowser.cs#L647

That would also mean that the scale transform needs to be updated at the same time, I believe.

amaitland:
I've merged this though I still think further research is needed. The fact that it works is definitely an improvement, though I'm not convinced what we have is the `correct` solution.

> Currently, all sizes passed in and out of the API for both CEF and Chromium, are in view units (meaning virtual pixels or unscaled pixels) by design

 It's my understanding that `WPF` also uses device independent pixels and as such I'm wondering if the `ScaleTransform` is required because we provide transformed `Width/Height`. I don't have a `high dpi` device, so can't really toy with this easily.

PR #852, opened 2015-02-27
Feature/device scale factor

bjarteskogoy:
Minor change needed to function correctly. (Proper render area)

I've also investigated a bit on the scale transform. It seems like CreateBitmapSourceFromMemorySection returns an image source with DpiX and DpiY set to 96 regardless of the actual dpi on the system. This is probably why the image is bitmap up-scaled. So for now the transform is needed to compensate for this.

### DIFF
--- a/CefSharp.Core/Internals/RenderClientAdapter.h
+++ b/CefSharp.Core/Internals/RenderClientAdapter.h
@@ -69,6 +69,7 @@ namespace CefSharp
                 }
 
                 screen_info.device_scale_factor = scaleFactor;
+
                 return true;
             }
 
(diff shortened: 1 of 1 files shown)


amaitland:
Width and height are already transformed, so I'm a little concerned.

I'm contemplating reverting the previous changes until more testing/feedback have been conducted.

I think we should ask upstream for clarification

bjarteskogoy:
Yes. The previuos change will not behave correctly without this last change. Might be a good idea to check upstream. We will anyway deploy this to production. It's better than nothing, for now.

amaitland:
I've merged in the changed, and commented them out in a later commit.

When I tried with `125%` on my standard `1080` monitor the mouse clicks/drags were out, so more work is required to get this working properly.

PR #854, opened 2015-03-02
WPF DPI

bjarteskogoy:
- pixel position was translated
- RenderClientAdapter was getting transformed sizes

Removed code probably needed when GetScreenInfo wasn't implemented. Removing this seems to fix all issues. Tested on 125% dpi.

Now the only transform needed is down scaling the bitmap from cef, probably due to the InteropBitmap having dpi size 96, even when running with higher DPI. Dont' now what could be done to fix this.

With the WritableBitmap dpi size is specified in the constructor (https://github.com/cefsharp/CefSharp/blob/master/CefSharp.Wpf/Rendering/WritableBitmapInfo.cs#L50). Using the correct dpi here gets rid of the need for transforms. Tested this too on 125%.

### DIFF
--- a/CefSharp.Core/Internals/RenderClientAdapter.h
+++ b/CefSharp.Core/Internals/RenderClientAdapter.h
@@ -56,6 +56,8 @@ namespace CefSharp
             // CefRenderHandler
             virtual DECL bool GetScreenInfo(CefRefPtr<CefBrowser> browser, CefScreenInfo& screen_info) OVERRIDE
             {
+                return false;
+
                 if ((IRenderWebBrowser^)_renderWebBrowser == nullptr)
                 {
                     return false;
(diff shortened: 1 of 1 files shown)


bjarteskogoy:
Have to clean this up a bit. Didn't see that you already added a struct for ScreenInfo. Will fix that.

amaitland:
> Have to clean this up a bit. Didn't see that you already added a struct for ScreenInfo. Will fix that.

:+1: 

> Now the only transform needed is down scaling the bitmap from cef, probably due to the InteropBitmap having dpi size 96, even when running with higher DPI. Dont' now what could be done to fix this.

Quick look and it appears that `InteropBitmap` is somewhat hardcoded to `96dpi`, will have a look to see if there are any ways around that. Potentially add a `DPI` check and switch to `WritableBitmap` if required.

amaitland:
@bjarteskogoy Very nice pickup on the hard coded DPI BTW :+1: 

I think short term, we remove the transform, and switch to `WritableBitmap` if using a `DPI` other than `96`. Thoughts?

PR #855, opened 2015-03-02
WPF scaling

bjarteskogoy:
Didn't mean to close the previous pull request.

I'm OK with WritableBitmap if DPI is anything else than 96 DPI. Have you seen any relevant performance differences.

The only transform needed for InteropBitmap to work now, I believe, is transforming the image du to the DPI issue: https://github.com/cefsharp/CefSharp/blob/master/CefSharp.Wpf/ChromiumWebBrowser.cs#L820

If the transform gives a performance penalty, then maybe WritableBitmap is better anyway?

### DIFF
--- a/CefSharp.Core/Internals/RenderClientAdapter.h
+++ b/CefSharp.Core/Internals/RenderClientAdapter.h
@@ -56,8 +56,6 @@ namespace CefSharp
             // CefRenderHandler
             virtual DECL bool GetScreenInfo(CefRefPtr<CefBrowser> browser, CefScreenInfo& screen_info) OVERRIDE
             {
-                return false;
-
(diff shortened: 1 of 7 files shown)


amaitland:
> Have you seen any relevant performance differences.

I haven't used it enough to say for sure, my initial thoughts were the difference is negligible.

bjarteskogoy:
WritableBitmap is at least a way to got then. I see that the bitmaps are created in the constructor of RenderClientAdapter. At that time we don't have a valid scale factor. So that would need reworking.

amaitland:
> So that would need reworking.

Let me know how you go, I'll take a look at it further tomorrow if required :+1:

PR #856, opened 2015-03-03
Enhance StreamAdapter to more effeciently deal with MemoryStream

rassilon:
...emoryStreams, and add lame header to ResourceHandler usage in CefExample.cs.

Additionally, allocate Headers NameValueCollection in constructor of ResourceHandler.

Sadly, without a change to ClientAdapter.cpp to pass the header data to CEF due to a debug flavor bug in CEF, you can't easily test this change. :(

If you do test this with the CEF change or a Release flavor with the ClientAdapter.cpp change, then this works correctly.

Bill

### DIFF
--- a/CefSharp.Core/Internals/StreamAdapter.h
+++ b/CefSharp.Core/Internals/StreamAdapter.h
@@ -17,6 +17,7 @@ namespace CefSharp
         {
             CriticalSection _syncRoot;
             gcroot<Stream^> _stream;
+            bool isMemoryStream = false;
 
         public:
             virtual ~StreamAdapter();
@@ -24,6 +25,7 @@ namespace CefSharp
             {
                 //Reset stream position
                 stream->Position = 0;
+                isMemoryStream = (dynamic_cast<MemoryStream^>(stream)) != nullptr;
             }
 
             virtual size_t Read(void* ptr, size_t size, size_t n);
--- a/CefSharp.Core/Internals/StreamAdapter.h
+++ b/CefSharp.Core/Internals/StreamAdapter.h
@@ -17,7 +17,7 @@ namespace CefSharp
         {
             CriticalSection _syncRoot;
             gcroot<Stream^> _stream;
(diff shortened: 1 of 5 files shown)


jankurianski:
Looks like the build failed due to the following line in `StreamAdapter.h`:

`bool isMemoryStream = false;`

```c:\projects\cefsharp\cefsharp.core\internals\StreamAdapter.h(20): error C2864: 'CefSharp::Internals::StreamAdapter::isMemoryStream' : only static const integral data members can be initialized within a class [C:\projects\cefsharp\CefSharp.Core\CefSharp.Core.vcxproj]
c:\projects\cefsharp\cefsharp.core\internals\StreamAdapter.h(20): error C2864: 'CefSharp::Internals::StreamAdapter::isMemoryStream' : only static const integral data members can be initialized within a class [C:\projects\cefsharp\CefSharp.Core\CefSharp.Core.vcxproj]```

**Comment by amaitland** [comment]
I was thinking we do the `MemoryStream` check in `C#` and pass into the constructor, little less uglier code :smile: Thoughts?

File: CefSharp.Core/Internals/StreamAdapter.h [hunk 1]
```diff
@@ -17,6 +17,7 @@ namespace CefSharp
         {
             CriticalSection _syncRoot;
             gcroot<Stream^> _stream;
+            bool isMemoryStream = false;
```
amaitland: In `C++` we prefix with `_`

(I would like to make the `C#` code consistent, though I've had some resistance).

rassilon: In next push.

rassilon:
> I was thinking we do the MemoryStream check in C# and pass into the constructor, little less uglier code :smile: Thoughts?

`StreamAdapter` isn't called directly from C# and it's a simple C# `as` check anyway as it is. (via `dynamic_cast<>` in C++.

Bill

PR #857, opened 2015-03-03
Rename IResourceHandler to IResourceHandlerFactory.

rassilon:
Start the process of cleaning up `IResouceHandler` by renaming it something sensible.

Thoughts?

If you think the name is ok, I'll push a rename of the files.

Bill


CefSharp/IResourceHandler.cs
```diff
@@ -4,18 +4,18 @@
...
-    public interface IResourceHandler
+    public interface IResourceHandlerFactory
     {
         void RegisterHandler(string url, ResourceHandler handler);
 
         void UnregisterHandler(string url);
 
         /// <summary>
-        /// Called before a resource is loaded. To specify a handler for the resource return a ResourceHandler object
+        /// Called before a resource is loaded. To specify a handler for the resource return a ResourceHandlerFactory object
```
amaitland: Little bit to aggressive with the rename? I think this should still be `ResourceHandler`

amaitland:
I'm happy with `IResourceHandlerFactory` :+1: Naming was confusing before.

I think the rename may have been a little too aggressive, at least a couple of instances that I think should stay as `ResourceHandler`.

Other than that, rename away.

CefSharp/IWebBrowser.cs
```diff
@@ -54,11 +54,11 @@ public interface IWebBrowser : IDisposable
         void Load(string url);
 
         /// <summary>
-        /// Registers and loads a <see cref="ResourceHandler"/> that represents the HTML content.
+        /// Registers and loads a <see cref="ResourceHandlerFactory"/> that represents the HTML content.
```
amaitland: Possibly this should stay the same as well? (And the next few instances).

Feel free to improve the comments as you see fit.

rassilon: Yes they should. Fixed.

amaitland:
@jankurianski Is renaming going to cause you any problems?

amaitland:
Will wait for feedback from @jankurianski before merging this one as he's likely to be affected.

Were you planning any other changes? I know you'd mentioned switching to a `ConcurrentDictionary`

rassilon:
Yeah, I do want to switch this to a `ConcurrentDictionary`. I figured the names might be a blocker, so I figured do this bit first, and then do the next. 

After `ConcurrentDictionary` is probably renaming of the Scheme ResoureHandlerWrapper class to prep for creating the ability to do `ResourceHandler's` via async. 

Either via a known `ResourceHandler` subtype i.e. `AsyncResourceHandler` and the C++ caller of `ResourceHandlers` can do the right thing based on being an `AsyncResourceHandler` child or not.
Or, by changing the `IResourceHandlerFactory` interface `GetResourceHandler` method to return an `IResourceHandler` instead.

Bill

amaitland:
@rassilon Are you interested in joining `Team CefSharp`? (Long as @jornh gives his blessing of course)

rassilon:
Sure. I think the `AsyncResourceHandler ` might be the simplest way to go forward in order to minimize backward compat problems.

Bill

amaitland:
> I think the AsyncResourceHandler might be the simplest way to go forward in order to minimize backward compat problems.

Sounds like the best of both worlds :+1:

jankurianski:
@amaitland My apologies for the delay - I am not affected by this.

I also think @rassilon would be a great addition to the team!

amaitland:
> @amaitland My apologies for the delay - I am not affected by this.

No problem :+1: Thanks for confirming.

amaitland:
> @rassilon Are you interested in joining Team CefSharp? (Long as @jornh gives his blessing of course)

@jornh Any comments on my idea?

jornh:
Sorry Alex, must have overlooked this one :blush: I'm very much in favour of adding @rassilon to the team as well.

Bill, welcome aboard! From what I've seen you have already contributed significantly to CefSharp in a very short span of time... Very much appreciated :+1:

amaitland:
> Sorry Alex, must have overlooked this one :blush:

No problem at all :smile:

PR #858, opened 2015-03-03
Enhancement - WPF High DPI Support

amaitland:
Based on #855 

- Reorder creation of `BitmapInfo` objects and use the matrix transform provided by `PresentationSourceChangedHandler`
- Remove more traces of transform
- More `IBitmapFactory` into the `WPF` project.


amaitland:
@bjarteskogoy I reordered the `BitmapInfo` creation slightly so it's a little cleaner now.

Thoughts?

bjarteskogoy:
I'm happy with this. Feels a bit cleaner to not be dependent on the MainWindow of the application. Not sure if it would matter, but this looks good.

Also important to be aware of that the PresentationSourceChangedHandler is triggered when you remove a control from it's parent, and triggered again when it's attached to the same or a new parent. Might be that the NotifyScreenInfoChanged should only be called if PresentationSourceChangedHandler has a new value with a different matrix than the last one?

amaitland:
> NotifyScreenInfoChanged should only be called if PresentationSourceChangedHandler has a new value with a different matrix than the last one?

Sounds reasonable, the initial call to `NotifyScreenInfoChanged` is basically a `noop` as the browser hasn't initialized yet, so adding a conditional block should be no problem.

bjarteskogoy:
Should it also, if a new and different matrix is introduced, get rid of the old bitmaps and create new ones? Not sure if it can happen. Maybe by dragging the window between monitors with different dpi? Windows 10 is reported to support different dpi per monitor.

amaitland:
Would the `Width/Height` change in that scenario? If that's the case then `WasResized` will be called and a new `Bitmap` will be generated (every change of size generates a new bitmap).

bjarteskogoy:
Width/Height should not change. Only the scale factor.

amaitland:
Probably requires a little more though then. Actually running it on `Windows 10` would be the defining test.

bjarteskogoy:
Maybe not worth spending time on yet, though. I would go with the solution as it is, for now.
Seems that the application need to tell Windows that it supports per monitor dpi and handle a windows message for picking it up: https://code.msdn.microsoft.com/windowsdesktop/Per-Monitor-Aware-WPF-e43cde33/sourcecode?fileId=102792&pathId=841375614

amaitland:
I've tested this a little further and unfortunately drop down lists render incorrectly, they're offset for some reason. So this will still need a bit of work.

Tested on Win7Sp1 `32bit` Cef build. using `125%` scale

bjarteskogoy:
Do you have an example page where this happens?

bjarteskogoy:
http://magpcss.org/ceforum/apidocs3/projects/%28default%29/CefRenderHandler.html#GetViewRect(CefRefPtr<CefBrowser>,CefRect&)

"Relative to screen coordinates" -> Does that mean that the returned CefRect should not have x,y = 0 ?

bjarteskogoy:
Does this mean that getscreeninfo should return the monitor size, not the window size: https://github.com/adobe/chromium/blob/cfe5bf0b51b1f6b9fe239c2a3c2f2364da9967d7/content/browser/renderer_host/render_widget_host_view_aura.cc#L119 ?

bjarteskogoy:
Getting monitor in WPF: http://www.codeproject.com/Tips/316111/Show-or-Activate-a-WPF-window-under-the-current-mo

bjarteskogoy:
Tested with: http://www.tutorialspoint.com/html/html_select_tag.htm

Translating width/height and offset in SetPopupSizeAndPositionImpl fixes the issue. Kind of makes sense. I think. CEF is working with virtual pixels, WPF seems not too: http://www.philosophicalgeek.com/2009/11/21/how-to-position-windows-correctly-at-120dpi-in-wpf/

bjarteskogoy:
more info on per monitor dpi: https://github.com/paulcbetts/PerMonitorDpi

bjarteskogoy:
Ok. Done a lot of testing. Translating offset and width / height in SetPopupSizeAndPositionImpl is a 99% solution. The only issue I've found is if the dropdown grows beyond the bottom of the screen. WPF will then reposition the dropdown to fit inside of the screen and mouse positions do not match what you see anymore.

I have tried to implement GetScreenInfo and GetViewInfo like they're supposed to be implemented here: https://github.com/bjarteskogoy/CefSharp/tree/enhancement/wpf-high-dpi

Can't get the popup in the correct place. And I'm not sure if GetScreenInfo wants device pixels or not. Now I'm sending transformed sizes for everything. Everything looks good except the positioning of the popup. My implementation og GetViewInfo now depends on being on the Dispatcher. Maybe that can be avoided? If not, it would be better with some kind of wrapping, like SetPopupSizeAndPosition.

bjarteskogoy:
OnPopupSize now passes view coordinates instead of (potentially scaled) pixel coordinates
https://github.com/svn2github/cef3/commit/bd28137b2158c1257c1189aad5b72691dd7deeb3

bjarteskogoy:
https://code.google.com/p/chromiumembedded/issues/detail?id=1566&sort=-id

amaitland:
Sounds like your making progress, will have a look when I get a chance :+1: 

As this is an ongoing issue, I might create a branch on the main repository so PR's can happen against that? Once we're happy then that can be merged into `master`

amaitland:
> - Add OSR unit tests for 2x (HiDPI) pixel scaling.
- Improve CefRenderHandler documentation.

Hopefully some of that will translate to `2171`, though I may mean that to get proper support we need to use `trunk`.

bjarteskogoy:
I believe we can anyway merge in and apply the writablebitmap implementation for DPI > 100%. And transform the sizes for the popup. This is probably as expected, due to this change not being relevant yet: "OnPopupSize now passes view coordinates instead of (potentially scaled) pixel coordinates"

Then we can continue to work on this on a new branch. Migh be best to upgrade CEF before going any further with that?

amaitland:
I've merged your changes into this `PR` along with `master`

Looks like the `OffScreen` project hasn't been implemented. Few changes I'd like to see before merging this:

- Fix formatting
- Remove extra space
- Remove mouse click dev tools
- Move the native method calls into `NativeMethodWrapper` (no need for dll imports as we have a `c++` project to handle all this :smile:)
- Like to investigate the `PointToScreen` call, see if there's another option that doesn't require the dispatcher. Potentially we only need to do that when using a `high dpi`, not just for standard dpi

Have you put this to use yet? Any comments?

bjarteskogoy:
We've put the original code with WriteableBitmap for scale <> 1, with popup transform into production. No issues yet.
I submitted this to get some clarification regarding CefRenderHandler: https://code.google.com/p/chromiumembedded/issues/detail?id=1576

amaitland:
Will wait for some clarification from upstream before proceeding I think. Let me know how it goes :+1:

amaitland:
@bjarteskogoy Is bumping this to the next release going to be a problem? I'd like to get `39` out the door ASAP. 

Think we really do need `upstream` clarification before proceeding.

bjarteskogoy:
Not a problem.

PR #863, opened 2015-03-05
Update ResourceHandler.cs

Jayman1305:
Missing mime type for svg added


jankurianski:
Looks good, thanks @Jayman1305! :+1:

Issue #825, opened 2015-02-20
Feature Enhancement - Calling back into an anonymous JS function


rassilon:
Currently, `IRenderProcess` only supports calling globally exposed JS functions. That doesn't mesh very well with modern JS programming styles that use oodles of anonymous functions.

Here is a very vague outline of how this might happen in CefSharp:

`TypeUtils.cpp:ConvertFromCEF`: If V8 function: then track function via map approach suggested by [CEF](https://code.google.com/p/chromiumembedded/wiki/GeneralUsage#Custom_Implementation), serialize via browser_id, frameId, callbackId back to browser process wrapper class. `JavascriptCallbackWrapper` or something. I'm not entirely sure which class the suggested map should exist in CefSharp though. TypeUtils is a static class currently. `CBrowserWrapper` maybe?

`IRenderProcess`: Add `IAsyncResult BeginExecuteCallback(callbackStateObject, object[] args)` and `EndExecuteCallback()` methods similar to existing script executions methods on `IRenderProcess`.

`CefAppUnmanagedWrapper::OnContextReleased`: Release the new map.

NOTE: this wouldn't try to do anything to track the JavaScript `this` value, that'd be up to the callback to supply the appropriate bind call.

Thoughts?
Bill

amaitland:
> TypeUtils is a static class currently. CBrowserWrapper maybe?

`TypeUtils` is just a helper class, being used in a few place, I'd try and void changing it if possible. I think `CefBrowserWrapper` is the place to keep the dictionary/map.

On the `C++` side of things there's probably some refactoring that needs to happen to implement this cleanly, that will be the messiest part.

For me the key parts are:
- Iterate through the params, convert where possible, if a function then generate a unique Id, store a reference to the context and function it's self in the dictionary/map. Insert a proxy object into the params passed to the `WCF` call. 
- When `OnContextReleased` then release the map contained within the browser wrapper.
- Make appropriate `WCF` calls

Perhaps roughly map out the public facing `API`? 

```c#
public string Repeat(string str, int n, IJavascriptCallback callBack)
{

}
```

The slightly less intuitive approach would be to have a method on the browser to execute the callback, I like this idea from the point of view the callback wrapper is just a lightweight object containing an `id` plus some other information, no need to keep track of a `WCF` proxy, so I'd say there will be fewer memory leaks. The browser actually executes the calls.
`browser.ExecuteJavascriptCallBack(callBack, params);`

The other more traditional approach is of course something like
[...]

PR #868, opened 2015-03-09
Javascript callbacks

arsher:
Prototype JS callback implementation as discussed in [Issue 825](https://github.com/cefsharp/CefSharp/issues/825).

### DIFF
--- a/CefSharp.BrowserSubprocess.Core/JavascriptCallbackRegistry.cpp
+++ b/CefSharp.BrowserSubprocess.Core/JavascriptCallbackRegistry.cpp
@@ -13,43 +13,43 @@ namespace CefSharp
     {
         JavascriptCallbackDto^ JavascriptCallbackRegistry::CreateWrapper(CefRefPtr<CefV8Context> context, CefRefPtr<CefV8Value> value)
         {
-            auto newId = Interlocked::Increment(lastId);
+            auto newId = Interlocked::Increment(_lastId);
             auto wrapper = gcnew JavascriptCallbackWrapper(value, context);
-            callbacks->Add(newId, wrapper);
+            _callbacks->Add(newId, wrapper);
 
             auto result = gcnew JavascriptCallbackDto();
             result->Id = newId;
-            result->BrowserId = browserId;
+            result->BrowserId = _browserId;
             return result;
         }
 
         JavascriptResponse^ JavascriptCallbackRegistry::Execute(Int64 id, array<Object^>^ params)
         {
             JavascriptResponse^ result = nullptr;
-            if (callbacks->ContainsKey(id))
+            if (_callbacks->ContainsKey(id))
             {
-                result = callbacks[id]->Execute(params);
+                result = _callbacks[id]->Execute(params);
             }
             return result;
         }
 
         void JavascriptCallbackRegistry::RemoveWrapper(Int64 id)
         {
-            if (callbacks->ContainsKey(id))
+            if (_callbacks->ContainsKey(id))
             {
-                auto callback = callbacks[id];
-                callbacks->Remove(id);
+                auto callback = _callbacks[id];
+                _callbacks->Remove(id);
                 delete callback;
             }
         }
 
         JavascriptCallbackRegistry::~JavascriptCallbackRegistry()
         {
-            for each (auto callback in callbacks)
+            for each (auto callback in _callbacks)
             {
                 delete callback.Value;
             }
-            callbacks->Clear();
+            _callbacks->Clear();
         }
     }
 }
--- a/CefSharp.BrowserSubprocess.Core/JavascriptCallbackRegistry.cpp
+++ b/CefSharp.BrowserSubprocess.Core/JavascriptCallbackRegistry.cpp
@@ -1,0 +1,55 @@
+// Copyright © 2010-2014 The CefSharp Authors. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style license that can be found in the LICENSE file.
+
+#include "stdafx.h"
+#include "JavascriptCallbackRegistry.h"
+
+using namespace System::Threading;
+
+namespace CefSharp
+{
+    namespace Internals
+    {
+        JavascriptCallbackDto^ JavascriptCallbackRegistry::CreateWrapper(CefRefPtr<CefV8Context> context, CefRefPtr<CefV8Value> value)
+        {
+            auto newId = Interlocked::Increment(lastId);
+            auto wrapper = gcnew JavascriptCallbackWrapper(value, context);
+            callbacks->Add(newId, wrapper);
+
+            auto result = gcnew JavascriptCallbackDto();
+            result->Id = newId;
+            result->BrowserId = browserId;
+            return result;
+        }
+
+        JavascriptResponse^ JavascriptCallbackRegistry::Execute(Int64 id, array<Object^>^ params)
+        {
+            JavascriptResponse^ result = nullptr;
+            if (callbacks->ContainsKey(id))
+            {
+                result = callbacks[id]->Execute(params);
+            }
+            return result;
+        }
+
+        void JavascriptCallbackRegistry::RemoveWrapper(Int64 id)
+        {
+            if (callbacks->ContainsKey(id))
+            {
+                auto callback = callbacks[id];
+                callbacks->Remove(id);
+                delete callback;
+            }
+        }
+
+        JavascriptCallbackRegistry::~JavascriptCallbackRegistry()
+        {
+            for each (auto callback in callbacks)
+            {
+                delete callback.Value;
+            }
+            callbacks->Clear();
+        }
+    }
+}
--- a/CefSharp.BrowserSubprocess.Core/JavascriptCallbackRegistry.cpp
+++ b/CefSharp.BrowserSubprocess.Core/JavascriptCallbackRegistry.cpp
@@ -13,11 +13,11 @@ namespace CefSharp
     {
         JavascriptCallbackDto^ JavascriptCallbackRegistry::CreateWrapper(CefRefPtr<CefV8Context> context, CefRefPtr<CefV8Value> value)
         {
-            auto newId = Interlocked::Increment(_lastId);
-            auto wrapper = gcnew JavascriptCallbackWrapper(value, context);
+            Int64 newId = Interlocked::Increment(_lastId);
+            JavascriptCallbackWrapper^ wrapper = gcnew JavascriptCallbackWrapper(value, context);
             _callbacks->Add(newId, wrapper);
 
-            auto result = gcnew JavascriptCallbackDto();
+            JavascriptCallbackDto^ result = gcnew JavascriptCallbackDto();
             result->Id = newId;
             result->BrowserId = _browserId;
             return result;
@@ -37,7 +37,7 @@ namespace CefSharp
         {
             if (_callbacks->ContainsKey(id))
             {
-                auto callback = _callbacks[id];
+                JavascriptCallbackWrapper^ callback = _callbacks[id];
                 _callbacks->Remove(id);
                 delete callback;
             }
@@ -45,9 +45,9 @@ namespace CefSharp
 
         JavascriptCallbackRegistry::~JavascriptCallbackRegistry()
         {
-            for each (auto callback in _callbacks)
+            for each (JavascriptCallbackWrapper^ callback in _callbacks->Values)
             {
-                delete callback.Value;
+                delete callback;
             }
             _callbacks->Clear();
         }
--- a/CefSharp.BrowserSubprocess.Core/JavascriptCallbackRegistry.h
+++ b/CefSharp.BrowserSubprocess.Core/JavascriptCallbackRegistry.h
@@ -10,14 +10,16 @@ namespace CefSharp
 {
     namespace Internals
     {
-        ref class JavascriptCallbackRegistry
+        private ref class JavascriptCallbackRegistry
         {
         private:
-            int browserId;
-            Int64 lastId;
-            Dictionary<Int64, JavascriptCallbackWrapper^>^ callbacks = gcnew Dictionary<Int64, JavascriptCallbackWrapper^>();
+            int _browserId;
+            Int64 _lastId;
+            Dictionary<Int64, JavascriptCallbackWrapper^>^ _callbacks;
         public:
-            JavascriptCallbackRegistry(int browserId) : browserId(browserId){}
+            JavascriptCallbackRegistry(int browserId) : _browserId(browserId){
+                _callbacks = gcnew Dictionary<Int64, JavascriptCallbackWrapper^>();
+            }
             JavascriptCallbackDto^ CreateWrapper(CefRefPtr<CefV8Context> context, CefRefPtr<CefV8Value> value);
             JavascriptResponse^ Execute(Int64 id, array<Object^>^ params);
             void RemoveWrapper(Int64 id);
--- a/CefSharp.BrowserSubprocess.Core/JavascriptCallbackRegistry.h
+++ b/CefSharp.BrowserSubprocess.Core/JavascriptCallbackRegistry.h
@@ -1,0 +1,28 @@
+// Copyright © 2010-2014 The CefSharp Authors. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style license that can be found in the LICENSE file.
+
+#pragma once
+
+#include "JavascriptCallbackWrapper.h"
+
+namespace CefSharp
+{
+    namespace Internals
+    {
+        ref class JavascriptCallbackRegistry
+        {
+        private:
+            int browserId;
+            Int64 lastId;
+            Dictionary<Int64, JavascriptCallbackWrapper^>^ callbacks = gcnew Dictionary<Int64, JavascriptCallbackWrapper^>();
+        public:
+            JavascriptCallbackRegistry(int browserId) : browserId(browserId){}
+            JavascriptCallbackDto^ CreateWrapper(CefRefPtr<CefV8Context> context, CefRefPtr<CefV8Value> value);
+            JavascriptResponse^ Execute(Int64 id, array<Object^>^ params);
+            void RemoveWrapper(Int64 id);
+            ~JavascriptCallbackRegistry();
+        };
+    }
+}
+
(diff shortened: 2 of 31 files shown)


CefSharp/Internals/JavascriptCallbackSurrogate.cs
```diff
@@ -0,0 +1,71 @@
+﻿using System;
+using System.CodeDom;
+using System.Collections.ObjectModel;
+// Copyright © 2010-2014 The CefSharp Project. All rights reserved.
```
amaitland: Copyright is out of whack.

CefSharp.Wpf.Example/ViewModels/BrowserTabViewModel.cs
```diff
@@ -158,11 +169,13 @@ private void OnWebBrowserLoadError(object sender, LoadErrorEventArgs args)
...
             // Don't display an error for downloaded files where the user aborted the download.
             if (args.ErrorCode == CefErrorCode.Aborted)
+            {
                 return;
+            }
 
             var errorMessage = "<html><body><h2>Failed to load URL " + args.FailedUrl +
-                  " with error " + args.ErrorText + " (" + args.ErrorCode +
-                  ").</h2></body></html>";
+                               " with error " + args.ErrorText + " (" + args.ErrorCode +
```
amaitland: Prefer to keep changes that are purely whitespace out of this for now.

CefSharp.BrowserSubprocess/CefRenderProcess.cs
```diff
@@ -8,20 +8,20 @@
 
 namespace CefSharp.BrowserSubprocess
 {
-    [CallbackBehavior(ConcurrencyMode = ConcurrencyMode.Multiple, IncludeExceptionDetailInFaults=true)]
+    [CallbackBehavior(ConcurrencyMode = ConcurrencyMode.Multiple, IncludeExceptionDetailInFaults = true)]
```
amaitland: Unnecessary whitespace changes.

Prefer if they were reverted.

CefSharp.BrowserSubprocess.Core/JavascriptCallbackWrapper.h
```diff
@@ -0,0 +1,29 @@
...
+
+#include<include\cef_v8.h>
+
+using namespace CefSharp::Internals;
+
+namespace CefSharp
+{
+    namespace Internals
+    {
+        ref class JavascriptCallbackWrapper
```
amaitland: Prefer if we explicitly set the access modifier. (keeps things consistent)

CefSharp.BrowserSubprocess.Core/JavascriptCallbackRegistry.h
```diff
@@ -0,0 +1,28 @@
...
+
+#pragma once
+
+#include "JavascriptCallbackWrapper.h"
+
+namespace CefSharp
+{
+    namespace Internals
+    {
+        ref class JavascriptCallbackRegistry
```
amaitland: Prefer if we explicitly set the access modifier. (keeps things consistent)

CefSharp/Internals/JavascriptCallback.cs
```diff
@@ -0,0 +1,68 @@
...
+        public JavascriptCallback(long id, int browserId, BrowserProcessServiceHost browserProcess)
+        {
+            this.id = id;
+            this.browserId = browserId;
+            _browserProcess = new WeakReference(browserProcess);
+        }
+
+        private BrowserProcessServiceHost BrowserProcessServiceHost
+        {
+            get { return (BrowserProcessServiceHost)_browserProcess.Target; }
```
amaitland: What happens when `Target = null`? Prob better to use `as`. Also no real need for the property, just cast using `as` inline.

amaitland: Basically I'd prefer if it didn't hide the implementation detail, I think it's important to know that there's a weak reference involved.

CefSharp/Internals/JavascriptCallback.cs
```diff
@@ -0,0 +1,68 @@
...
+// Use of this source code is governed by a BSD-style license that can be found in the LICENSE file.
+
+using System;
+using System.Runtime.Serialization;
+using System.Threading.Tasks;
+
+namespace CefSharp.Internals
+{
+    [DataContract]
+    internal sealed class JavascriptCallback : IJavascriptCallback
```
amaitland: To make things a little more `DRY`, should probably inherit `DisposableResource`.

CefSharp/Internals/JavascriptCallbackEndpointBehavior.cs
```diff
@@ -0,0 +1,49 @@
...
+
+using System.Collections.Generic;
+using System.Linq;
+using System.ServiceModel.Channels;
+using System.ServiceModel.Description;
+using System.ServiceModel.Dispatcher;
+
+namespace CefSharp.Internals
+{
+    internal sealed class JavascriptCallbackEndpointBehavior : IEndpointBehavior
```
amaitland: With regards to using an `IEndpointBehavior` to convert to a `DTO`, this becomes inconsistent with the rest of the `JSB` implementation.

Have you used this approach in a production system? Is it rock solid? Any potential memory leaks?

In general I think the idea is interesting, and could possibly be used for all the `WCF` proxy classes. Just need to evaluate of the extra code complexity is worth it?

CefSharp3.sln.DotSettings
```diff
@@ -98,6 +98,7 @@
 	<s:Boolean x:Key="/Default/Environment/SearchAndNavigation/GotoSingleUsageImmediately/@EntryValue">False</s:Boolean>
 	<s:Boolean x:Key="/Default/Environment/SearchAndNavigation/MergeOccurences/@EntryValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002ECSharp_002ECodeStyle_002ESettingsUpgrade_002EMigrateBlankLinesAroundFieldToBlankLinesAroundProperty/@EntryIndexedValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002ECSharp_002ECodeStyle_002ESettingsUpgrade_002EMigrateThisQualifierSettings/@EntryIndexedValue">True</s:Boolean>
```
amaitland: Which particular resharper settings is this for?

amaitland:
I've made a few comments inline, also the `CI` build failed, so that will need to be investigated.

Most of my comments are code style/consistency. Overall my target is to have the project as uniform as possible, which can be a little difficult as style can be such a personal thing. Ideally whitespace changes are reverted at this stage, makes for a much easier code review. If your not familiar then checkout the `Files Changed` tab above you'll have a very easy way to review the changes as a whole.

There's a few things I'd like to discuss further, just need access to `VS` to look into them properly before making further comment.

Overall, fantastic effort! :+1: When I get a chance I'll take it for a proper `test drive`.

CefSharp.BrowserSubprocess.Core/JavascriptCallbackRegistry.h
```diff
@@ -0,0 +1,28 @@
...
+        {
+        private:
+            int browserId;
+            Int64 lastId;
+            Dictionary<Int64, JavascriptCallbackWrapper^>^ callbacks = gcnew Dictionary<Int64, JavascriptCallbackWrapper^>();
```
amaitland: It's likely `callbacks` will need to be initialized in the constructor to avoid the current build errors. (Were you developing on `VS2013`? There are a few differences between that and `VS2012` in terms of `VC++`).

CefSharp.BrowserSubprocess.Core/JavascriptCallbackRegistry.h
```diff
@@ -0,0 +1,28 @@
...
+#include "JavascriptCallbackWrapper.h"
+
+namespace CefSharp
+{
+    namespace Internals
+    {
+        ref class JavascriptCallbackRegistry
+        {
+        private:
+            int browserId;
```
amaitland: `C++` variables are prefixed with `_`.

CefSharp.BrowserSubprocess.Core/JavascriptMethodHandler.cpp
```diff
@@ -0,0 +1,94 @@
...
+        delete _callbackRegistry;
+    }
+
+    bool JavascriptMethodHandler::Execute(const CefString& name, CefRefPtr<CefV8Value> object, const CefV8ValueList& arguments, CefRefPtr<CefV8Value>& retval, CefString& exception)
+    {
+        auto parameter = gcnew array<Object^>(arguments.size());
+
+        for (std::vector<CefRefPtr<CefV8Value>>::size_type i = 0; i != arguments.size(); i++)
+        {
+            parameter[i] = TypeUtils::ConvertFromCef(arguments[i]);
```
amaitland: This line should probably be in an `else` block of the `if` statement below.

CefSharp.BrowserSubprocess.Core/JavascriptCallbackRegistry.cpp
```diff
@@ -0,0 +1,55 @@
...
+            {
+                auto callback = _callbacks[id];
+                _callbacks->Remove(id);
+                delete callback;
+            }
+        }
+
+        JavascriptCallbackRegistry::~JavascriptCallbackRegistry()
+        {
+            for each (auto callback in _callbacks)
```
amaitland: Still seeing a build error https://ci.appveyor.com/project/cefsharp/cefsharp/build/39.0.0-CI758#L11316

From memory `auto` in a `for each` can cause problems when building with `2012`, if we explicitly type `callback` then hopefully the problem will go away.

CefSharp.BrowserSubprocess.Core/JavascriptCallbackRegistry.h
```diff
@@ -0,0 +1,30 @@
...
+    namespace Internals
+    {
+        private ref class JavascriptCallbackRegistry
+        {
+        private:
+            int _browserId;
+            Int64 _lastId;
+            Dictionary<Int64, JavascriptCallbackWrapper^>^ _callbacks;
+        public:
+            JavascriptCallbackRegistry(int browserId) : _browserId(browserId){
```
amaitland: Brace on new line.

amaitland:
Excellent, we have a working build :+1:

CefSharp.BrowserSubprocess.Core/JavascriptCallbackRegistry.h
```diff
@@ -0,0 +1,30 @@
...
+namespace CefSharp
+{
+    namespace Internals
+    {
+        private ref class JavascriptCallbackRegistry
+        {
+        private:
+            int _browserId;
+            Int64 _lastId;
+            Dictionary<Int64, JavascriptCallbackWrapper^>^ _callbacks;
```
amaitland: Should we switch this to a `ConcurrentDictionary`?

arsher:
I think I went through all of the minor things, thanks for pointing them out. It seems like there were a few changes in VS2013 around C++ I wasn't aware of, unfortunately I don't do too much C++ dev :).

A few other things (I'm not sure if I should react inline, so forgive me if that's the usual way) :

> What happens when Target = null? Prob better to use as. Also no real need for the property, just cast using as inline.

Casting a null value will be still null, but the property is indeed unnecessary so I removed it.

> Should we switch this to a ConcurrentDictionary?

Most definitely, but I think I'm doing that tomorrow.

> With regards to using an IEndpointBehavior to convert to a DTO, this becomes inconsistent with the rest of the JSB implementation.
> 
> Have you used this approach in a production system? Is it rock solid? Any potential memory leaks?
> 
> In general I think the idea is interesting, and could possibly be used for all the WCF proxy classes. Just need to evaluate of the extra code complexity is worth it?

I used `IDataContractSurrogate` before for similar purposes, although in this case it's only one-way. First I had an implementation without the `DTO` but to be honest I didn't like the fact that the `JavascriptCallback` class had multiple responsibilities and also I still needed a way to pass an instance of `BrowserProcessServiceHost` to it. I know it looks totally different from the other `WCF` DataContract solutions in the project but I think they're all for just transporting data, they don't have to serve as a point of interaction. I'm not sure this is the best solution though, so I'm open to any ideas :). I couldn't think of a way which doesn't seem a bit hacky.

amaitland:
> Casting a null value will be still null, but the property is indeed unnecessary so I removed it.

I couldn't remember, I out of habit use `as`.

> I'm not sure this is the best solution though, so I'm open to any ideas

Let's think about it a little more and see what we can come up with. Perhaps @rassilon has some ideas on the subject?

rassilon:
> > I'm not sure this is the best solution though, so I'm open to any ideas
> Let's think about it a little more and see what we can come up with. Perhaps @rassilon has some ideas on the subject?

Ok, so it looks like the idea of the `IEndPointBehavior/IDataContractSurrogate` is to auto-provide the deseralized JavaScriptCallbacks with the BrowserServiceHost instance. Right?

However, for this particular set of changes....

Couldn't you just inherit from JavaScriptCallback to have a JavaScriptCallbackBrowser and mark the service host as non-seralizable? Isn't there one location in the BrowserProcess to populate the service host into JavaScriptCallbackBrowser other than via `IEndPointBehavior`?

i.e. I don't even recall if you can serialize IJavaScriptCallback via WCF for example. If so, can't you bind JavaScriptCallbackBrowser in the browser process, but JavaScritpCallback in the render process?

In general notes follow:
It certainly seems like a lot of extra reflection overhead for so little gain in this particular instance.
I certainly could understand you might want to do something like this in a non-greenfield scenario, However, this is essentially a new feature to CefSharp so I would tend to argue that such usage of `IEndpointBehavior` would start with -25 hypothetical feature points or so and you'd use it if it outweighs 25 fake feature points or so.

Additionally, I'd ideally like to eventually replace the WCF mechanism with CEF's (i.e. Chromium's) IPC mechanism.

Hopefully, this will help us understand this approach a bit better.

Thanks a lot for your great work here, btw! This is definitely not a simple set of changes.

Super generalist :smile: ,
Bill

amaitland:
> Additionally, I'd ideally like to eventually replace the WCF mechanism with CEF's (i.e. Chromium's) IPC mechanism.

I'd be on board with replacing `WCF` once `39` is out the door :+1:

CefSharp.Example/Resources/BindingTest.html
```diff
@@ -3,106 +3,122 @@
     <head>
         <title>Binding Test</title>
     </head>
-    <body>
-        <p>
-        Result of calling bound.repeat("hi ", 5) =
+<body>
```
amaitland: Can the formatting be fixed up? It's hard to actually see what's changed.

amaitland:
> It certainly seems like a lot of extra reflection overhead for so little gain in this particular instance.

I hadn't actually considered the reflection overhead. I think overall it would be best to find a simpler solution.

amaitland:
So I've implemented an alternate to `IDataContractSurrogate` which is here https://github.com/amaitland/CefSharp/commit/746e51432acdca5c6adbda7782418dbcce519281

I'm not sure about this approach either, I'm sure there's a bit more that can be cleaned up. Though I'd put it out there for public review first.

Thoughts?

arsher:
Hi,

let me add a few comments for the above.

> Ok, so it looks like the idea of the IEndPointBehavior/IDataContractSurrogate is to auto-provide the deseralized JavaScriptCallbacks with the BrowserServiceHost instance. Right?

That's right. And additionaly the actual call functionality is not used in the render process so it seemed like a mix of responsibilities using the same object on both sides. Although I agree, it's not how the rest of the WCF stuff designed in CefSharp so probably it should be changed.

> Couldn't you just inherit from JavaScriptCallback to have a JavaScriptCallbackBrowser and mark the service host as non-seralizable? Isn't there one location in the BrowserProcess to populate the service host into JavaScriptCallbackBrowser other than via IEndPointBehavior?
> 
> i.e. I don't even recall if you can serialize IJavaScriptCallback via WCF for example. If so, can't you bind JavaScriptCallbackBrowser in the browser process, but JavaScritpCallback in the render process?

You're right, I think inheritance probably the best way to go if we stick to te `IEndpointBehavior` solution. I've chosen that because this way doesn't matter where we pass the JavascriptCallback object the BrowserProcessService gets injected anyway. 
For example I took quick a look at amaitland's alternative which indeed seems to be nicer, although I think (but it's very much possible I missed something) it won't work if a `JavascriptCallback` is returned as the result of a `EvaluateScriptAsync`. I'm going to try to integrate that change and add some additional support to test this.

> If there's a reason frameId isn't stored here, I would think you'd want to comment about it. I know the non-subprocess stuff doesn't currently expose frameIds, but it should in the future since that's how the V8 context is tracked. (i.e. via BrowserId and frameId)

Actually I store the V8 context for each callback so the browserId is only necessary to lookup the `CefBrowserWrapper` instance in `CefRenderProcess`. Which is again not necesserily the best solution but I'll write this up as a comment too.


So i'll take a closer look at 746e51432acdca5c6adbda7782418dbcce519281 and continue the conversation afterwards.

Cheers,
David

amaitland:
> For example I took quick a look at amaitland's alternative which indeed seems to be nicer, although I think (but it's very much possible I missed something) it won't work if a JavascriptCallback is returned as the result of a EvaluateScriptAsync. I'm going to try to integrate that change and add some additional support to test this.

You are correct, I only got the basic example working, my implementation is partially complete. I wasn't completely happy with the solution, so I though it better to open up the discussion before proceeding.

I've had a bit of a rethink overnight, and if we're planning to replace `WCF` in the near future then most of this is moot. So I'm thinking the way forward is.

- Rename `JavascriptCallback` to `JavascriptCallbackProxy`
- Rename `JavascriptCallbackDto` to `JavascriptCallback`
- Add documentation to `JavascriptCallbackProxy` explaining how `WCF` intercepts
- Change `JavascriptCallbackSurrogate` and `JavascriptCallbackEndpointBehavior` to use a `WeakReference` and pass that around instead of the actual reference.
- Expand example to execute callback in a `Task`, demo the `async` benefit.

amaitland:
Changes made in PR to @arsher's branch

https://github.com/arsher/CefSharp/pull/1

Just needs to add some documentation to the `JavascriptCallbackProxy` class explaining how the `Surrogate` works. After that I'm happy to see this merged. (Ideally we should memory profile the changes).

amaitland:
@arsher Any follow comments?

I can scrape the PR against your code and create one directly against `master` if that's easier for you?

amaitland:
I've created #887 which includes all the changes in this `PR` plus my additional ones.

Feel free to comment there. Would be great to get this merged, this week

Issue #870, opened 2015-03-10
Handling 404's


tbowers:
First, how do I know I've hit a bad page? In my testing I see my `LoadError` handler called w/ a code -118 but I wasn't sure if that's the most robust indicator.

Second, I'm if anyone has any pointers to documentation or thoughts on handling 404's or bad URLs in general. The built in WinForms `WebBrowser` gives you the nicely formatted "Page Not Found" page.

Thanks.

amaitland:
Please include full details please, version, flavor, steps to reproduce.

`Contributing.md` contains a basic `Bug Report Template` to make life easier see 
https://github.com/cefsharp/CefSharp/blob/master/CONTRIBUTING.md#what-should-i-include-when-creating-an-issue

It's likely `CefErrorCode` should be marked as `[Flags]`. How do you reproduce the error code you seeing?

tbowers:
I apologize for the lack of information. I wasn't intending to report a bug so much as ping the community for thoughts so I wasn't sure if that information was relevant.

#### Issue Summary
I would like CEF to display an error page when a bad URL is entered similar to Chrome or the `Forms.WebBrowser` control.

![chrome](https://cloud.githubusercontent.com/assets/456638/6598204/16551164-c7d0-11e4-869a-f908407d6ccd.png)
![ie](https://cloud.githubusercontent.com/assets/456638/6598206/183efee0-c7d0-11e4-9ebb-bc1490bdb350.png)

#### Steps to Reproduce
1. Startup CEF WinForms example, type in a bad URL like http://http//www.nowayisthisarealsite.com
2. Note that no page is displayed at all.

#### Technical details
* CefSharp Version: tag v39.0.0-pre03
* Operating System: Windows 7
* x86 
* WinForms

jankurianski:
CEF doesn't do anything with load errors by default (except call the [`OnLoadError` handler](https://github.com/svn2github/chromium-embedded-framework/blob/801006110224ec0c967c3ff89d9c2f82d3162377/cef3/include/cef_load_handler.h#L93-L105)), that's why you're getting the blank page.

Here is a C++ example of how you could load a custom HTML error page upon receiving the `LoadError` event:
https://github.com/svn2github/chromium-embedded-framework/blob/801006110224ec0c967c3ff89d9c2f82d3162377/cef3/tests/cefsimple/simple_handler.cc#L77-L95

amaitland:
[...]

PR #877, opened 2015-03-12
Added many missing error codes to CefErrorCode.cs

jankurianski:
From Chromium's `net/base/net_error_list.h`.

Also added a comment to indicate where the error codes were sourced from.

Also changed copyright date to end in `2015` (should this be done for all files at once, or individually as they are updated? not sure).

Addresses issue #870 

I used the trunk version of `net_error_list.h` for the error codes.  The only interesting thing was that Chromium has added several new certificate errors, so `CertEnd` is now `-213` instead of `-208`, however this should be harmless as `-208` should not be reported by older versions of Chromium anyway.

### DIFF
--- a/CefSharp/CefErrorCode.cs
+++ b/CefSharp/CefErrorCode.cs
@@ -1,4 +1,4 @@
-﻿// Copyright © 2010-2014 The CefSharp Authors. All rights reserved.
+﻿// Copyright © 2010-2015 The CefSharp Authors. All rights reserved.
 //
 // Use of this source code is governed by a BSD-style license that can be found in the LICENSE file.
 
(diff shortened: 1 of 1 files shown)


amaitland:
> Also changed copyright date to end in 2015 (should this be done for all files at once, or individually as they are updated? not sure).

Maybe just as they're updated? Otherwise it's a lot of effort just to change a copyright Year

> I used the trunk version of net_error_list.h for the error codes. The only interesting thing was that Chromium has added several new certificate errors, so CertEnd is now -213 instead of -208, however this should be harmless as -208 should not be reported by older versions of Chromium anyway.

I think that will be fine, there's only so much effort we can spend on something like this.

Merge at will :+1:

jankurianski:
> Maybe just as they're updated? Otherwise it's a lot of effort just to change a copyright Year

Makes sense.

Issue #758, opened 2015-01-22
Issue when CEF is put inside listbox


myrnazhou:
There is a problem when we put ```cefsharp``` wpf control inside a ```listbox```. It makes some keyboard shortcut such as ```ctrl+a``` not working. 
I've pushed it here https://github.com/Bodekaer/CefSharp/tree/CefSharpInsideListBoxTest
Is there a workaround to fix this?

myrnazhou:
And to add more, performance when typing is noticeably slower.

jankurianski:
Perhaps some shortcuts such as CTRL+A are handled by the listbox to select all items in a list?

Have you tried to put more items in the listbox and see if they get selected when you press CTRL+A?

Here is where you could start debugging to see whether the keyboard events are being captured by the WPF control : https://github.com/cefsharp/CefSharp/blob/master/CefSharp.Wpf/ChromiumWebBrowser.cs#L605-L630

amaitland:
If you put a break point at https://github.com/cefsharp/CefSharp/blob/master/CefSharp.Wpf/ChromiumWebBrowser.cs#L894 in `OnPreviewKey` and another at https://github.com/cefsharp/CefSharp/blob/master/CefSharp.Wpf/ChromiumWebBrowser.cs#L620 in the `SourceHook`, you'll find out if `WPF` is handling the key combination.

`OnPreviewKey` will be first, and then the `SourceHook`, if only `OnPreviewKey` is fired then the event was handled elsewhere.

> And to add more, performance when typing is noticeably slower.

Noticeably slower when compared to???

myrnazhou:
@amaitland
when typing between the cefsharp that is inside listbox, and the other which is not.

amaitland:
> when typing between the cefsharp that is inside listbox, and the other which is not.

At a guess if that's the case then it's likely that a total rewrite would be required, one that doesn't rely on a `SourceHook`.

Let us know how the debugging goes :+1:

amaitland:
@myrnazhou Any further feedback on this one?

rassilon:
I haven't tried this, but does CEF send data back out for unhandled events through KeyboardHandler for OSR like we're using in WPF?

If it does, then that might be part of the solution. (I don't recall much about WPF keyboard event propigation.)

[...]

PR #883, opened 2015-03-16
Wpf list box issue

myrnazhou:
Fixing issue when ctrl+A doesn't work when web browser is put in listbox, #758


amaitland:
Merged, Thanks :+1:

PR #887, opened 2015-03-17
Javascript Callbacks - Take 2

amaitland:
Base on #868 

Includes additional changes

- Uses a `ConcurrentDictionary`
- Rename classes
- Pass around `WeakReference` rather than the process host it's self


amaitland:
Still need to add some documentation about the `WCF Surrogate`.

If anyone has time to look over this once more before merging that would be greatly appreciated :smile:

rassilon:
I haven't given this a try, but the code looks fine to me.

PR #888, opened 2015-03-18
Use correct string encoding in CefRequest.Body

jankurianski:
This fixes a bug where unicode characters were not being decoded correctly in the POST request body in CefRequest.

This meant that Unicode text was being corrupted in custom schemes and `RequestHandler`.

For example, U with umlaut - `"Ü"` - that was sent correctly as utf-8 in an AJAX request, was then incorrectly decoded using the default ASCII codepage as `"Ã"`).

### DIFF
--- a/CefSharp.Core/Internals/CefRequestWrapper.cpp
+++ b/CefSharp.Core/Internals/CefRequestWrapper.cpp
@@ -133,7 +133,7 @@ namespace CefSharp
         String^ CefRequestWrapper::CharSet::get()
         {
             // Extract the Content-Type header value.
-            NameValueCollection^ headers = this->Headers;
+            auto headers = this->Headers;
             
             String^ contentType = nullptr;
             for each(String^ key in headers)
--- a/CefSharp.Core/Internals/CefRequestWrapper.cpp
+++ b/CefSharp.Core/Internals/CefRequestWrapper.cpp
@@ -51,6 +51,26 @@ namespace CefSharp
 
                         el->GetBytes(count, bytes);
 
+                        // Attempt to honour the charset specified by the request's Content-Type header.
+                        String^ charset = this->CharSet;
+                        if (charset != nullptr)
+                        {
+                            System::Text::Encoding^ encoding;
+                            try
+                            {
+                                encoding = System::Text::Encoding::GetEncoding(charset);
+                            }
+                            catch (ArgumentException^)
+                            {
+                                encoding = nullptr;
+                            }
+                            if (encoding != nullptr)
+                            {
+                                return gcnew String(bytes, 0, count, encoding);
+                            }
+                        }
+
+                        // Revert to using the system's default code page.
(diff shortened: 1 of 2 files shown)


jankurianski:
This is my first significant C++ change so please do be very picky in your feedback otherwise I won't learn :smile:

amaitland:
Looks solid on first inspection :+1:

I'd probably have put in a using for `System::Text` and you can use `auto` to save yourself typing in cases where it makes sense.

jankurianski:
> I'd probably have put in a using for System::Text and you can use auto to save yourself typing in cases where it makes sense.

Cheers :beers:

`auto` seems the same as C#'s `var` - so I am guessing it makes to use when the type name is long, but not for simple types like `int` or `String^`.

amaitland:
> auto seems the same as C#'s var - so I am guessing it makes to use when the type name is long, but not for simple types like int or String^.

`auto` is exactly the same as `var`. In general it's safe enough to use all over, when you type a lot of `^` characters, you become very grateful for `auto` :smile: 

Is there an easy way to test this? If not I'm happy for you to merge at will as I'm guessing your using this code successfully in your app.

jankurianski:
> Is there an easy way to test this? If not I'm happy for you to merge at will as I'm guessing your using this code successfully in your app.

Ah yes, good point! :smile: 

1. In `CefSharp.Example.RequestHandler`, change `OnBeforeResourceLoad` to:
```
        bool IRequestHandler.OnBeforeResourceLoad(IWebBrowser browser, IRequest request, IResponse response)
        {
            if (request.Url == "https://jquery.com/test")
            {
                Console.WriteLine(request.Body);
            }
            return false;
        }
```
2. Debug `CefSharp.Wpf.Example`
3. Navigate to `https://jquery.com/`
4. In the `Execute Javascript (sychronously)` textbox, paste: `$.post('https://jquery.com/test', 'Ü');`
5. Click Run.
6. In Visual Studio, go to `DEBUG > Windows > Output` and observe that the `Ü` is printed correctly in the Output window.  Before the bug fix, it would have shown `Ãœ`.

jankurianski:
Also thanks for the guidance about `auto` :+1:

amaitland:
Working as advertised :+1: Merge at will :smile:

amaitland:
At some point I'd consider changing `NameValueCollection` returned from `Headers` to a case insensitive version.

e.g. `auto headers = gcnew NameValueCollection(StringComparer::InvariantCultureIgnoreCase);`.

Can then change to `GetValues` or `Get` at some point.

jankurianski:
> At some point I'd consider changing NameValueCollection returned from Headers to a case insensitive version.

Good idea, better than looping through all keys to do the comparison.  Doesn't require a public API change either.  OK to do this for `39`?

amaitland:
> Doesn't require a public API change either. OK to do this for 39?

Whatever suites you :smile:

PR #890, opened 2015-03-18
Make DefaultFocusHandler public class

rassilon:
instead of internal so that you can subclass and reuse the tricky default OnGotFocus implementation.

Add doc comment warnings about the danagers of overriding IFocusHandler.OnGotFocus for WinForm ChromiumWebBrowsers.

Thoughts?
Bill


amaitland:
I remember making the methods `virtual`, guess it would help if the class was `public` :smile:

amaitland:
Merged :+1:

PR #892, opened 2015-03-20
Don't change the .Address property when (among other things) you display...

rassilon:
... DevTools.

Ought to be pretty straightforward.

I wasn't sure if just checking IsPopup was good enough.

The existing test wasn't good enough because DevTools is the main frame for the popup DevTools browser. 

Bill

### DIFF
--- a/CefSharp.Core/Internals/ClientAdapter.cpp
+++ b/CefSharp.Core/Internals/ClientAdapter.cpp
@@ -124,7 +124,11 @@ namespace CefSharp
 
         void ClientAdapter::OnAddressChange(CefRefPtr<CefBrowser> browser, CefRefPtr<CefFrame> frame, const CefString& address)
         {
-            if (frame->IsMain())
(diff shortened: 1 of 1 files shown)


amaitland:
I would have though `!browser->IsPopup()` would have done the trick. Did it not resolve the issue?

rassilon:
IsPopup did indeed work just fine apparently.

amaitland:
Still need the main frame check I believe. Otherwise ifames etc will change the url also.

amaitland:
Looks like the frame check isn't required anymore :+1: From memory it was added for a reason, just can't remember what for. Also possible the behavior of `CEF` has changed. `CEF` usually only passes in a frame object when it's relevant. I tested with regular frames and iframes, and it doesn't seem to be called multiple times anymore. So let's ignore my previous comment.

rassilon:
Cool. Thank's for checking.

PR #893, opened 2015-03-22
Feature/register object using a dictionary

kirilpopov:
It's usefully to be able to register dynamic objects - objects with unknown structure until run-time.

This pull is about adding support for registering a Dictionary object to JS. This enables the above behavior - each key of the dictionary  becomes a property of the new JS object, e.g.

```c#
browser.RegisterJsObject("bound", new Dictionary<string, object>
{
	{"prop1", "value1"},
	{"prop2", new BoundObject()},
});
```

will produce the following in JS

```
bound.prop1 --> value1
bound.prop2 --> boundObject
bound.prop2.MyReadOnlyProperty --> "I'm immutable!"
```

### DIFF
--- a/CefSharp/Internals/JavascriptObjectRepository.cs
+++ b/CefSharp/Internals/JavascriptObjectRepository.cs
@@ -187,26 +187,25 @@ namespace CefSharp.Internals
             if (dict != null)
             {
                 // in case of dictionary, add each key/value pair as read only property of obj                              
-                foreach (var key in dict.Keys)
+                foreach (DictionaryEntry dictionaryEntry in dict)
                 {
-                    var name = key.ToString();
-                    var value = dict[key];                    
+                    var propertyName = dictionaryEntry.Key.ToString();
+                    var propertyValue = dictionaryEntry.Value;                    
 
                     var jsProperty = new JavascriptProperty();
-
-                    jsProperty.ManagedName = name;
-                    jsProperty.JavascriptName = LowercaseFirst(name);
+                    jsProperty.ManagedName = propertyName;
+                    jsProperty.JavascriptName = LowercaseFirst(propertyName);
                     jsProperty.SetValue = (o, v) => { };
-                    jsProperty.GetValue = (o) => value;
+                    jsProperty.GetValue = (o) => propertyValue;
 
-                    jsProperty.IsComplexType = IsComplexType(value.GetType());
+                    jsProperty.IsComplexType = IsComplexType(propertyValue.GetType());
                     jsProperty.IsReadOnly = true;
 
                     if (jsProperty.IsComplexType)
                     {
                         var jsObject = CreateJavascriptObject();
-                        jsObject.Name = name;
-                        jsObject.JavascriptName = LowercaseFirst(name);
+                        jsObject.Name = propertyName;
+                        jsObject.JavascriptName = LowercaseFirst(propertyName);
                         jsObject.Value = jsProperty.GetValue(obj.Value);
                         jsProperty.JsObject = jsObject;
 
--- a/CefSharp/Internals/JavascriptObjectRepository.cs
+++ b/CefSharp/Internals/JavascriptObjectRepository.cs
@@ -291,7 +291,7 @@ namespace CefSharp.Internals
                 baseType = Nullable.GetUnderlyingType(type);
             }
 
-            if (baseType != null && baseType.Name.StartsWith("System.Collections"))
+            if (baseType != null && baseType.Namespace.StartsWith("System.Collections"))
             {
                 return true;
             }
--- a/CefSharp/Internals/JavascriptObjectRepository.cs
+++ b/CefSharp/Internals/JavascriptObjectRepository.cs
@@ -6,6 +6,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
+using System.Collections;
 
 namespace CefSharp.Internals
 {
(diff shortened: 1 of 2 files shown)


CefSharp/Internals/JavascriptObjectRepository.cs
```diff
@@ -182,6 +183,46 @@ private void AnalyseObjectForBinding(JavascriptObject obj, bool analyseMethods,
...
             }
 
+            var dict = obj.Value as IDictionary;
+            if (dict != null)
+            {
+                // in case of dictionary, add each key/value pair as read only property of obj                              
+                foreach (var key in dict.Keys)
+                {
+                    var name = key.ToString();
+                    var value = dict[key];
```
amaitland: Prefer not to use `value` as a variable name, whilst it's not a reserved work, it does have special meaning in some contexts and I'd prefer to avoid using it.

CefSharp/Internals/JavascriptObjectRepository.cs
```diff
@@ -182,6 +183,46 @@ private void AnalyseObjectForBinding(JavascriptObject obj, bool analyseMethods,
                 return;
             }
 
+            var dict = obj.Value as IDictionary;
+            if (dict != null)
+            {
+                // in case of dictionary, add each key/value pair as read only property of obj                              
+                foreach (var key in dict.Keys)
```
amaitland: Can you just perform a straight `foreach` on the dictionary and access it's `KeyValuePair` objects rather than getting the key then performing the lookup?

amaitland:
@kirilpopov Thanks for the PR :+1: I've posted a couple of comments/questions inline. Hopefully someone will have time to review this in more detail shortly

Thanks!

kirilpopov:
@amaitland updated the pull based on your feedback

amaitland:
@kirilpopov Fantastic, thanks :+1:

LeMoussel:
I reported CefSharp/Internals/JavascriptObjectRepository.cs file changed.
I tested with CefSharp.OffScreen.Example in the following manner.

I got this JS Error
`CEF Console:
Uncaught ReferenceError: bound is not defined`

C# Code
-----------
        public static void Main(string[] args)
        {
            const string testUrl = "https://www.google.com/";

            Console.WriteLine("This example application will load {0}, take a screenshot, and save it to your desktop.", testUrl);
            Console.WriteLine("You may see a lot of Chromium debugging output, please wait...");
            Console.WriteLine();

            // You need to replace this with your own call to Cef.Initialize();
            CefExample.Init();

            // Create the offscreen Chromium browser.
            using (browser = new ChromiumWebBrowser(testUrl))
            {

                // An event that is fired when the first page is finished loading.
                // This returns to us from another thread.
                if (captureFirstRenderedImage)
                {
                    browser.ResourceHandlerFactory.RegisterHandler(testUrl, ResourceHandler.FromString("<html><body><h1>CefSharp OffScreen</h1></body></html>"));
                    browser.ScreenshotAsync().ContinueWith(DisplayBitmap);
                }
                else
                {
                    browser.FrameLoadEnd += BrowserFrameLoadEnd;
                    // Output Javascript Console
                    browser.ConsoleMessage += BrowserConsoleMessage;
                }

                // We have to wait for something, otherwise the process will exit too soon.
                Console.ReadKey();
            }

            // Clean up Chromium objects.  You need to call this in your application otherwise
            // you will get a crash when closing.
            Cef.Shutdown();
        }
    
        private static void BrowserConsoleMessage(object sender, ConsoleMessageEventArgs e)
        {
            Console.WriteLine("CEF Console: {0}", e.Message);
        }

        private static void BrowserFrameLoadEnd(object sender, FrameLoadEndEventArgs e)
        {
            // Check to ensure it is the main frame which has finished loading
            // (rather than an iframe within the main frame).
            if (e.IsMainFrame)
            {
                // Remove the load event handler, because we only want one snapshot of the initial page.
                browser.FrameLoadEnd -= BrowserFrameLoadEnd;

                // Wait for the screenshot to be taken.
                browser.ScreenshotAsync().ContinueWith(DisplayBitmap);

                // TEST: Feature/register object using a dictionary #893 
                string JavaScriptResult = "";
                browser.RegisterJsObject("bound", new System.Collections.Generic.Dictionary<string, string>
                {
                    {"prop1", "value1"},
                    {"prop2", "value2"},
                });

                var JavaSCriptTask = browser.EvaluateScriptAsync(@"console.log(bound); ");
                JavaSCriptTask.ContinueWith(t =>
                {
                    if (!t.IsFaulted)
                    {
                        CefSharp.JavascriptResponse response = t.Result;
                        JavaScriptResult = response.Success ? (response.Result.ToString() ?? "null") : response.Message;
                    }
                }).Wait();

            }
        }

amaitland:
If you reread the FAQ, binding has to happen straight after browser init. No support for dynamic binding yet.

amaitland:
For posting large blocks of code it's better to post a [Gist](https://gist.github.com/)

LeMoussel:
**@amaitland** OK. Sorry for the misunderstanding.

amaitland:
Just an FYI, this should be merged sometime next week. Code needs a few minor changes because of #923

rassilon:
Do we care about non-string keys here? i.e. numbers/whatever?
The `.ToString` for the key name and using the non-generic `IDictionary` interface makes me nervous.

If we do this completely, wouldn't any serialize-able object work as the key? And wouldn't we also want to support `IDictionary<TKey, TValue>' instances as well as `IDictionary`?

amaitland:
> If we do this completely, wouldn't any serialize-able object work as the key?

How would that work as the `key` is mapped to the property name in `js`? Maybe we only support key being a string and throw an exception otherwise? Wouldn't stop people from using `"1"`, though we can't guard for everything.

PR #896, opened 2015-03-23
Create ProxyConfig.cs

Jayman1305:
Added the ability to automatically configure the proxy in the CefExample Init class to use the proxy options set in IE. 
To use just call
```c#
var proxy = ProxyConfig.GetProxyInformation();
switch (proxy.AccessType)
{
	case InternetOpenType.Direct:
		//Don't use a proxy server, always make direct connections. Overrides any other proxy server flags that are passed.
		settings.CefCommandLineArgs.Add("no-proxy-server", "1");
		break;
	case InternetOpenType.Proxy:
		settings.CefCommandLineArgs.Add("proxy-server", proxy.ProxyAddress);
		break;
	case InternetOpenType.Preconfig:
		settings.CefCommandLineArgs.Add("proxy-auto-detect", "1");
		break;
}
```

### DIFF
--- a/CefSharp/ProxyConfig.cs
+++ b/CefSharp/ProxyConfig.cs
@@ -1,0 +1,59 @@
+// Copyright © 2010-2014 The CefSharp Authors. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style license that can be found in the LICENSE file.
+
+using System;
+using System.ComponentModel;
+using System.Runtime.InteropServices;
+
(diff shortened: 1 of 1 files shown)


amaitland:
@Jayman1305 I merged this into the `CefSharp.Example` project.

Jayman1305:
@amaitland Thanks for that. I clean forgot about merging it back in Sorry about that.

Issue #598, opened 2014-11-12
WinForms CEFSharp 37.0.0.0 C# X86 (Master) 12-Nov-14 Build Red Scroll Bars?


daveaton:
Hi.

I was testing 33.0.2.0 and then you guys pushed out 37.0.0.0 recently and in my code I just referenced the new dlls just to see if it's going to work with the old code.It seems to work but now I got a weird outcome.

Why are my scroll bars red?  I guess something changed in the code?

http://www.eatoncode.com/shareit/2014-11-12_0426.swf

Any notable changes between 33.0.2.x and 37.0.0.0 ?

This does not happen in the new code. Just Trying to track what changed.

Thanks.

PR #900, opened 2015-03-24
Implement DependencyChecker

### LINKED ISSUE #876: Add DependencyChecker
It's a very common problem for people to be missing dependencies. So I'm thinking we add a simply class that has a list of hard coded dependencies and provide diagnostic information if some are missing. 

Might add the code to the example project, just incase it gets out of sync.

Thoughts?

amaitland:
This is a rough first draft of `DependencyChecker`. Basically loops through list of all Cef/CefSharp dependencies.

Creating this PR before I go too far, some questions to be answered I think.

- Should we provide the option to validate have some dependencies marked as optional?
- What's the expected output if calling from user land? Is a list of strings containing which files are missing ok? Would people just prefer a formatted string?
- Anyone have any thoughts on obtaining the path? I've started with the `ExecutingAssembly` and allowed an overload so that users can call themselves.

Will include an overload for `Cef.Initialize()` that enables/disables the dependency check when I'm happy the other questions have been answered.

Resolves #876

### DIFF
--- a/CefSharp/DependencyChecker.cs
+++ b/CefSharp/DependencyChecker.cs
@@ -90,7 +90,9 @@ namespace CefSharp
                 }
             }
 
-            var localePath = Path.Combine(path, localePackFile);
+            //If path path is not rooted (doesn't start with a drive letter + folder)
+            //then make it relative to the executing assembly.
+            var localePath = Path.IsPathRooted(localePackFile) ? localePackFile : Path.Combine(path, localePackFile);
 
             if (!File.Exists(localePath))
             {
@@ -101,7 +103,7 @@ namespace CefSharp
         }
 
         /// <summary>
-        /// Shortcut method that calls <see cref="CheckDependencies()"/>
+        /// Shortcut method that calls <see cref="CheckDependencies(string)"/>
         /// </summary>
         /// <returns>Returns true of missing dependency count is 0</returns>
         public static bool AreAllDependenciesPresent()
--- a/CefSharp/DependencyChecker.cs
+++ b/CefSharp/DependencyChecker.cs
@@ -11,10 +11,13 @@ using System.Text;
 namespace CefSharp
 {
     /// <summary>
-    /// TODO: Expand to support optional dependencies
+    /// DependencyChecker provides a known list of Cef/CefSharp dependencies and 
+    /// provides helper methods to check for their existance.
     /// </summary>
     public static class DependencyChecker
     {
+        public const string LocalesPackFile = @"locales\en-US.pak";
+
         /// <summary>
         /// List of Cef Dependencies
         /// </summary>
@@ -72,11 +75,11 @@ namespace CefSharp
         /// </summary>
         /// <param name="checkOptional">check to see if optional dependencies are present</param>
         /// <param name="path">path to check for dependencies</param>
-        /// <param name="localePackFile">The locale pack file e.g. locales\en-US.pak</param>
         /// <param name="resourcesDirPath"></param>
         /// <param name="packLoadingDisabled">Is loading of pack files disabled?</param>
+        /// <param name="localePackFile">The locale pack file e.g. <see cref="LocalesPackFile"/> </param>
         /// <returns>List of missing dependencies, if all present an empty List will be returned</returns>
-        public static List<string> CheckDependencies(bool checkOptional, string path, string localePackFile, string resourcesDirPath, bool packLoadingDisabled)
+        public static List<string> CheckDependencies(bool checkOptional, string path, string resourcesDirPath, bool packLoadingDisabled, string localePackFile = LocalesPackFile)
         {
             var missingDependencies = new List<string>();
 
@@ -172,7 +175,7 @@ namespace CefSharp
                 resourcesDirPath = path;
             }
 
-            var missingDependencies = CheckDependencies(true, path, localesDirPath + locale + ".pak", resourcesDirPath, packLoadingDisabled);
+            var missingDependencies = CheckDependencies(true, path, resourcesDirPath, packLoadingDisabled, localesDirPath + locale + ".pak");
 
             if (missingDependencies.Count > 0)
             {
--- a/CefSharp/DependencyChecker.cs
+++ b/CefSharp/DependencyChecker.cs
@@ -13,6 +13,8 @@ namespace CefSharp
     /// </summary>
     public static class DependencyChecker
     {
+        private const string LocalPackPath = @"locales\en-US.pak";
+
         /// <summary>
         /// List of Cef Dependencies - currently contains all possabilties
         /// </summary>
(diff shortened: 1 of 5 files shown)


amaitland:
@jornh @jankurianski @rassilon Any thoughts/comments?

(Full disclaimer, first draft is a little rough).

jankurianski:
> Should we provide the option to validate have some dependencies marked as optional?

Yes, I would like that.  I do not include `pdf.dll` or `ffmpegsumo.dll` with my app as they are unnecessary for me.

Is `CefSharp.BrowserSubprocess.*` required?  Though I have not tried, I thought you could re-use your main .exe somehow.

> What's the expected output if calling from user land? Is a list of strings containing which files are missing ok? Would people just prefer a formatted string?

An exception with a formatted string sounds good to me.  It ensures that an informative message appears in Visual Studio if debugging, and when their app is deployed, it will ensure their exception handling code will log the relevant missing files (as most people will at a minimum log the exception type and message via `e.ToString()`).

> Anyone have any thoughts on obtaining the path? I've started with the ExecutingAssembly and allowed an overload so that users can call themselves.

Default works for me, and overload seems flexible enough to handle anything weird.

Good job BTW :+1:

amaitland:
> Yes, I would like that. I do not include pdf.dll or ffmpegsumo.dll with my app as they are unnecessary for me.

Any thoughts on handling this? There are so many dependencies that are only used in specific scenarios. Is it enough to check for `Required` or `Required + Optional` without going down into the meticulous detail of checking that `d3dcompiler_43.dll` is present for a `windows xp` machine?

> Is CefSharp.BrowserSubprocess.* required? Though I have not tried, I thought you could re-use your main .exe somehow.

`CEF` allows for the scenario where you can use your main exe and launch the `render/gpu` processes via command line options. `CefSharp` explicitly implements `render/gpu` with `BrowserSubprocess`. Quite a common mistake to forget them, which means `Cef` will load without exception, just won't render any content.

amaitland:
Think I've got something that I'm relatively happy with.

Further comments? It's not very well tested yet.

amaitland:
There's likely to be a few refactoring to `DRY` the code out a bit. Keeping it simple until I'm happy we've got something that will work in 99% of cases.

amaitland:
Should also note that disabling pack loading will cause the scrollbars to become red, known issue. #598

CefSharp/DependencyChecker.cs
```diff
@@ -0,0 +1,196 @@
...
+        /// CheckDependencies iterates through the list of Cef and CefSharp dependencines
+        /// relative to the path provided and returns a list of missing ones
+        /// </summary>
+        /// <param name="checkOptional">check to see if optional dependencies are present</param>
+        /// <param name="packLoadingDisabled">Is loading of pack files disabled?</param>
+        /// <param name="path">path to check for dependencies</param>
+        /// <param name="resourcesDirPath"></param>
+        /// <param name="localePackFile">The locale pack file e.g. <see cref="LocalesPackFile"/> </param>
+        /// <returns>List of missing dependencies, if all present an empty List will be returned</returns>
+        public static List<string> CheckDependencies(bool checkOptional, bool packLoadingDisabled, string path, string resourcesDirPath, string localePackFile = LocalesPackFile)
```
rassilon: Whenever I see gobs of bool parameters I usually think: Are you really sure using a flag enum isn't the correct thing to do here?

CefSharp/DependencyChecker.cs
```diff
@@ -0,0 +1,196 @@
...
+            {
+                localesDirPath = @"locales\";
+            }
+
+            if (string.IsNullOrEmpty(resourcesDirPath))
+            {
+                resourcesDirPath = path;
+            }
+
+            var missingDependencies = CheckDependencies(true, packLoadingDisabled, path, resourcesDirPath, localesDirPath + locale + ".pak");
```
rassilon: Path.Combine here I think.

CefSharp/DependencyChecker.cs
```diff
@@ -0,0 +1,196 @@
...
+            "cef_200_percent.pak"
+        };
+
+        public static string[] CefOptionalDependencies =
+        {
+            // Angle and Direct3D support
+            // Note: Without these components HTML5 accelerated content like 2D canvas, 3D CSS and WebGL will not function.
+            "libEGL.dll",
+            "libGLESv2.dll",
+            "d3dcompiler_43.dll",
```
rassilon: Even thought we don't care yet, we should probably comment about the XP vs Vista+ thing here.

rassilon:
This is looking good. I think the optional dlls fall into these separate catagories:
* GPUAcceleration
* OSGPUAcceleration
* AudioVideo
* PDFPlugin

The flag enum idea would allow you to easily check which set to examine.

I don't really care at the moment whether or not the above suggestions are put in the PR before merging though.

Bill

CefSharp/DependencyChecker.cs
amaitland: I guess so, I'm tempted to remove `43` as we don't support `XP` officially.

CefSharp/DependencyChecker.cs
amaitland: Pack loading is a special case. `checkOptional` would be replaced with an enum if we went down that path.

CefSharp/DependencyChecker.cs
amaitland: Thanks :+1:

amaitland:
I've added a slightly crude `IsWindowsXp` static property, not great. Will do for now I guess. Also moved the code into `Cef.Initialize()`, it's not called by default as yet.  Expanded the example to demo it.

rassilon:
Still looking good.

rassilon:
If we eventually do the flag thing might as well include:
* WinForm
* WPF
* Generic OSR

Bill

CefSharp/DependencyChecker.cs
```diff
@@ -0,0 +1,200 @@
...
+                var builder = new StringBuilder();
+                builder.AppendLine("Unable to locate required Cef/CefSharp dependencies:");
+
+                foreach (var missingDependency in missingDependencies)
+                {
+                    builder.AppendLine("Missing:" + missingDependency);
+                }
+
+                builder.AppendLine("Executing Assembly Path:" + path);
+
```
rassilon: We could also add an informational reminder in this string about locales\xxxx.pak that explains what they do if we know even though I would never suggest checking for the dependency to exist.

amaitland: @rassilon Not sure I understand, can you clarify?

amaitland:
Anyone up for taking this for a spin and reporting back? 

At some point I'd still like to see the `enum` option implemented, though I'm thinking this might be enough for `39` and we improve for `41`?

amaitland:
Merging this one now, if anyone has improvements they wish to make just shout out so we can work out if they are done before `39` release.

PR #904, opened 2015-03-26
Fix #884 by adding sourceUrl parameter to OnBeforePopup .Net callback.

### LINKED ISSUE #884: C# Winforms .Net 4.0  How to pass the Referer Url to a new popup ?
Is there a way to look-up the referrer information to pass to a new popup?

I can see from the Master default windows when you open a new popup the referrer information is passed to the new popup. I can't see where it's doing this.  

Here is a demo of this.

http://www.eatoncode.com/shareit/2015-03-16_2150.swf

I was expecting to get the referrer information from this method

public bool OnBeforePopup(IWebBrowser browser, string url, ref int x, ref int y, ref int width, ref int height)

But I did not see any values for Referrer.  I am trying to use my own custom version of the popup window instead of using the default hard coded windows.

I could simply pass the url it came from [Browser.Url] but if it's done from frames I will incorrectly set the wrong referrer if the link actually came from a sub inner frame.

Any ideas ?

rassilon:
Here's the change we discussed in #884.

Bill

### DIFF
--- a/CefSharp.Core/Internals/ClientAdapter.cpp
+++ b/CefSharp.Core/Internals/ClientAdapter.cpp
@@ -66,7 +66,7 @@ namespace CefSharp
                 return false;
             }
 
-            return handler->OnBeforePopup(_browserControl, StringUtils::ToClr(target_url),
+            return handler->OnBeforePopup(_browserControl, StringUtils::ToClr(frame->GetURL()), StringUtils::ToClr(target_url),
                 windowInfo.x, windowInfo.y, windowInfo.width, windowInfo.height);
         }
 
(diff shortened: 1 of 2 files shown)


amaitland:
I prefer the name `sourceUrl` over `referrerUrl` :+1: 

We probably should add some documentation, though as there isn't any as it stands I'm not too worried.

Merge away :smile:

rassilon:
@daveaton, CefSharp is pretty straightforward, code doesn't get more difficult to follow until you start looking inside of CEF/Chromium.  (but honestly, that's mostly Chromium's fault) :smile: 

Bill

PR #911, opened 2015-03-28
Cleanup OnBeforeResourceLoad 

amaitland:
Remove `IResponse` and `Response` classes, to perform Redirect simply set the `IRequest.Url`

Was reading the code, answering a question and `OnBeforeResourceLoad` seemed overly complicated (I know it was leftover from `CefSharp1`, when you could return a stream to fulfill a request).

This is a breaking change. I've updated the example to demo how this should work.


amaitland:
Unless anyone has objections to this simplification, I'll merge during the week.

Issue #909, opened 2015-03-27
Bug: IsLoaded is unreliable until we've migrated to use OnLoadingStateChange instead


perlun:
Hi,

I'm experiencing issues with <subj>. This is with CefSharp 37.0.1 (btw, should publish it from MyGet...).

I can reproduce the issue with the WPF example as well. If I open up http://www.google.com, and then type in something, and press Enter or click Search, the progress bar (bottom) will keep rolling, like this:

![image](https://cloud.githubusercontent.com/assets/630613/6869389/3c201ea6-d49c-11e4-86f0-4f80978f3d31.png)

Anyone feel like investigating? Has it already been fixed in the latest `master` mayhaps?

amaitland:
I am aware of this one, `IsLoading` relies on `OnLoadStart` and `OnLoadEnd` which leads to this http://magpcss.org/ceforum/viewtopic.php?f=6&t=10906 post which should look familiar :smile: 

Relevant code
https://github.com/cefsharp/CefSharp/blob/master/CefSharp.Wpf/ChromiumWebBrowser.cs#L321
https://github.com/cefsharp/CefSharp/blob/master/CefSharp.Wpf/ChromiumWebBrowser.cs#L330

There's also this https://bitbucket.org/chromiumembedded/cef/issue/1050/cef3-onloadend-not-always-called-for

amaitland:
Just as a followup http://magpcss.org/ceforum/apidocs3/projects/%28default%29/CefLoadHandler.html#OnLoadStart%28CefRefPtr%3CCefBrowser%3E,CefRefPtr%3CCefFrame%3E%29

>  For notification of overall browser load status use OnLoadingStateChange instead. 

`OnLoadStart` documentation actually now suggests using `OnLoadingStateChange` instead.

perlun:
> I am aware of this one, IsLoading relies on OnLoadStart and OnLoadEnd which leads to this http://magpcss.org/ceforum/viewtopic.php?f=6&t=10906 post which should look familiar :smile:

Hehe, nice! :wink:

I think we should migrate over to using `OnLoadingStateChange` instead. Let's leave this issue open until we've done that. OK with you?

amaitland:
Care to take #913 for a spin? Change only implemented in the context of `WPF` for now. If that solves the problem then we can look at the other flavors :smile:

amaitland:
[...]

PR #913, opened 2015-03-30
Resolve: IsLoaded is unreliable

amaitland:
http://magpcss.org/ceforum/apidocs3/projects/%28default%29/CefLoadHandler.html#OnLoadStart%28CefRefPtr%3CCefBrowser%3E,CefRefPtr%3CCefFrame%3E%29

`For notification of overall browser load status use OnLoadingStateChange instead.`

Initially just migrating `WPF` control. When it's configured, update the other two flavours.

### DIFF
--- a/CefSharp.OffScreen/ChromiumWebBrowser.cs
+++ b/CefSharp.OffScreen/ChromiumWebBrowser.cs
@@ -66,6 +66,7 @@ namespace CefSharp.OffScreen
         public event EventHandler<StatusMessageEventArgs> StatusMessage;
         public event EventHandler<NavStateChangedEventArgs> NavStateChanged;
         public event EventHandler<AddressChangedEventArgs> AddressChanged;
+        [Obsolete("IsLoadingChanged is unreliable and will be removed. Use NavStateChanged instead.")]
         public event EventHandler<IsLoadingChangedEventArgs> IsLoadingChanged;
 
         /// <summary>
--- a/CefSharp.WinForms/ChromiumWebBrowser.cs
+++ b/CefSharp.WinForms/ChromiumWebBrowser.cs
@@ -232,6 +232,7 @@ namespace CefSharp.WinForms
         public event EventHandler<AddressChangedEventArgs> AddressChanged;
         public event EventHandler<TitleChangedEventArgs> TitleChanged;
         public event EventHandler<IsBrowserInitializedChangedEventArgs> IsBrowserInitializedChanged;
+        [Obsolete("IsLoadingChanged is unreliable and will be removed. Use NavStateChanged instead.")]
         public event EventHandler<IsLoadingChangedEventArgs> IsLoadingChanged;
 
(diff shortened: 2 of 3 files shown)


amaitland:
When complete this will fix #909

rassilon:
I can push this through WinForms separately if you like.
Looks like it might be the right thing, but I haven't done much WPF programming.

Bill

rassilon:
Oh, If we do this push this through, we should add a depreciated attribute to `SetIsLoading` or something similar.

Bill

amaitland:
> Oh, If we do this push this through, we should add a depreciated attribute to SetIsLoading or something similar.

Hmm, yeah maybe making the `IsLoadingChanged` event as deprecated is the way to go for now. For `WPF` it makes sense to have a bind able property. For the others I think removing the event is better long term, people can still subscribe to `FrameLoadStart` and `FrameLoadEnd` which is would give the same output if absolutely nessicary. Otherwise `NavStateChanged` seems to be the way to go.

Looking at `OnLoadStart` and `OnLoadEnd`, I wonder if we still need the `AutoLock lock_scope(_syncRoot);`? I'm not actually sure what their purposes is. For all three flavors, they simply fire events, so I wouldn't have though locking was necessary. 

So I think the short term question is, do we remove `IsLoadingChanged` from `WinForms` and `OffScreen` or mark them as deprecated?

amaitland:
If we remove `IsLoadingChanged` then `NavStateChanged` could be renamed to `LoadingStateChanged` to bring it inline with the `CEF Api`. Thoughts?

rassilon:
I think the `lock_scope(_syncRoot)` bit might be trying to prevent the users event callbacks being called simultaneously on two different threads.

Well, at a minimum depreciated I think, although if you're ok with the renaming to `LoadingStateChanged` for 41.x that's fine by me, it certainly is a cleaner naming.

If we were going to rename in 41.x I think it would be ok to depreciate in 39.0.1 (or whatever) if we wanted.

Bill

amaitland:
> I think the lock_scope(_syncRoot) bit might be trying to prevent the users event callbacks being called simultaneously on two different threads.

As they basically just fire events, I wonder if that's necessary? My initial though would be that locking maybe slowing things down, by how much I'm not sure. 

> If we were going to rename in 41.x I think it would be ok to depreciate in 39.0.1 (or whatever) if we wanted.

Sounds like a plan :+1:

rassilon:
I don't know if the lock is necessary or not, its certainly nicer to the consumer of CefSharp if you can have the two events fire on different threads from CEF and not have to care about it at the level of CefSharp's users.

amaitland:
I think if we remove the call to `SetIsLoading` which would be part of this PR then it should be safe enough. If we change in `2272` early enough then hopefully any issues will arise before the release of `41.0.0`

amaitland:
#931 has changes for `2272` branch.

Just a matter of making the `IsLoadingChanged` as deprecated.

Issue #906, opened 2015-03-26
select alignment issue


svantreeck:
Hi, I've been having an issue where select tags seem to have their dropdowns misaligned using bootstrap 3.  The code below seems to illustrate this in the Wpf Example project.  Any suggestions on fixing this would be appreciated.  Thanks!

```
<!DOCTYPE html>
<html>
<head lang="en">
    <meta charset="UTF-8">
    <title>Sample</title>
    <!-- Latest compiled and minified CSS -->
    <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.4/css/bootstrap.min.css">

    <!-- Optional theme -->
    <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.4/css/bootstrap-theme.min.css">


</head>
<body>
    <div class="container">
        <div class="row">
            <div class="col-lg-4 form-group">
                <label class="control-label">
                    Test Select
                </label>
                <select class="form-control">
                    <option value="0" label="0">0</option>
                    <option value="1" label="1">1</option>
                    <option value="2" label="2">2</option>
                    <option value="3" label="3">3</option>
                    <option value="4" label="4">4</option>
                </select>
            </div>
        </div>
    </div>

    <script src="http://code.jquery.com/jquery-2.1.3.min.js" />
    <!-- Latest compiled and minified JavaScript -->
    <script src="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.4/js/bootstrap.min.js"></script>
</body>
</html>
```

amaitland:
Can you please confirm the versions you've tested with. Also html code is best posted as a fiddler link.

amaitland:
(Should have been `jsfiddle` link).

amaitland:
- Can you provide a `jsfiddle` link that reproduces the problem
- Does the problem exist in the `WinForms` example?
- Are you using a `DPI > 1`?
- What versions have you tested with?
- Is it just bootstrap dropdowns that are the problem?

svantreeck:
Here's a JSFiddle:
https://jsfiddle.net/4be895dw/

I cloned the Wpf.Example project to reproduce it there (using cef 3.2171.1979) on the 24th, using DPI=1.  I'll rerun using the winform example tomorrow when I have the computer again.  

[...]

PR #916, opened 2015-03-30
WPF Popup - Change to absolute positioning

svantreeck:
This avoids win 8/8.1 menu alignment settings where they can cause devices to not show relative positioned popups where expected.  

This fixes issue #906

### DIFF
--- a/CefSharp.Wpf/ChromiumWebBrowser.cs
+++ b/CefSharp.Wpf/ChromiumWebBrowser.cs
@@ -929,7 +929,7 @@ namespace CefSharp.Wpf
             {
                 Child = popupImage = CreateImage(),
                 PlacementTarget = this,
-                Placement = PlacementMode.Relative,
+                Placement = PlacementMode.Absolute,
             };
 
             newPopup.MouseEnter += PopupMouseEnter;
@@ -1068,8 +1068,11 @@ namespace CefSharp.Wpf
             popup.Height = height;
 
             var popupOffset = new Point(x, y);
-            popup.HorizontalOffset = popupOffset.X / matrix.M11;
-            popup.VerticalOffset = popupOffset.Y / matrix.M22;
+            var locationFromScreen = this.PointToScreen(popupOffset);
+            var targetPoint =
+                PresentationSource.FromVisual(this).CompositionTarget.TransformFromDevice.Transform(locationFromScreen);
+            popup.HorizontalOffset = targetPoint.X;
+            popup.VerticalOffset = targetPoint.Y;
         }
 
         private void OnTooltipTimerTick(object sender, EventArgs e)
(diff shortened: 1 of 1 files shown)


CefSharp.Wpf/ChromiumWebBrowser.cs
```diff
@@ -1068,8 +1068,11 @@ private void SetPopupSizeAndPositionImpl(int width, int height, int x, int y)
             popup.Height = height;
 
             var popupOffset = new Point(x, y);
-            popup.HorizontalOffset = popupOffset.X / matrix.M11;
-            popup.VerticalOffset = popupOffset.Y / matrix.M22;
+            var locationFromScreen = this.PointToScreen(popupOffset);
+            var targetPoint =
```
amaitland: Use `matrix` to perform the transform.

CefSharp.Wpf/ChromiumWebBrowser.cs
```diff
@@ -1068,8 +1068,11 @@ private void SetPopupSizeAndPositionImpl(int width, int height, int x, int y)
             popup.Height = height;
 
             var popupOffset = new Point(x, y);
-            popup.HorizontalOffset = popupOffset.X / matrix.M11;
-            popup.VerticalOffset = popupOffset.Y / matrix.M22;
+            var locationFromScreen = this.PointToScreen(popupOffset);
```
amaitland: Remove redundant `this` qualifier.

amaitland:
@svantreeck Couple of minor changes, apart from that looks solid :+1: Thanks for the contribution!

svantreeck:
I updated the code...tested and working on windows 10.  I'll check it out on a few other boxes (8.1 and 7) tomorrow morning to double check things look ok there because things didn't look correct with DPI scaling when I first turned it tried the code in Win10, so now that its right there, I very possibly messed up on 7/8/8.1

amaitland:
> I very possibly messed up on 7/8/8.1

No problem, `DPI` scaling can be fiddly to get working in all scenarios. Let us know how you go :+1:

svantreeck:
It looks fine to me on win 8.1, I think its good to go unless you can see any issues with it.

amaitland:
It will need testing using a range of scenarios. Does it work under win 7?

amaitland:
Looks like the `width/height` also needs to be scaled.

CefSharp.Wpf/ChromiumWebBrowser.cs
```diff
@@ -1067,9 +1067,10 @@ private void SetPopupSizeAndPositionImpl(int width, int height, int x, int y)
             popup.Width = width;
```
amaitland: I think we need to scale the width and height from the results I'm seeing on Win7 at 1.25

svantreeck:
I would second that...I have a screen with 150% scaling and was getting an ugly black background around the edges.  Commited an update to scale the width and height by the matrix.

amaitland:
> Commited an update to scale the width and height by the matrix.

Great :+1: 

Can you give a quick summary of all the combinations you've tested with so far? Basically looks like it's working now, just need to make sure we haven't missed anything before committing.

svantreeck:
I've tested on:
Windows 8.1 with 100% DPI scaling
Windows 10 with 100% and 150% DPI scaling
Windows Server 2008 R2 with 100% DPI scaling

Issue #922, opened 2015-04-01
Javascript exposed .NET functions lowercase


exinferis:
Since my last attempt to post this issue was a bit...impulsive (sorry about that), I try again.

All Functions in C#/.NET which I expose towards Javascript are made lowercase, which in our case prevents Integration of CefSharp into our project.

What is the background of this and how can this be prevented? I do not have the possibility to change the JS, so I need to expose Uppercase Functions within an object towards JS. 

My class is looking like this:

```C#
public class JavascriptInterface
{
  public void TokenCreated() {
    // Ultra fancy stuff here
  }
}
```

Within Javascript it will look like this:

```Javascript
window.JSInterface.tokenCreated()
```

avspeed:
you might want to look at https://github.com/cefsharp/CefSharp/blob/9fe07e50cedde2d8b8c17d2d081118c16ed11ae1/CefSharp/Internals/JavascriptObjectRepository.cs

mores specifically this 

jsMethod.JavascriptName = LowercaseFirst(methodInfo.Name);

amaitland:
> What is the background of this and how can this be prevented? I do not have the possibility to change the JS, so I need to expose Uppercase Functions within an object towards JS.

The simplest approach would be to create an overload for `RegisterJsObject(string name, object objectToBind, bool lowerCaseJavascriptNames)`, from there it's just a matter of passing through the chain until it get to `JavascriptObjectRepository.Register`. Another option would be to create an `Attribute` to decorate your class, this has a few benefits for allowing other configuration options, and having them on a per class basis. I think for now the simple approach of another param is the best option.

Quickest way to make this happen is if you submit a `PR`.

> I am not sure why this was done this way .. maybe cef is excepting it to be lower case.

It's just a convention, not strictly required. It was the default for `CefSharp1`, so it just got carried over to `CefSharp3`.

perlun:
This is by design, done by me. Rationale: It should be possible to write your .NET code using ".NET style", i.e. `Class.SomeMethodName()` while still exposing it JS style in the JS code. `class.someMethodName()`.

But making it parameterized or controlled by a .NET attribute is of course a way to work around it.

PR #923, opened 2015-04-02
Added parameter for lowercasing JS functions (or not)

exinferis:
Used to give the dev the possibility to decide wether JS methods/objects/parameters should be automatically made first letter lowercase or not.

### DIFF
--- a/CefSharp.Core/ManagedCefBrowserAdapter.h
+++ b/CefSharp.Core/ManagedCefBrowserAdapter.h
@@ -661,9 +661,9 @@ namespace CefSharp
             }
         }
 
-		void RegisterJsObject(String^ name, Object^ object, bool lowerCaseJavascriptNames)
+        void RegisterJsObject(String^ name, Object^ object, bool lowerCaseJavascriptNames)
         {
-			_javaScriptObjectRepository->Register(name, object, lowerCaseJavascriptNames);
+            _javaScriptObjectRepository->Register(name, object, lowerCaseJavascriptNames);
         }
 
         void ReplaceMisspelling(String^ word)
--- a/CefSharp.Core/ManagedCefBrowserAdapter.h
+++ b/CefSharp.Core/ManagedCefBrowserAdapter.h
@@ -661,9 +661,9 @@ namespace CefSharp
             }
         }
 
-        void RegisterJsObject(String^ name, Object^ object)
+		void RegisterJsObject(String^ name, Object^ object, bool lowerCaseJavascriptNames)
         {
-            _javaScriptObjectRepository->Register(name, object);
+			_javaScriptObjectRepository->Register(name, object, lowerCaseJavascriptNames);
         }
 
         void ReplaceMisspelling(String^ word)
--- a/CefSharp.OffScreen/ChromiumWebBrowser.cs
+++ b/CefSharp.OffScreen/ChromiumWebBrowser.cs
@@ -306,7 +306,11 @@ namespace CefSharp.OffScreen
 
         public void RegisterJsObject(string name, object objectToBind)
         {
-            managedCefBrowserAdapter.RegisterJsObject(name, objectToBind);
+            RegisterJsObject(name, objectToBind, true);
+        }
+        public void RegisterJsObject(string name, object objectToBind, bool lowerCaseJavascriptNames)
+        {
+            managedCefBrowserAdapter.RegisterJsObject(name, objectToBind, lowerCaseJavascriptNames);
         }
 
         public void Stop()
--- a/CefSharp.OffScreen/ChromiumWebBrowser.cs
+++ b/CefSharp.OffScreen/ChromiumWebBrowser.cs
@@ -304,11 +304,7 @@ namespace CefSharp.OffScreen
             Load(url);
         }
 
-        public void RegisterJsObject(string name, object objectToBind)
-        {
-            RegisterJsObject(name, objectToBind, true);
-        }
-        public void RegisterJsObject(string name, object objectToBind, bool lowerCaseJavascriptNames)
+        public void RegisterJsObject(string name, object objectToBind, bool lowerCaseJavascriptNames = true)
         {
             managedCefBrowserAdapter.RegisterJsObject(name, objectToBind, lowerCaseJavascriptNames);
         }
--- a/CefSharp.WinForms/ChromiumWebBrowser.cs
+++ b/CefSharp.WinForms/ChromiumWebBrowser.cs
@@ -198,9 +198,13 @@ namespace CefSharp.WinForms
             Load(url);
         }
 
-        public void RegisterJsObject(string name, object objectToBind)
+        public void RegisterJsObject(string name, object objectToBind) 
         {
-            managedCefBrowserAdapter.RegisterJsObject(name, objectToBind);
+            RegisterJsObject(name, objectToBind, true);
+        }
+        public void RegisterJsObject(string name, object objectToBind, bool lowerCaseJavascriptNames)
+        {
+            managedCefBrowserAdapter.RegisterJsObject(name, objectToBind, lowerCaseJavascriptNames);
         }
 
         public void ExecuteScriptAsync(string script)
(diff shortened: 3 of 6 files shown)


exinferis:
This is the suggested fix for #922

CefSharp.Core/ManagedCefBrowserAdapter.h
```diff
@@ -661,9 +661,9 @@ namespace CefSharp
             }
         }
 
-        void RegisterJsObject(String^ name, Object^ object)
+		void RegisterJsObject(String^ name, Object^ object, bool lowerCaseJavascriptNames)
```
amaitland: Spaces not tabs see https://github.com/cefsharp/CefSharp/blob/master/CONTRIBUTING.md#coding-style

CefSharp.OffScreen/ChromiumWebBrowser.cs
```diff
@@ -306,7 +306,11 @@ public void LoadHtml(string html, string url, Encoding encoding)
 
         public void RegisterJsObject(string name, object objectToBind)
         {
-            managedCefBrowserAdapter.RegisterJsObject(name, objectToBind);
+            RegisterJsObject(name, objectToBind, true);
+        }
+        public void RegisterJsObject(string name, object objectToBind, bool lowerCaseJavascriptNames)
```
amaitland: Missing line break between functions.

CefSharp.WinForms/ChromiumWebBrowser.cs
```diff
@@ -198,9 +198,13 @@ public void LoadHtml(string html, string url, Encoding encoding)
             Load(url);
         }
 
-        public void RegisterJsObject(string name, object objectToBind)
+        public void RegisterJsObject(string name, object objectToBind) 
         {
-            managedCefBrowserAdapter.RegisterJsObject(name, objectToBind);
+            RegisterJsObject(name, objectToBind, true);
+        }
+        public void RegisterJsObject(string name, object objectToBind, bool lowerCaseJavascriptNames)
```
amaitland: Missing line break between functions.

CefSharp/Internals/JavascriptObjectRepository.cs
```diff
@@ -35,14 +35,15 @@ internal JavascriptObject CreateJavascriptObject()
             return result;
         }
 
-        public void Register(string name, object value)
+        public void Register(string name, object value) {  Register(name, value, true); }
```
amaitland: Remove unused method signature. As there's not way to call this publicly, the overload here is not required.
(Also the formatting is out, need to follow existing styling in general)

CefSharp/Internals/JavascriptObjectRepository.cs
```diff
@@ -169,7 +170,12 @@ public bool TrySetProperty(long objectId, string name, object value, out string
         /// <param name="obj">Javascript object</param>
         /// <param name="analyseMethods">Analyse methods for inclusion in metadata model</param>
         /// <param name="readPropertyValue">When analysis is done on a property, if true then get it's value for transmission over WCF</param>
-        private void AnalyseObjectForBinding(JavascriptObject obj, bool analyseMethods, bool readPropertyValue)
+        /// <param name="lowerCaseJavascriptNames">decide wether your JS mathods and properties are automatically lowercased or not</param>
+        private void AnalyseObjectForBinding(JavascriptObject obj, bool analyseMethods, bool readPropertyValue)
```
amaitland: Remove extra method signature, not required internally.

CefSharp/Internals/JavascriptObjectRepository.cs
```diff
@@ -204,12 +210,19 @@ private void AnalyseObjectForBinding(JavascriptObject obj, bool analyseMethods,
...
+                var jsProperty = CreateJavaScriptProperty(propertyInfo, lowerCaseJavascriptNames);
                 if (jsProperty.IsComplexType)
                 {
                     var jsObject = CreateJavascriptObject();
                     jsObject.Name = propertyInfo.Name;
-                    jsObject.JavascriptName = LowercaseFirst(propertyInfo.Name);
+
+                    // depending on lowerCaseJavascriptNames JS Properties will be lowercase or unaltered 
+                    // defaults to LowercaseFirst, in order to not change existing behaviour
+                    if (lowerCaseJavascriptNames)
```
amaitland: If statements need braces. In this case, even for one liners.

In this case I think it's cleaner to pass in the param to `LowercaseFirst` and have it decide. (potentially rename `LowercaseFirst` to be more descriptive given it's additional responsibility).

CefSharp/Internals/JavascriptObjectRepository.cs
```diff
@@ -223,12 +236,21 @@ private void AnalyseObjectForBinding(JavascriptObject obj, bool analyseMethods,
             }
         }
 
-        private static JavascriptMethod CreateJavaScriptMethod(MethodInfo methodInfo)
+        private static JavascriptMethod CreateJavaScriptMethod(MethodInfo methodInfo)
```
amaitland: Remove extra method signature, not required internally.

CefSharp/Internals/JavascriptObjectRepository.cs
```diff
@@ -237,10 +259,19 @@ private static JavascriptMethod CreateJavaScriptMethod(MethodInfo methodInfo)
 
         private static JavascriptProperty CreateJavaScriptProperty(PropertyInfo propertyInfo)
```
amaitland: Remove extra method signature, not required internally.

amaitland:
Comments inline, needs a bit of refactoring and cleanup. Nothing major though.

amaitland:
Thinking about this a little more, might be worth just using an `optional param` instead of overloading, keeps the lines of code down as they're mostly just wrapper methods until they get passed for analysis.

Also I think `IWebBrowser` probably needs to be updated to include the new signature for whatever we go with.

exinferis:
Shall I change my pull request to fulfill the coding guidelines etc, or would you suggest to implement the alternative solution using optional parameters?

amaitland:
I think optional params makes sense. What do you think?

exinferis:
I do not know the internals of the project well enough to really be sure, but I also think that an optional parameter makes sense.
Why and where needs the ```IWebBrowser```to be updated in that case?

amaitland:
It's only the external methods that require optional params, specifically those on the `ChromiumWebBrowser` instances, all other methods just need the value passed.

The `IWebBrowser.RegisterJsObject` method will need to be updated including xml docs.
https://github.com/cefsharp/CefSharp/blob/master/CefSharp/IWebBrowser.cs#L89

perlun:
Coming in a bit late here, but just a suggestion: `lowercaseMethodNames` isn't really a proper name, since it's not about "lower-casing" per se, but rather "javascript-casing". So "javascriptCaseMethodNames" or something to the like would make more sense, IMHO. What do you think @amaitland?

perlun:
(just so we keep the public API clean)

amaitland:
> So "javascriptCaseMethodNames" or something to the like would make more sense,

Hadn't really though about it :smile:  As it applies to both properties as well, it should probably have a more generic name. I'm all for improving variable naming as a general thing :+1: 

@perlun Do you have time to submit a `PR`?

amaitland:
(Or even just apply the changes directly to `master`)

exinferis:
That means that you go under the assumption, that all js functions are per se lowercase (so ```javascriptCaseMethodNames``` refers to lowercase) which isn't always the case - this is a widespread behaviour, but not necessary nor wanted in many cases. I find ```javascriptCaseMethodNames``` misleading - it is about lowercasing, since "javascript-casing" does not really exist as such.

amaitland:
Technically it's probably `camel-case` that we're referring to, to me that's a fairly universal term. So something like `camelCaseJavascriptNames` is probably quite universal. Thoughts?

exinferis:
Sounds good.

PR #932, opened 2015-04-08
Add CefVisitWebPluginInfo Wrapper

amaitland:
Implement `CefVisitWebPluginInfo` using a simple `TaskCompletionSource` to notify when all plugins have been visited.

Also added `custom://cefsharp/plugins` that outputs a basic html table showing plugin details.

### DIFF
--- a/CefSharp.Core/Cef.h
+++ b/CefSharp.Core/Cef.h
@@ -455,8 +455,12 @@ namespace CefSharp
             return CefClearSchemeHandlerFactories();
         }
 
-        /// <summary>Visits all plugins using the provided Plugin Visitor.</summary>
-        /// <param name="visitor">A user-provided Plugin Visitor implementation.</param>
+        /// <summary>
+        /// Async returns a list containing Plugin Information
+        /// (Wrapper around CefVisitWebPluginInfo)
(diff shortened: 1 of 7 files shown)


CefSharp.Core/Cef.h
```diff
@@ -454,6 +455,17 @@ namespace CefSharp
             return CefClearSchemeHandlerFactories();
         }
 
+        /// <summary>Visits all plugins using the provided Plugin Visitor.</summary>
+        /// <param name="visitor">A user-provided Plugin Visitor implementation.</param>
+        static Task<List<Plugin>^>^ GetPlugins()
```
jankurianski: Visitor parameter comment should be removed.

amaitland: Thanks :+1:

amaitland:
@jankurianski I've updated the xml doc, any suggestions on the wording are most welcome :smile:

jankurianski:
I think that's good.  I also checked out the PR to give it a spin and it works for me. :+1:

amaitland:
> I also checked out the PR to give it a spin and it works for me.

Thanks :+1: Much appreciated :smile: Figure this might make debugging plugin problems a little easier, at least we can now see what's being detected.

PR #933, opened 2015-04-09
Add Experiment TaskScheduler

amaitland:
Rebased against `2272`


rassilon:
This is cool, I haven't looked at the code, but being able to use a .Net Task TaskScheduler for CEF thread issues would be awesome. Do we have wrappers for checking if we're on a specific CEF thread?

amaitland:
The functionality has existed in the `BrowserSubProcess` for some time, thought it would be useful to make it a little more generic. I can't claim credit for the idea, I just ported it from a very old PR from @JanEggers 

> Do we have wrappers for checking if we're on a specific CEF thread?

Not as yet, that's an excellent idea though.

If you've got some time and want to take the idea and run with it, be my guest :smile:

amaitland:
I was thinking it's probably worth creating a `ThreadManager` class or something similar, to house all the functions and task wrappers.

For now I'm reverting the `ZoomLevel` code that was using the `TaskFactory`. Merged into `2272`, so at some point in the future work can continue on this.

rassilon:
I looked some more into .Net Task stuff today, I'm thinking we probably need the following:

* A separate `SyncronizationContext` for each CEF thread type
* A `TaskScheduler` for each CEF thread type (unsurprisingly from a common base class)
* A `TaskFactory` instance per CEF thread type to that helps creates tasks against our `TaskScheduler` instance.
* Setting `TaskScheduler.Current` during callbacks from CEF to the correct scheduler.
* ditto for `SyncronizationContext.Current`
* The `TaskScheduler`s would only post one task at a time to the thread so that the `QueryPending` (or whatever its exact name is) method can return data on what isn't being processed yet. The idea would be to Post the next job after the current one finishes.
* We need to remember to use the new `base::Bind` stuff instead of the runnable stuff in the CEF API due to depreciation of the latter.
* We might not need a custom `TaskScheduler` if the default implementation for `ThreadScheduler.FromSyncronizationContext` static method meets our needs combined with our `SyncronizationContext` code.

Those are the broad strokes just thinking about it briefly after looking at the .Net Task docs today.
Some of this goo could be wrapped up in a ThreadManager component, having the `.Current` goo in the same method is a good idea. :smile:

Bill

PR #936, opened 2015-04-10
Cleanup WPF TooltipTimer

amaitland:
- Extract Tooltip timer functionality into an extension method
- Extract `UiThreadRunAsync` into an extension method


amaitland:
There's a small problem with this still, needs to be revisited, when moving the mouse over multiple items, the previous value can be seen for a brief second.

PR #940, opened 2015-04-10
Create JS objects from dictionaries

darvell:
This allows any JS binding to return a dictionary of Dictionary< String, Object > and converts that into a standard javascript object.

Not sure if this is the cleanest way to do it, this is my first foray in to .NET C++!

### DIFF
--- a/CefSharp.BrowserSubprocess.Core/TypeUtils.cpp
+++ b/CefSharp.BrowserSubprocess.Core/TypeUtils.cpp
@@ -146,6 +146,19 @@ namespace CefSharp
 
             return cefArray;
         }
+        
+        if (type == Dictionary<String^, Object^>::typeid)
+        {
+            CefRefPtr<CefV8Value> cefObject = CefV8Value::CreateObject(NULL);
+            for each(KeyValuePair<String^, Object^>^ kvp in dynamic_cast<Dictionary<String^, Object^>^>(obj))
+            {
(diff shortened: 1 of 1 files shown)


amaitland:
@darvell Thanks for the contribution! This is somewhat an interesting one, there is already PR #893 which adds `Dictionary` support in a very different fashion.

I'm unsure which approach is better long term, so probably needs to be discussed.

What was your usage scenario? In general it would be handy to include an example to demo your project. e.g. add to `CefSharp.Example.BoundObject`

darvell:
My usage case would be passing a Javascript object to C#, modifying it then sending it back. 

It seems most usages would suggest to just send the object to C#, modify it, then send it back as a serialized string and run JSON.parse() on the result but that seems awfully unnecessary to do deserialize-modify-serialize-deserialize.

I'll see about creating an example in CefSharp.Example.BoundObject in the next day since I don't think it's very clear without it!

amaitland:
There are quite a few limitations with this though, like only supporting primitive types for the value in the dictionary. Complex objects won't pass through the `WCF` proxy, hence the need to `JSON` serialize yourself in some scenarios.

If #893 were to be merged then this code would most likely be redundant. Hence the need for discussion.

rassilon:
Hrm. I think I agree with @amaitland  that #893 is more promising. Not to mention serialization issues are detected much more quickly for #893 (in the browser process instead of in the render sub process)

amaitland:
@darvell Did you have a chance to checkout #893? Any comments?

darvell:
Looks like a better approach in the end. While I still think that being able to properly send JS objects around would be really beneficial, my approach is further looking like the wrong way to handle that use-case.

PR #949, opened 2015-04-15
Refactored dependency checking functions to use CefSettings and check...

pushplay:
... the BrowserSubprocessPath setting.

### DIFF
--- a/CefSharp/CefSharp.csproj
+++ b/CefSharp/CefSharp.csproj
@@ -76,6 +76,9 @@
     <DocumentationFile>bin\x64\Release\CefSharp.XML</DocumentationFile>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="CefSharp.Core">
+      <HintPath>..\CefSharp.Core\bin\Debug\CefSharp.Core.dll</HintPath>
+    </Reference>
     <Reference Include="System" />
     <Reference Include="System.ComponentModel.Composition" />
     <Reference Include="System.Core" />
--- a/CefSharp/DependencyChecker.cs
+++ b/CefSharp/DependencyChecker.cs
@@ -69,22 +69,36 @@ namespace CefSharp
         {
             "CefSharp.Core.dll",
             "CefSharp.dll",
-            "CefSharp.BrowserSubprocess.Core.dll",
-            "CefSharp.BrowserSubprocess.exe"
+            "CefSharp.BrowserSubprocess.Core.dll"
         };
 
         /// <summary>
         /// CheckDependencies iterates through the list of Cef and CefSharp dependencines
         /// relative to the path provided and returns a list of missing ones
         /// </summary>
+        /// <remarks>
+        /// The following <see cref="CefSettings"/> parameters will take on default values if not specified:
+        /// <list>
+        /// <item><description>Locale - if empty then en-US will be used.</description></item>
+        /// <item><description>LocalesDirPath - if empty locales\ will be used.</description></item>
+        /// <item><description>ResourcesDirPath - if empty the Executing Assembly path is used.</description></item>
+        /// </list>
+        /// </remarks>
         /// <param name="checkOptional">check to see if optional dependencies are present</param>
-        /// <param name="packLoadingDisabled">Is loading of pack files disabled?</param>
         /// <param name="path">path to check for dependencies</param>
-        /// <param name="resourcesDirPath"></param>
-        /// <param name="localePackFile">The locale pack file e.g. <see cref="LocalesPackFile"/> </param>
+        /// <param name="cefSettings">The settings object specifying dependencies</param>
         /// <returns>List of missing dependencies, if all present an empty List will be returned</returns>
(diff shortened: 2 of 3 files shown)


pushplay:
I made the two dependency checking functions more consistent by using ```CefSettings``` in both and putting the defaulting behaviour in ```CheckDependencies```.

CefSharp/CefSharp.csproj
```diff
@@ -76,6 +76,9 @@
     <DocumentationFile>bin\x64\Release\CefSharp.XML</DocumentationFile>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="CefSharp.Core">
```
amaitland: This will likely cause a problem, as `CefSharp.Core` has a dependency on `CefSharp`, so we're creating a circular reference here.

So we'll have to revert to using the old method of passing in the params.

CefSharp/DependencyChecker.cs
```diff
@@ -69,22 +69,36 @@ public static class DependencyChecker
         {
             "CefSharp.Core.dll",
             "CefSharp.dll",
-            "CefSharp.BrowserSubprocess.Core.dll",
```
amaitland: Need to also check that `CefSharp.BrowserSubprocess.Core.dll` is present for completeness.

CefSharp/DependencyChecker.cs
```diff
@@ -137,8 +153,19 @@ public static List<string> CheckDependencies(bool checkOptional, bool packLoadin
                 }
             }
 
+            var browserSubprocessPath = string.IsNullOrEmpty(cefSettings.BrowserSubprocessPath)
```
amaitland: By default `BrowserSubprocessPath` defaults to `CefSharp.BrowserSubprocess.exe` so I think it's probably better just to check the property, if empty throw a unique exception, possibly in `Cef.Initialize()` it's self?

amaitland:
@pushplay As you can see the build fails. From a quick look I believe this is due to the circular reference you've added. I commented inline. So it looks like we'll have to use the original approach and just pass in individual params.

pushplay:
@amaitland Yeah I just figured that out.  The real question here is why Visual Studio didn't point that out.

CefSharp/DependencyChecker.cs
pushplay: Ok I misunderstood.  I thought that was a DLL for CefSharp, not the BrowserSubprocess.

amaitland:
> The real question here is why Visual Studio didn't point that out.

It doesn't explicitly checks for circular references, just relies on builds failing to handle the scenario.

Anyways, few things to come out of this

- Check `BrowserSubProcessPath` in `Cef.Initialize()` and throw exception if path is empty
- Add `BrowserSubProcessPath` param to `DependencyChecker`

CefSharp/DependencyChecker.cs
amaitland: It's the `C++` wrapper around the `CEF` functionality. `BrowserSubProcess` basically exposes the functionality (using `c#` where possible, as it's easier for the most part)

amaitland:
For completeness I think it's still worth adding some checking for `BrowserSubProcessPath `, though it's somewhat unlikely for it to be located outside of the bin directory for your average scenario. Fact is that it requires `libcef.dll` and so does `CefSharp.Core`.

I've seen people try to move the dependencies into a sub folder to group them together, which of course makes sense. Nobody has reported that they were successful, so some of this maybe moot.

pushplay:
Aha!  If I dump a bunch of CEF files into ```CefSharp.BrowserSubprocess\bin\x86\Debug``` then BrowserSubprocess launches from VisualStudio.  ```CefSharp.BrowserSubprocess.exe``` has more dependencies than ```libcef.dll``` but I'm out of time to narrow it down today.  I'll do it tomorrow.

What ever the rules are for DLL loading and child processes it's different when run out of ```vhost```.

I think ```DependencyChecker``` should check for all dependencies of ```CefSharp.BrowserSubprocess.exe``` being in the same directory and fail if they're not there.  That means ```CefExample``` as it currently is would fail ```DependencyChecker```, which matches my experience.  Thoughts?

amaitland:
> That means CefExample as it currently is would fail DependencyChecker, which matches my experience. Thoughts?

Were you running the build in `Release` mode by any chance? If we can update the example to differentiate between `Release` and `Debug` then the problem is likely to go away.

pushplay:
> Were you running the build in Release mode by any chance?

Nope.  ```settings.BrowserSubprocessPath = "..\\..\\..\\..\\CefSharp.BrowserSubprocess\\bin\\" + architecture + "\\Debug\\CefSharp.BrowserSubprocess.exe"``` only happens when the debugger is attached anyways.

Maybe the Visual Studio version affects the vhost's DLL loading rules.  I have 2013 here.

pushplay:
I narrowed down the files ```CefSharp.BrowserSubprocess.exe``` needs to be happy.  It's ```icudtl.dat``` and ```libcef.dll``` which makes sense.

PR #950, opened 2015-04-16
Rewrite ResourceHandler/SchemeHandler

amaitland:
This is the start of rewriting `ResourceHandler/SchemeHandler`

Have gotten to a point where a clearer direction is required. Perhaps it is better to completely overhaul `ResourceHandler`? Maybe even remove the `LoadHtml` component to simplify?

Discussion required, so pausing development for now. Feel free to comment inline or below.


amaitland:
@rassilon The further I got into this the more I'm inclined to make sweeping changes, code is a bit of a mess overall.

If you have any inspiration or ideas, you should be able to commit to my personal `CefSharp` repository directly.

rassilon:
This is some pretty good cleanup! Awesome work!

I think the only additional work is to figure out how badly we want to make a breaking change with the `GetResouceHandler` return type from the `ResourceHandlerFactory`. (in addition to the other comments I made)

i.e. should it return an `IResourceHandler` unconditionally? or should the caller be able to return an `AsyncRequestHandler` like we talked about last we discussed?

Bill

amaitland:
> I think the only additional work is to figure out how badly we want to make a breaking change with the GetResouceHandler return type from the ResourceHandlerFactory. (in addition to the other comments I made)

The continuation delegate seems like it's going to be the hardest part to integrate overall.

> i.e. should it return an IResourceHandler unconditionally? or should the caller be able to return an AsyncRequestHandler like we talked about last we discussed?

I think if we use `Async` as the basis that will make the code cleaner. We could provide some sort of wrapper that mimics the current behavior by calling continue immediately, keeping the `API` mostly the same.

If it becomes to hard then we go with the total rewrite. I get the impression that people are actually trying to perform work in the resource handler context currently not knowing that it's `sync` and slowing down their app. So they'll be slightly annoyed to start with, their app overall will benefit from a rewrite.

rassilon:
> The continuation delegate seems like it's going to be the hardest part to integrate overall.

I think given the abstract position we're in, we can only call the continuation when the stream is done reading. (Since the `CEF` docs say to only call it when you have available headers, not when the response has finished reading.) Except for possibly the current `sync` ResponseHeaders class since as soon as we get it we have the stream (in an unknown state) and the header data.

(i.e. its an abstract stream and not a streaming HTTP response reader that can let us know asap when the headers are finished being received)

Bill

rassilon:
> I think if we use Async as the basis that will make the code cleaner. We could provide some sort of wrapper that mimics the current behavior by calling continue immediately, keeping the API mostly the same.

For the Asnc, i'd make it just the ResourceHandler goo, so it'd up to the user to call the ProcessRequest(...callback param) ->Continue method when they have or (or don't) the header data.

Some of these details might be less interesting if we make it easy to utilize `CefURLRequest/CefURLRequestClient` in some a sane async fashion. (i.e. this stuff might be the 20% or 5%, instead of maybe the 40-50%)

Bill

Bill

amaitland:
> Some of these details might be less interesting if we make it easy to utilize CefURLRequest/CefURLRequestClient in some sane async fashion. (i.e. this stuff might be the 20% or 5%, instead of maybe the 40-50%)

Maybe...`CefURLRequest` still doesn't provide the ability to modify the response body, so it's not quite as useful as it could be.

rassilon:
> Maybe...CefURLRequest still doesn't provide the ability to modify the response body, so it's not quite as useful as it could be.

True, but you can't do all of the work for our users. :smile:  They have to put in SOME work...

amaitland:
> They have to put in SOME work...

I've spoken to them and they don't agree! Apparently we're supposed to drop everything and help them with every little problem, even those totally unrelated to `CefSharp`! :frowning:

amaitland:
Need to re target this on `2272`

amaitland:
Have you dug into `CefStreamReader::CreateForFile()` at all? 

Short term improvement to `39.0.1` might be to skip reading the stream into memory for files and use the file path? Potentially a quick performance improvement?

rassilon:
It's pretty boring and doesn't contain itself to just the FILE thread. `CefStreamResourceHandler` is more interesting it buffers reads through  `CefStreamReader` and ensures actual reads from the stream occur on the FILE thread.

amaitland:
Hmm, yeah `CefStreamResourceHandler` is quite interesting. Anything stopping us from passing in a `StreamAdapter` that's late bound to the stream, override `ProcessRequest` to capture the callback (not calling it straight away as currently occurs). Set the headers ourselves in `GetResponseHeaders` and then let `CefStreamResourceHandler` do the rest?

rassilon:
Ideally we would also do a bit more fiddling to prevent double buffering of file streams. 

Otherwise, yeah that's the idea.

jornh:
> I've spoken to them and they don't agree!

lol, sounds totally in line with what you would normally expect :wink:

PR #951, opened 2015-04-16
Rename lowerCaseJavascriptNames to camelCaseJavascriptNames

amaitland:
As per discussion starting https://github.com/cefsharp/CefSharp/pull/923#issuecomment-92310713

### DIFF
--- a/CefSharp.OffScreen/ChromiumWebBrowser.cs
+++ b/CefSharp.OffScreen/ChromiumWebBrowser.cs
@@ -305,9 +305,9 @@ namespace CefSharp.OffScreen
             Load(url);
         }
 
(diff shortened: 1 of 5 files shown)


amaitland:
@perlun Happy with `camelCaseJavascriptNames`?

perlun:
Yep, that's much better.

There are probably people that would argue that this should be `false` by default. I don't know; I like the current behavior (after all, I was the one that implemented it in the first place. :sunglasses:) but I understand it's a but unintuitive. So you guys could consider revisiting it for v41.0 or so.

amaitland:
> So you guys could consider revisiting it for v41.0 or so.

That would be a breaking change, so I'm happy just to leave it. Fact that you can now choose is enough I think :+1:

PR #954, opened 2015-04-16
Call GetZoomLevel On CEF UI Thread

rassilon:
GetZoomLevel is required to be called from the UI thread, but it has a failure mode of just returning 0 when its not on the UI thread. Here's a cheesy temporary solution to avoid adding a full on GetZoomLevelAsync method that returns a .Net task.

Bill

### DIFF
--- a/CefSharp.Core/CefSharp.Core.vcxproj
+++ b/CefSharp.Core/CefSharp.Core.vcxproj
@@ -110,7 +110,7 @@
     <ClCompile>
       <Optimization>Disabled</Optimization>
       <AdditionalIncludeDirectories>$(SolutionDir)packages\$(CefSdkVer)\CEF;%(AdditionalIncludeDirectories);$(ProjectDir)</AdditionalIncludeDirectories>
-      <PreprocessorDefinitions>_DEBUG;_WIN32_WINNT=0x0600;WINVER=0x0600;WIN32;_WINDOWS;WIN32_LEAN_AND_MEAN;USING_CEF_SHARED;_CRT_SECURE_NO_WARNINGS;EXPORT;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <PreprocessorDefinitions>_DEBUG;_WIN32_WINNT=0x0600;WINVER=0x0600;WIN32;_WINDOWS;WIN32_LEAN_AND_MEAN;USING_CEF_SHARED;_CRT_SECURE_NO_WARNINGS;EXPORT;OS_WIN;%(PreprocessorDefinitions)</PreprocessorDefinitions>
       <RuntimeLibrary>MultiThreadedDebugDLL</RuntimeLibrary>
       <PrecompiledHeader>Use</PrecompiledHeader>
       <GenerateXMLDocumentationFiles>true</GenerateXMLDocumentationFiles>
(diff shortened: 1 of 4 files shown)


CefSharp.Core/CefSharp.Core.vcxproj
```diff
@@ -110,7 +110,7 @@
     <ClCompile>
       <Optimization>Disabled</Optimization>
       <AdditionalIncludeDirectories>$(SolutionDir)packages\$(CefSdkVer)\CEF;%(AdditionalIncludeDirectories);$(ProjectDir)</AdditionalIncludeDirectories>
-      <PreprocessorDefinitions>_DEBUG;_WIN32_WINNT=0x0600;WINVER=0x0600;WIN32;_WINDOWS;WIN32_LEAN_AND_MEAN;USING_CEF_SHARED;_CRT_SECURE_NO_WARNINGS;EXPORT;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <PreprocessorDefinitions>_DEBUG;_WIN32_WINNT=0x0600;WINVER=0x0600;WIN32;_WINDOWS;WIN32_LEAN_AND_MEAN;USING_CEF_SHARED;_CRT_SECURE_NO_WARNINGS;EXPORT;OS_WIN;%(PreprocessorDefinitions)</PreprocessorDefinitions>
```
amaitland: Does `CEF` require `OS_WIN` to be set?

rassilon: It does if you want the base::Bind(...) overloads for __stdcall. (Sadly, `Marshal.GetNativeFunctionPointerForDelegate` doesn't like generic delegates or generic parameters. :frowning: So that atttempt didn't work either. Ugh.

I figured it would probably be good to leave it in case CEF uses the define for other stuff.

Bill

amaitland: > It does if you want the base::Bind(...) overloads for __stdcall.

I couldn't figure out what changes in this `PR` would require a new pre-processor directive, makes sense now. Since it has a potential purpose then happy to leave in :+1:

CefSharp.WinForms.Example/BrowserForm.cs
```diff
@@ -12,6 +12,9 @@ public partial class BrowserForm : Form
     {
         private const string DefaultUrlForAddedTabs = "https://www.google.com";
 
+        // Default to WPF default increment:
+        private double zoomIncrement = 0.10;
```
amaitland: Make this a `const` and get rid of the `WPF` reference?

CefSharp.WinForms.Example/BrowserForm.cs
```diff
@@ -230,5 +233,36 @@ private void CloseDevToolsMenuItemClick(object sender, EventArgs e)
...
+                control.Browser.ZoomLevel -= zoomIncrement;
+            }
+        }
+
+        private void currentZoomLevelToolStripMenuItem_Click(object sender, EventArgs e)
+        {
+            var control = GetCurrentTabControl();
+            if (control != null)
+            {
+                MessageBox.Show("Current ZoomLevel: " + control.Browser.ZoomLevel);
```
amaitland: Showing the zoom level is a nice touch :+1:

CefSharp.WinForms.Example/BrowserForm.cs
```diff
@@ -230,5 +233,36 @@ private void CloseDevToolsMenuItemClick(object sender, EventArgs e)
                 control.Browser.CloseDevTools();
             }
         }
+
+        private void zoomInToolStripMenuItem_Click(object sender, EventArgs e)
```
amaitland: Rename to `ZoomInToolStripMenuItemClick` for consistency, other two click handlers also.

CefSharp.WinForms.Example/BrowserForm.cs
rassilon: Done

CefSharp.WinForms.Example/BrowserForm.cs
rassilon: Done.

amaitland:
Works as advertised :+1: Merge away.

PR #955, opened 2015-04-17
Rewrite ResourceHandler/SchemeHandler

amaitland:
Rebased version of #950 on `2272`


amaitland:
@rassilon Thanks for the merge :+1:

amaitland:
Make sure to test that problems discovered with https://github.com/cefsharp/CefSharp/commit/a0460e83d81bb7964d9027af2f69f277507443d1 have been resolved.

rassilon:
The changes in a0460e8 seem odd. What was the underlying cause of the problem?

rassilon:
Oh, could it be due to `CefExmaple.RegisterTestResources` being an async method but it doesn't:

* Return a Task type like it should.
* Utilize `.ConfigureAwait(continueOnCapturedContext: false)`

Thus, the `SyncronizationContext` is the UI thread for `BrowserTabUserControl` (in the WinForms example).

I'm not sure how this relates exactly to a0460e8 though. I would have thought `ProcessRequestAsync` would run on a non-UI thread so that `Task.Run` would run against the correct `SyncronizationContext/TaskScheduler`.

Bill

amaitland:
> The changes in a0460e8 seem odd. What was the underlying cause of the problem?

Unsure, just know it was throwing a `ContextSwitchDeadlock`.

Seems to be sorted now, so I've switched to using a task again, aka https://github.com/amaitland/CefSharp/commit/48268721c054f84c4da59b2f090c1030372a58f9

rassilon:
Problems that disappear are always nice. :smile:

amaitland:
> Problems that disappear are always nice. :smile:

Indeed :smile:

rassilon:
Merged up from 2272 head again.

amaitland:
Merging this now, will look at finished the `async` implementation of resource handler in the not too distant future.

PR #959, opened 2015-04-21
Add BrowserSubProcess dependency checking.

pushplay:


CefSharp/DependencyChecker.cs
```diff
@@ -81,64 +91,44 @@ public static class DependencyChecker
...
+            }
 
-                if (!File.Exists(dependencyPath))
-                {
-                    missingDependencies.Add(cefSharpDependency);
-                }
+            var browserSubprocessDir = Path.GetDirectoryName(browserSubProcessPath);
+            if (browserSubprocessDir == null)
+            {
+                missingDependencies.AddRange(BrowserSubprocessDependencies);
```
amaitland: Should there be a `CheckDependencyList` call here?

pushplay: No at this point browserSubProcessPath is something that doesn't exist, so everything that should be in there is by default missing.  I considered adding browserSubProcessPath itself to the collection instead, but that didn't seem consistent with the rest of the code.  Also browserSubProcessPath could be null and then there's not enough explanation of what's wrong.

amaitland:
I made one small tweak, more from a style/readability point of view
https://github.com/cefsharp/CefSharp/commit/969b6944fe438ea01d23feea3b2ca120920035a8

Other than that, looks solid :+1: Thanks!

amaitland:
When I tried in `Debug` mode the extra subprocess checking throws an exception.

So temporarily I've disabled if the debugger is attached see https://github.com/cefsharp/CefSharp/commit/3e3ea4f30561f5a365aa7c3f47af475ab3aca9f9

Previously the `BrowserSubProcess` worked fine being isolated on it's own without `libcef.dll` etc, so I'm not sure we actually need those checks or not?

pushplay:
In my dev environment ```BrowserSubProcess``` would not launch unless ```libcef.dll``` is there.  I think the difference might come down to ```vhost.exe``` versions.  Which version of VS are you running?

amaitland:
I use `VS2012` at work and `VS2013` at home. I've tested in `VS2013`, can try `VS2012` later.

I've never had a problem with either previously.

pushplay:
Yeah I have VS 2013 at work.  So much for that theory.

amaitland:
`VS 2013 Update 4`?

pushplay:
Yup.

amaitland:
It's easy enough to add a `.props` file to copy the dependencies. I'm just curious as to what's different between environments?

I'll try on my work machine later, it has both `VS2012` and `VS2013` (I use `VS2012` for the most part as the Resharper license I have is old and doesn't support `VS2013`). It's `Windows 7` where I'm running `Windows 8.1` at home

pushplay:
I'm also curious but I'm out of ideas.  I don't know too much about the rules for loading and sharing libraries in Windows.  OS is Windows 8.1 Enterprise here.  Maybe there's some obscure security policy that's relevant?

Programming in VS without ReSharper is sadness.

rassilon:
@amaitland , maybe we could ask for ReSharper/NDepend OpenSource licenses for CefSharp?

That'd be very cool.

amaitland:
I believe at one point in time @perlun had organized `Resharper` licenses for `CefSharp`. I think it's worth following up on.

I'd like to try out `Resharper for C++`, that could be very handy :+1:

perlun:
I have that, yes. Give me your email (perlun in Google Mail) and I'll see what I can do. Didn't even know about ReSharper for C++, that sounds incredibly nice, since C++ is so horrible to work with otherwise. :stuck_out_tongue:

amaitland:
> Didn't even know about ReSharper for C++

It was only just released a couple of weeks ago. See release announcement for some pretty detailed info http://blog.jetbrains.com/dotnet/2015/04/10/introducing-resharper-cpp/

amaitland:
> Give me your email (perlun in Google Mail)

Email sent, thanks :+1:

rassilon:
Ditto, thanks :+1:

rassilon:
Hey @perlun, @amaitland , the NDepend folks got back to me, they are still willing to give out licenses for OSS projects, however they require additional details in order to issue the licenses.

I emailed @perlun about it in hopes he'd reply-all to all of us via email about it, but I haven't heard back yet. I'm <firstname>.<lastname> at Google Mail. You can find that info in my github profile now.

Fyi,
Bill

amaitland:
> the NDepend folks got back to me, they are still willing to give out licenses for OSS projects

Nice one :+1: 

> however they require additional details in order to issue the licenses.

Out of curiosity? What sort of information are they asking for?

rassilon:
Postal address & phone number.

rassilon:
@amaitland , would you like to be included on the NDepend thing?

amaitland:
Sure. Best to email?

rassilon:
Yeah. See previous comment about my email address. I've edited the Markdown so it renders properly now.

amaitland:
Email sent.

PR #961, opened 2015-04-22
Add TaskFactories for UI/File/IO and use for GetZoomLevel

rassilon:
As an added bonus, add GetZoomLevelAsync.

The bummer is that IOT_SetCookie/IOT_DeleteCookie wasn't worth converting over, even though I really would love doing that. CefCookie isn't allowed to be an `Action/Func` generic parameter due to it being a C++ template class.

### DIFF
--- a/CefSharp.Core/Internals/CefTaskScheduler.h
+++ b/CefSharp.Core/Internals/CefTaskScheduler.h
@@ -57,9 +57,9 @@ namespace CefSharp
         public ref class CefTaskScheduler : TaskScheduler, ITaskScheduler
         {
         public:
-            cef_thread_id_t _thread;
+            CefThreadId _thread;
 
-            CefTaskScheduler(cef_thread_id_t thread) :
+            CefTaskScheduler(CefThreadId thread) :
                 _thread(thread)
             {
             };
@@ -75,6 +75,24 @@ namespace CefSharp
             {
                 TryExecuteTask(task);
             };
+
+            static bool CurrentlyOnThread(CefThreadId threadId)
+            {
+                return CefCurrentlyOn(threadId);
+            }
+
+            static void EnsureOn(CefThreadId threadId, String^ context)
+            {
+                if (!CefCurrentlyOn(threadId))
+                {
+                    throw gcnew InvalidOperationException(String::Format("Executed '{0}' on incorrect thread. This method expects to run on the CEF {1} thread!", context, ((CefThreadIds)threadId).ToString()));
+                }
+            }
+
+            static void EnsureOn(CefThreadIds threadId, String^ context)
+            {
+                EnsureOn((CefThreadId)threadId, context);
+            }
 
         protected:
             virtual bool TryExecuteTaskInline(Task^ task, bool taskWasPreviouslyQueued) override
--- a/CefSharp.Core/Internals/CefTaskScheduler.h
+++ b/CefSharp.Core/Internals/CefTaskScheduler.h
@@ -6,6 +6,7 @@
 
 #include "Stdafx.h"
 #include <include/cef_runnable.h>
+#include <include/cef_task.h>
 
 using namespace System;
 using namespace System::Threading::Tasks;
@@ -42,7 +43,11 @@ namespace CefSharp
                 }
                 catch (Exception^)
                 {
-
+                    // This should never ever happen.
+                    // If this occurs then someone has broken a .Net ThreadScheduler/Task invariant
+                    // i.e. trying to run a task on the wrong scheduler, or some
+                    // weird exception during task completion. 
+                    // TODO: We need to forward this exception somewhere...
                 }
             };
 
@@ -74,6 +79,15 @@ namespace CefSharp
         protected:
             virtual bool TryExecuteTaskInline(Task^ task, bool taskWasPreviouslyQueued) override
             {
+                // You might think this method might not get called,
+                // but a .ContinueWith(..., TaskContinuationOpation.ExecuteSyncronously)
+                // will probably end up calling this method,
+                // so assume the callers know what they're doing and execute the task
+                // inline on this thread (if the current thread is correct for this scheduler)
+                if (CefCurrentlyOn(_thread))
+                {
+                    return TryExecuteTask(task);
+                }
                 return false;
             };
 
--- a/CefSharp.Core/ManagedCefBrowserAdapter.cpp
+++ b/CefSharp.Core/ManagedCefBrowserAdapter.cpp
@@ -1,0 +1,32 @@
+#include "Stdafx.h"
+#include "Cef.h"
+
+double ManagedCefBrowserAdapter::GetZoomLevel()
+{
+    auto browser = _clientAdapter->GetCefBrowser();
+
+    if (browser != nullptr)
+    {
+        auto host = browser->GetHost();
+        if (CefCurrentlyOn(TID_UI))
+        {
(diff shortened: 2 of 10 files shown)


CefSharp.Core/ManagedCefBrowserAdapter.cpp
```diff
@@ -0,0 +1,32 @@
...
+        {
+            return host->GetZoomLevel();
+        }
+        else
+        {
+            // TODO: Add an async version of GetZoomLevel at some point.
+            // NOTE: Use of ManualResetEvent is required here in order
+            // for simple marshaling of some kind of synchronization primitive
+            // to the callback.
+            Task<double>^ task = Cef::UIThreadTaskFactory->StartNew(gcnew Func<double>(this, &ManagedCefBrowserAdapter::GetZoomLevel));
```
amaitland: As this is hidden away, I'd probably move this into the `C#` world. Thoughts?  The `ChromiumWebBrowser` implementations can deal with as required. I think we'd be better to migrate the users over to using the `async` version. This would also coincide with removing the `ZoomLevel` from `IWebBrowser` and replacing with `GetZoomLevel`. 

Thoughts?

CefSharp.Core/ManagedCefBrowserAdapter.cpp
```diff
@@ -0,0 +1,32 @@
+#include "Stdafx.h"
```
amaitland: Add license disclaimer.

CefSharp.Core/Internals/CefTaskScheduler.h
```diff
@@ -42,7 +43,11 @@ namespace CefSharp
                 }
                 catch (Exception^)
                 {
-
+                    // This should never ever happen.
```
amaitland: Perhaps just remove the `try/catch` and let the exception bubble?

CefSharp.Core/ManagedCefBrowserAdapter.cpp
amaitland: Also if we're going to use a `.cpp` file should probably migrate everything over? Little inconsistent otherwise.

rassilon: Added license. I only moved the bare minimum to get it to compile since the .h files include each otherwise. I can move the rest though.

CefSharp.Core/Internals/CefTaskScheduler.h
rassilon: But it would bubble into the CEF logic and not into .Net. (Since it's on the CEF thread) Otherwise I would have just removed it as you suggest.

rassilon: Manually call AppDomain.UnhandledException handler if we can? I'm not sure we can do that. I'll take a look.

amaitland: Ahh, true. There really is not nice place to raise a `.Net` error from within `CEF`. Have to think about that.

amaitland: Maybe we create something similar, like `Cef.UnhandledException` and raise errors (at least in the browser process, not much we can do in the sub process).

CefSharp.Core/ManagedCefBrowserAdapter.cpp
rassilon: We can totally remove the sync version as a breaking change. We'd need to either export `CefCurrentlyOn` or implement `SyncronizationContexts` in order to have the non-async version in C#. I'd rather not do `SyncronizationContexts` since I think ensuring the current context got updated correctly wouldn't be maintainable long term. (since you'd have to call the context installation method on every callback from CEF to .Net)

No argument on the change you suggest to `IWebBrowser`.

amaitland: I haven't had time to read over the `SyncronizationContext` stuff in any details, so I'll be guided by you on this one.

I think we take the pragmatic approach, something that's simple and workable. The `ZoomLevel` property may stay for the `WPF` implementation, just moved to the `IWpfWebBeowser` interface.

rassilon: Ok, I've yanked sync version from here, and call the async version from WinForms/OSR, but do so syncronously. WPF doesn't actually call `GetZoomLevel` atm. Seems a bit odd.

CefSharp.Core/Internals/CefTaskScheduler.h
rassilon: Probably something like that, but on both `Cef` and `CefAppWrapper` or something so we handle browser and render process cases. Probably via an interface/abstract C++ class.

CefSharp.Core/ManagedCefBrowserAdapter.cpp
amaitland: > WPF doesn't actually call GetZoomLevel atm. Seems a bit odd.

I didn't work until your previous PR, so there wasn't much point calling it. It just kept an internal reference to the amount that you incremented it by. It could now be updated.

rassilon:
Remaining tasks:
- [ ] Unhandled exception pattern
- [ ] Move more of implementation to `ManagedBrowserAdapter.cpp`
- [ ] Breaking change on `IWebBrowser` for `ZoomLevel` -> `GetZoomLevelAsync`

Anything else you can think of?

amaitland:
Leave `ZoomLevel` on the `IWpfWebBrowser` interface for now, have to think about the best way to deal with that one.

So for clarity, we keep the Task's that are linked to the `CefTaskScheduler` hidden behind the `API`? That way we don't need to worry about `SyncContexts`?

rassilon:
> So for clarity, we keep the Task's that are linked to the `CefTaskScheduler` hidden behind the API? That way we don't need to worry about `SyncContexts`?

I'm not sure what you're asking. Are you asking about `GetZoomLevel` now being private or something else?

I'd rather not worry about `SyncContexts` because you'd have to remember to call a `CefSyncContext.InstallIfNeeded` method on EVERY callback into CefSharp. (i.e. we could do it, but it would be hard to maintain over the long haul, and the benefit would be pretty small) 

WinForms gets off easy since there are ~3 places it calls its equivalent method.

rassilon:
The current `SyncContext` can affect how the C# `await` keyword behaves. 
i.e. If I'm remembering correctly if you call your method that uses await from say a WinForms UI thread the task (by default) will get scheduled on the UI thread.

Yeah, see [here](http://social.technet.microsoft.com/wiki/contents/articles/23103.c-avoiding-deadlock-in-asyncawait.aspx) for what I mean about `SyncContext` driving `await` behavior.

Here's a MSDN article on [`Async/Await` best practices](https://msdn.microsoft.com/en-us/magazine/jj991977.aspx) as well.

Other than that it doesn't really matter as long as you ensure you create your tasks with the correct factory.

(As near as I can tell by staring at the Task.cs and TaskScheduler.cs sources on referencesource, that is.)

amaitland:
> Yeah, see here for what I mean about SyncContext driving await behavior.
> 
> Here's a MSDN article on Async/Await best practices as well.

Will have a read when I've got a minute, thanks :+1: 

If you'd prefer to merge this sooner rather than later that's fine with me. Can always address the other items at a later date.

rassilon:
Yeah, merging this in will help make it easier on other chagnes with this batch of code being moved into the .cpp file.

PR #962, opened 2015-04-22
Implement Missing CefRequestHandler Methods

amaitland:
Implemented
- `OnResourceRedirect`
- `OnProtocolExecution`
- `OnQuotaRequest`

TODO: Cleanup method ordering to make alphabetical

### DIFF
--- a/CefSharp.Core/Internals/ClientAdapter.cpp
+++ b/CefSharp.Core/Internals/ClientAdapter.cpp
@@ -317,7 +317,16 @@ namespace CefSharp
 
                 //TODO: Determine if there is a better way to pass the reference in rather than reassigning the string
                 newUrl = StringUtils::ToNative(managedNewUrl);
-            }			
+            }	
+        }
+
+        void ClientAdapter::OnProtocolExecution(CefRefPtr<CefBrowser> browser, const CefString& url, bool& allowOSExecution)
+        {
+            IRequestHandler^ handler = _browserControl->RequestHandler;
+            if (handler != nullptr)
+            {
+                allowOSExecution = handler->OnProtocolExecution(_browserControl, StringUtils::ToClr(url));
+            }
         }
 
         // Called on the IO thread before a resource is loaded. To allow the resource
--- a/CefSharp.Core/Internals/ClientAdapter.cpp
+++ b/CefSharp.Core/Internals/ClientAdapter.cpp
@@ -315,7 +315,6 @@ namespace CefSharp
                 auto managedNewUrl = StringUtils::ToClr(newUrl);
                 handler->OnResourceRedirect(_browserControl, frame->IsMain(), StringUtils::ToClr(oldUrl), managedNewUrl);
 
-                //TODO: Determine if there is a better way to pass the reference in rather than reassigning the string
                 newUrl = StringUtils::ToNative(managedNewUrl);
             }	
         }
--- a/CefSharp.Core/Internals/ClientAdapter.cpp
+++ b/CefSharp.Core/Internals/ClientAdapter.cpp
@@ -307,6 +307,19 @@ namespace CefSharp
             }			
         }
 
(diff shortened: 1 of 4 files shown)


amaitland:
Also need to investigate if there's a better way to pass a `CefString` into managed land by reference
https://github.com/cefsharp/CefSharp/pull/962/files#diff-2731f23ba4e8fcf96f8501a49c6d687eR315

amaitland:
Also need to look into string marshaling, the method we have is a little complex, there are built in methods in `.Net` and `CEF` that could be used instead (reduce the string code down to a few lines, would still keep the helper methods).

http://stackoverflow.com/a/1405251
http://magpcss.org/ceforum/apidocs3/projects/%28default%29/CefStringBase.html#FromString%28constchar_type*,size_t,bool%29

amaitland:
https://github.com/cefsharp/CefSharp/blob/master/CefSharp.Core/Internals/StringUtils.cpp#L49

CefSharp.Core/Internals/ClientAdapter.cpp
```diff
@@ -290,6 +307,28 @@ namespace CefSharp
...
+        void ClientAdapter::OnResourceRedirect(CefRefPtr<CefBrowser> browser, CefRefPtr<CefFrame> frame, const CefString& oldUrl, CefString& newUrl)
+        {
+            IRequestHandler^ handler = _browserControl->RequestHandler;
+            if (handler != nullptr)
+            {
+                auto managedNewUrl = StringUtils::ToClr(newUrl);
+                handler->OnResourceRedirect(_browserControl, frame->IsMain(), StringUtils::ToClr(oldUrl), managedNewUrl);
+
+                //TODO: Determine if there is a better way to pass the reference in rather than reassigning the string
+                newUrl = StringUtils::ToNative(managedNewUrl);
```
amaitland: @rassilon Any ideas on the best way to pass a string by ref into `.Net` land?

rassilon: The same way you'd pass any other non-value type into a .Net method. Isn't the C++/CLI syntax for that the % thingy or something? I'd have to look it up in the C++/CLI syntax.

Other than that, no, `StringBuilder` is as good as it gets. :frowning: 

Bill

amaitland: The `ref` part works without any extra symbols. I was hoping to avoid converting to a `managed string`, letting the user assign the string that's passed by `ref` and then converting back to `CefString`.

rassilon: No such thing, unfortunately. There are tricks you can do if you're going the other way, or if you want to treat a `String` as an `IStream`, but nothing to help out here so far as I know. :frowning:

amaitland: I wasn't overly optimistic, just though I'd ask incase you'd come across something. 

I'd like to avoid doing the `newUrl = StringUtils::ToNative(managedNewUrl);` unless the url actually changed. I was thinking maybe return a bool and only assign if the user returns true.

rassilon: You might be able to make the statement conditional on if the strings are actually different I suppose, but that degrades into a string comparison. Given that the Windows heaps are of the very low fragmenting variety, and any tiny allocations this might add to the pain are probably dwarfed by CEF doing its browser UI thing, I wouldn't worry about it and just leave it like it is.

rassilon: Since, redirects don't happen by the hundreds of thousands (I hope... :smile: )

amaitland: Fair enough, I'll stop over thinking it then :smile:

PR #964, opened 2015-04-23
Async SetCookie

amaitland:
Create wrapper class to call `SetCookie` async using new Task factory.

### DIFF
--- a/CefSharp.Core/Cef.h
+++ b/CefSharp.Core/Cef.h
@@ -19,6 +19,7 @@
 #include "CefSettings.h"
 #include "SchemeHandlerWrapper.h"
 #include "Internals/CefTaskScheduler.h"
+#include "SetCookieInvoker.h"
 
 using namespace System::Collections::Generic; 
 using namespace System::Linq;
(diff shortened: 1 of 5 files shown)


amaitland:
Disclaimer I haven't actually tried to run this yet, it builds. Just toying with an idea.

amaitland:
@rassilon What about something like this to make `SetCookie` use the new `TaskFactory`? It's a little crude, though have to work with what we've got I suppose.

Probably need to come up with a better name than `SetCookieInvoker` and it should be in the `Internals` folder.

rassilon:
Oh, I wouldn't have made it a separate class. If I were to do the work I would have just pushed the argument conversions down into IOT_SetCookie.

And, if we wanted to propagate .Net exceptions inside of IOT_SetCookie we would rethrow them after the `->Wait()` && `->IsFaulted` is `TRUE`.

rassilon:
However, the general idea of creating some kind of .Net class to bridge the gap ought to work and we should strive to do that for complicated cases. 

I think the only reason for IOT_SetCookie to have an Invoker class like you did here is if it exceeds the Func/Action generic parameter count limits.

amaitland:
> Oh, I wouldn't have made it a separate class.

I think it makes the code nicer to read :smile: Overhead of a new class is minimal.

rassilon:
It is minimal, the pattern is just a pain to replicate is all. :smile:

amaitland:
It took like `2mins` :stuck_out_tongue_winking_eye: to create the class!

amaitland:
Renamed class and now wraps both Set and Delete.

I think it's an improvement for now. We can revisit later of a preferable pattern is established. Thoughts?

rassilon:
Looks fine to me.

PR #965, opened 2015-04-23
Add late binding to Javascript object

Erguotou:
Added late binding to JavaScript object, so that objects with cyclic
reference don't cause stack overflow (updated example to prove that),
and also improve performance when binding objects with large or deep
object tree.

For consistency, modified method calls to return a true object, just
like property calls. Also pass down camelCaseJavascriptNames flag to
sub-objects so that naming remain the same.

Added a few KnownTypes to support value type arrays.

### DIFF
--- a/CefSharp.BrowserSubprocess.Core/JavascriptMethodHandler.cpp
+++ b/CefSharp.BrowserSubprocess.Core/JavascriptMethodHandler.cpp
@@ -5,6 +5,7 @@
 #include "Stdafx.h"
 #include "TypeUtils.h"
 #include "JavascriptMethodHandler.h"
+#include "JavascriptObjectWrapper.h"
 
 using namespace CefSharp::Internals;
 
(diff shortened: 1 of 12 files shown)


amaitland:
@Erguotou Cool Thanks! 

Can you start be re-targeting this against https://github.com/cefsharp/CefSharp/tree/cef/2272, you'll likely have to close this PR and create a new one. Only minor bug fixes will be merged into `master` before the `39.0.1` release.

Just as a general comment, `spaces` instead of `tabs`, `braces` on new lines.

If you review the `Files Changed` tab above you'll get a sense of the formatting changes required. Will review properly when this is re-targeted against `2272`.

Welcome to `CefSharp` :smile:

amaitland:
Also in general It's preferably to have a number of smaller commits for review rather than one large one. See https://github.com/cefsharp/CefSharp/blob/master/CONTRIBUTING.md#pull-requestsfeature-branches

amaitland:
Questions for later discussion

- Should late binding be optional?
- What performance impact does this have?

amaitland:
@Erguotou Also please use expressive variable names, e.g. `javascriptObjectWrapper` is preferable to `j`.

amaitland:
Closing this PR as it needs to be re-targeted at `2272`.

PR #967, opened 2015-04-24
OffScreen Performance Tweak

amaitland:
Attempt to improve efficiency of OffScreen rendering by only creating a new bitmap when the old one changes in size (same behavior as WPF).

Moved bitmap creation into GdiBitmapInfo

### DIFF
--- a/CefSharp.OffScreen.Example/Program.cs
+++ b/CefSharp.OffScreen.Example/Program.cs
@@ -6,7 +6,6 @@ using System;
 using System.Diagnostics;
 using System.Drawing;
 using System.IO;
-using System.Text;
 using System.Threading.Tasks;
 using CefSharp.Example;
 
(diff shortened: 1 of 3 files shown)


amaitland:
@jankurianski Do you mind having a look over this one when you get a chance and letting me know if I missed something?

I removed the lock as it's called within the `OnPaint` locking scope (unless I've totally missed something).

jankurianski:
You were right, the extra lock was unnecessary as `OnPaint` already locks the same object.  The use of `LoadingStateChanged` in the example is a good idea, that way it'll still work upon errors rather than freezing. Tested it out and all working well :+1:

amaitland:
@jankurianski Appreciate you having a look :+1: 

Probably need to expand the example a little bit at some point.

jankurianski:
> Probably need to expand the example a little bit at some point.

I think that's a good idea.  If I get a feel for the most common questions, I'll update the example.  Though I think your use of `LoadingStateChanged` and the efficiency fix should address the most common question to date (blank or early screenshot being taken).

amaitland:
> Though I think your use of LoadingStateChanged and the efficiency fix should address the most common question to date (blank or early screenshot being taken).

I think maybe something that provides a little detail about rendering and that it's not a finite process, specifically that there's no easy way to tell when it's complete.

jankurianski:
> I think maybe something that provides a little detail about rendering and that it's not a finite process, specifically that there's no easy way to tell when it's complete.

Perhaps an example could demonstrate this too.  Something like visiting a HTML5 animation demo page and either using a fixed delay or polling via `EvaluateScriptAsync` to recognise when the animation is complete.  Hopefully the render will have been received before JavaScript evaluation returns.

amaitland:
> Perhaps an example could demonstrate this too. Something like visiting a HTML5 animation demo page and either using a fixed delay or polling via EvaluateScriptAsync to recognise when the animation is complete. Hopefully the render will have been received before JavaScript evaluation returns.

Excellent idea :smile:

PR #969, opened 2015-04-24
Propagate TaskWrapper exceptions somewhere

rassilon:
Log an ERROR message via CEF logging, and throw an appropriate exception on thread pool thread.

The idea is that the AppDomain's UnhandledException or Application.ThreadException handler will get called.

This PR is for discussion, I'll be testing out the approach with a non-CEF sample app in a bit.

### DIFF
--- a/CefSharp.Core/Internals/CefTaskWrapper.h
+++ b/CefSharp.Core/Internals/CefTaskWrapper.h
@@ -62,7 +62,7 @@ namespace CefSharp
                     // AppDomain::UnhandledException event handler will get called 
                     // with the exception.
                     ThreadPool::UnsafeQueueUserWorkItem(
-                        gcnew WaitCallback(&CefTaskWrapper::ReportUnhanledException),
+                        gcnew WaitCallback(&CefTaskWrapper::ReportUnhandledException),
                         static_cast<Object^>(e));
                 }
             };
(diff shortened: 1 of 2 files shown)


CefSharp.Core/Internals/CefTaskWrapper.h
```diff
@@ -18,6 +20,15 @@ namespace CefSharp
     {
         private class CefTaskWrapper
         {
+        private:
+            // Throw the exception, this method executes on the thread pool
+            // in order to report a .net exception from a CEF thread.
+            static void ReportUnhanledException(Object ^state)
```
amaitland: Can the `Exception^` object it's self be passed through and rethrown?

jankurianski:
Good idea.

Would it be possible to pass the exception `e` instead of the `msg` to the threadpool worker, so it could  `throw new InvalidOperationException("Unexpected CefTaskWrapper exception", e);` and so retain the original exception object and stacktrace via the `InnerException`?

CefSharp.Core/Internals/CefTaskWrapper.h
jankurianski: Typo in "Unhan**d**led"

amaitland:
Works as you'd hoped I take it?

rassilon:
Yep. Sample app worked as expected.

amaitland:
Nice :+1: I'd say it's worth extracting into a helper function that takes the exception. Quite a few places in the `API` currently that just swallow exceptions, so could be very useful.

rassilon:
I suppose. Any particular place in mind? or a brand new MiscUtilities.h file? :smile:

amaitland:
> or a brand new MiscUtilities.h file?

Sure :smile: Maybe a slightly better name though :stuck_out_tongue:

PR #970, opened 2015-04-24
Add Visual Studio 2015 Preview support

Erguotou:
Added .vs/ to .gitignore and added
<PlatformToolset>v120</PlatformToolset> to the two C++ projects.

You can now open and build in Visual Studio 2015 CTP 6 as long as you
have VS 2013 installed as well. VS 2013 still works fine.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,4 @@ CefSharp.BrowserSubprocess.lib
 Win32/Release
 x64/Release
 Win32/Release
+.vs/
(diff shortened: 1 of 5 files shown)


amaitland:
@Erguotou Appreciate your efforts! It's best to discuss some of this before implementing as `CefSharp` does things a little different because of the `VC++` dependency.

See https://github.com/cefsharp/CefSharp/blob/master/CefSharp.props

> Added .vs/ to .gitignore and added

What's the purpose of this? Does `VS2015` generate some different files? Please be more descriptive.

amaitland:
Should be a simple matter of including `VS2015` in the props file.

PR #972, opened 2015-04-25
Add late binding to Javascript object and array

Erguotou:
Added late binding to JavaScript object, so that objects with cyclic
reference don't cause stack overflow (updated example to prove that).

For consistency, modified method calls to return a true object, just
like property calls.

Add IsNull property to JavascriptObject  to indicate if the object is
null, so that on browser side we don't need to create a CEF object.


CefSharp.BrowserSubprocess.Core/JavascriptMethodHandler.cpp
```diff
@@ -59,39 +60,6 @@ namespace CefSharp
 
     CefRefPtr<CefV8Value> JavascriptMethodHandler::ConvertToCefObject(Object^ obj)
```
amaitland: This method can probably be removed and a call to `ConvertToCef` made directly.

CefSharp.BrowserSubprocess.Core/JavascriptMethodWrapper.cpp
```diff
@@ -21,6 +22,13 @@ namespace CefSharp
 
     BrowserProcessResponse^ JavascriptMethodWrapper::Execute(array<Object^>^ parameters)
     {
-        return _browserProcess->CallMethod(_ownerId, _javascriptMethod->JavascriptName, parameters);
+        auto resp = _browserProcess->CallMethod(_ownerId, _javascriptMethod->JavascriptName, parameters);
```
amaitland: rename to `response`

CefSharp.BrowserSubprocess.Core/JavascriptMethodWrapper.cpp
```diff
@@ -21,6 +22,13 @@ namespace CefSharp
...
+        if (resp->Result->GetType() == JavascriptObject::typeid) {
```
amaitland: braces on new lines as a general rule.

CefSharp.BrowserSubprocess.Core/JavascriptObjectWrapper.cpp
```diff
@@ -12,8 +12,19 @@ using namespace CefSharp::Internals;
 
 namespace CefSharp
 {
-	void JavascriptObjectWrapper::Bind()
+    void JavascriptObjectWrapper::Bind()
```
amaitland: Spaces instead of tabs (probably affects a few files I believe).

CefSharp.BrowserSubprocess.Core/JavascriptObjectWrapper.cpp
```diff
@@ -12,8 +12,19 @@ using namespace CefSharp::Internals;
...
 	{
+        auto objectName = StringUtils::ToNative(_object->JavascriptName);
+
+        if (_object->IsNull) {
+            auto javascriptObject = V8Value.get() ? V8Value->CreateNull() : CefV8Value::CreateNull();        
+            if (V8Value.get() && !V8Value->HasValue(objectName))
```
amaitland: Single like blocks still need braces.

CefSharp.BrowserSubprocess.Core/JavascriptObjectWrapper.cpp
```diff
@@ -12,8 +12,19 @@ using namespace CefSharp::Internals;
 
 namespace CefSharp
 {
-	void JavascriptObjectWrapper::Bind()
+    void JavascriptObjectWrapper::Bind()
 	{
+        auto objectName = StringUtils::ToNative(_object->JavascriptName);
+
+        if (_object->IsNull) {
```
amaitland: Braces on new line.

CefSharp.BrowserSubprocess.Core/JavascriptObjectWrapper.cpp
```diff
@@ -12,8 +12,19 @@ using namespace CefSharp::Internals;
...
+            auto javascriptObject = V8Value.get() ? V8Value->CreateNull() : CefV8Value::CreateNull();
```
amaitland: `CreateNull()` is a static function, so can just be `CefV8Value::CreateNull()`

amaitland: Should probably replace all instance calls with the static method on `CefV8Value` for clarity, there's a few historic cases that should be cleaned up.

CefSharp.BrowserSubprocess.Core/JavascriptObjectWrapper.cpp
```diff
@@ -40,13 +52,59 @@ namespace CefSharp
...
 
-			_wrappedProperties->Add(wrappedproperty);
+            _wrappedProperties->Add(prop->JavascriptName, wrappedproperty);
 		}
 	}
 
 	BrowserProcessResponse^ JavascriptObjectWrapper::GetProperty(String^ memberName)
 	{
-		return _browserProcess->GetProperty(_object->Id, memberName);
+		auto resp = _browserProcess->GetProperty(_object->Id, memberName);
```
amaitland: General formatting including, spaces, braces on new lines. More expressive variable naming. 
I won't point out each case, I'll let you clean it up as a whole.

CefSharp.BrowserSubprocess.Core/TypeUtils.cpp
```diff
@@ -30,6 +31,11 @@ namespace CefSharp
             type = obj->GetType();
         }
 
+        if (type == JavascriptObjectWrapper::typeid) {
```
amaitland: At first glance this looks somewhat out of place. Will have to look at in more detail.

CefSharp.Example/BoundObject.cs
```diff
@@ -11,13 +11,48 @@ public class BoundObject
...
 
+        public SubBoundObject[] MyObjects
+        {
+            get
+            {
+                return new SubBoundObject[] {
+                    new SubBoundObject() {SimpleProperty="Hello"}, new SubBoundObject() {SimpleProperty="CefSharp"}
+                };
+            }
+            set { }
```
amaitland: Remove the empty set. Same goes for other properties.

CefSharp/Internals/JavascriptObject.cs
```diff
@@ -10,6 +11,9 @@ namespace CefSharp.Internals
     [DataContract]
     public class JavascriptObject //: DynamicObject maybe later
     {
+        private bool bound = false;
+        private object value = null;
```
amaitland: Avoid using `value` as a property name, it being a work with special meaning.

amaitland:
@Erguotou I've only comment on style/consistency issues so far.

As a group I think this needs discussion, whilst as a whole it gives greater flax-ability, changing the functional behavior needs to be reviewed in meticulous detail. Performance implications also need to be considered.

amaitland:
@Bodekaer Any chance you or someone on your team could take a look at this one? I suspect it may impact you guys dramatically.

amaitland:
@rassilon @jankurianski @jornh @perlun I'd appreciate any feedback from you guys also.

amaitland:
@Erguotou Cherry picked your first two commits and they're been merged into `master` and `2272` :+1: 

Minor changes made https://github.com/cefsharp/CefSharp/commit/c24607286b1bbc40a2ce64f62b96d9dac6df9d55

Bodekaer:
Definitely. Thanks Alex.
FYI @Antonyo and @myrnazhou, looks like this will impact our app API significantly, and potentially mail editor. Hopefully in a good way :)

Erguotou:
Updated the commit per Alex's comments. Thanks, Alex!

P.S. I only updated 12 files. Not sure why it says there are 198 files changed.

rassilon:
I did a copyright update pass to 2272, no need for you to worry.

amaitland:
> I did a copyright update pass to 2272, no need for you to worry.

Ahh I see. @Erguotou You'll need to merge in the latest `2272` branch.

CefSharp.BrowserSubprocess.Core/JavascriptMethodHandler.cpp
```diff
@@ -5,6 +5,7 @@
 #include "Stdafx.h"
 #include "TypeUtils.h"
 #include "JavascriptMethodHandler.h"
+#include "JavascriptObjectWrapper.h"
```
amaitland: Is this additional include required?

CefSharp.BrowserSubprocess.Core/JavascriptObjectWrapper.cpp
```diff
@@ -40,13 +62,72 @@ namespace CefSharp
...
     {
-        return _browserProcess->GetProperty(_object->Id, memberName);
+        auto response = _browserProcess->GetProperty(_object->Id, memberName);
+
+        auto type = response->Result->GetType();
+        if (type == JavascriptObject::typeid)
+        {
+            auto obj = safe_cast<JavascriptObject^>(response->Result);
+            JavascriptPropertyWrapper^ propWrapper;
+            bool exits = false;
```
amaitland: `exists`?

amaitland:
> Updated the commit per Alex's comments. Thanks, Alex!

Thanks, looking much cleaner :smile:

@Erguotou Have you got any thoughts on the performance implications of these changes?

amaitland:
@Erguotou It looks like your making changes to this `PR`, any chance you can make them as additional commits. Small granular commits are preferable :+1:

amaitland:
(Otherwise we have no idea what changed)

Erguotou:
Updated per Alex's latest comments.

Performance wise, I think it's comparable for small objects. Late binding is essentially the same as where objects implement all their API as methods. One can hardly call that a bad design and a performance drag. Granted, there will be more round-trip in WCF to get the properties, but I don't think this can be the bottleneck. For large objects, one should see tremendous performance gain. Imaging you have an object that represents a company's org-chart. For large companies, in early binding that could serialize to MBs of XML, and most likely you don't want all of it on the browser side. Also imaging that you have an object that represents hardware. To save battery the hardware is powered down. You can only read the on/off property. Reading the rest will either block, or may cause bus error. In early binding you have no control.

Erguotou:
@amaitland Yes I was changing this PR to avoid small commits. I was told by another groupd that commits that change previous commit is called "churn" and is basically considered undesirable as adds unnecessary fluff to the record.

The last change was rename exists to same, and remove #include "JavascriptObjectWrapper.h" in JavascriptMethodHandler.cpp. Thanks for catching that.

amaitland:
> I was told by another groupd that commits that change previous commit is called "churn" and is basically considered undesirable as adds unnecessary fluff to the record.

I have no idea how they expect to perform a code review with a single commit. As per https://github.com/cefsharp/CefSharp/blob/master/CONTRIBUTING.md#pull-requestsfeature-branches

`Small (really, minimalistic) commits. Each individual commit only adds one specific thing. The basic approach to achieving this is to read your commit message. Do you feel the need to add multiple lines? Then you're doing too much at the same time.`

Erguotou:
@amaitland Interestingly, I was told of that by a member in the Linux kernel group. 

I have the previous commit saved locally. Do you want me to update it and check in the two changes?

amaitland:
> Performance wise, I think it's comparable for small objects.

Hopefully someone from `Mailbird` will take a look and provide some feedback on how this performs in a real world scenario. Just for clarity, I think it's fantastic that you've taken the time to implement this :+1: I just need to make sure that there are no unforeseen side effects for those using the existing implementation and performance is either better or within an acceptable margin.

amaitland:
> @amaitland Interestingly, I was told of that by a member in the Linux kernel group.

That's their choice. Would make it very hard to verify that feedback was actioned as there is no before and after. I guess when you have more than half a million commits you look to limit the number. For us in the low thousands, the granularity is welcomed.

> I have the previous commit saved locally. Do you want me to update it and check in the two changes?

Just going forward is fine.

CefSharp.BrowserSubprocess.Core/JavascriptObjectWrapper.cpp
```diff
@@ -14,6 +14,22 @@ namespace CefSharp
...
+        if (_object->IsNull)
+        {
+            auto javascriptObject = CefV8Value::CreateNull();
+            if (V8Value.get() && !V8Value->HasValue(objectName))
+            {
+                V8Value->SetValue(objectName, javascriptObject, V8_PROPERTY_ATTRIBUTE_NONE);
+            }
+            else
+            {
+                V8Value = javascriptObject;
```
amaitland: What scenario is this condition met? Setting the `V8Value` to a null version seems strange. Code block in general needs some descriptive comments about what it's purpose is.

CefSharp.BrowserSubprocess.Core/JavascriptObjectWrapper.cpp
```diff
@@ -21,9 +37,15 @@ namespace CefSharp
...
-        auto objectName = StringUtils::ToNative(_object->JavascriptName);
-        V8Value->SetValue(objectName, javascriptObject, V8_PROPERTY_ATTRIBUTE_NONE);
+        auto javascriptObject = CefV8Value::CreateObject(_jsPropertyHandler.get());
+        if (V8Value.get() && !V8Value->HasValue(objectName))
+        {
+            V8Value->SetValue(objectName, javascriptObject, V8_PROPERTY_ATTRIBUTE_NONE);
+        }
+        else
+        {
+            V8Value = javascriptObject;
```
amaitland: Again, needs comment what purpose this serves? Seems strange. Comment required.

CefSharp.BrowserSubprocess.Core/JavascriptObjectWrapper.cpp
```diff
@@ -40,13 +62,72 @@ namespace CefSharp
...
 
-            _wrappedProperties->Add(wrappedproperty);
+            _wrappedProperties->Add(prop->JavascriptName, wrappedproperty);
         }
     }
 
     BrowserProcessResponse^ JavascriptObjectWrapper::GetProperty(String^ memberName)
     {
-        return _browserProcess->GetProperty(_object->Id, memberName);
+        auto response = _browserProcess->GetProperty(_object->Id, memberName);
```
amaitland: This block as a whole is way too complex. Does too many things, needs to be split where possible. Possibly a helper method or two. Desperately needs comments as the single line that was before has now grown into ~50

CefSharp.BrowserSubprocess.Core/JavascriptPropertyWrapper.h
```diff
@@ -17,9 +17,9 @@ namespace CefSharp
         JavascriptProperty^ _javascriptProperty;
         int64 _ownerId;
         IBrowserProcess^ _browserProcess;
+    internal:
         //TODO: Strongly type this variable - currently trying to include JavascriptObjectWrapper.h creates a circular reference, so won't compile
         Object^ _javascriptObjectWrapper;
```
amaitland: Making this property internal means that it also needs to be renamed. Prefix with `_` is only for private variables.

CefSharp/Internals/JavascriptObject.cs
```diff
@@ -40,9 +44,40 @@ public class JavascriptObject //: DynamicObject maybe later
...
+                LateBinding();
+            bound = true;
+            return this;
+        }
+
+        /// <summary>
         /// Gets or sets the value.
         /// </summary>
-        public object Value { get; set; }
+        public object Value
```
amaitland: Would prefer to keep this as an auto property and have `IsNull` return Value == null.

CefSharp/Internals/JavascriptObjectRepository.cs
```diff
@@ -119,6 +119,13 @@ public bool TryGetProperty(long objectId, string name, out object result, out st
                 throw new InvalidOperationException(string.Format("Property {0} not found on Object of Type {1}", name, obj.Value.GetType()));
             }
 
+            if (property.JsObject != null)
+            {
+                obj = property.JsObject.Bind();
+                result = !obj.IsNull && obj.Value.GetType().IsArray ? obj.Value : obj;
```
amaitland: Be nicer to have a property for `IsArray` similar to `IsNull`

CefSharp/Internals/JavascriptObjectRepository.cs
```diff
@@ -164,13 +171,11 @@ public bool TrySetProperty(long objectId, string name, object value, out string
         /// <summary>
         /// Analyse the object and generate metadata which will
         /// be used by the browser subprocess to interact with Cef.
-        /// Method is called recursively
+        /// Method is NOT called recursively and use late binding instead
```
amaitland: Just remove this line in the comment. Doesn't really make sense to say what it doesn't do.

CefSharp/Internals/JavascriptObjectRepository.cs
```diff
@@ -211,15 +213,31 @@ private void AnalyseObjectForBinding(JavascriptObject obj, bool analyseMethods,
                     var jsObject = CreateJavascriptObject(camelCaseJavascriptNames);
                     jsObject.Name = propertyInfo.Name;
                     jsObject.JavascriptName = GetJavascriptName(propertyInfo.Name, camelCaseJavascriptNames);
-                    jsObject.Value = jsProperty.GetValue(obj.Value);
+                    jsObject.LateBinding = () =>
+                    {
+                        if (!propertyInfo.PropertyType.IsArray)
```
amaitland: Switch logic in `if` statement, preferable to eval  `==` rather than `!=`, reads just that little bit easier.

CefSharp/Internals/JavascriptObjectRepository.cs
```diff
@@ -211,15 +213,31 @@ private void AnalyseObjectForBinding(JavascriptObject obj, bool analyseMethods,
...
+                        {
+                            jsObject.Value = jsProperty.GetValue(obj.Value);
+                            AnalyseObjectForBinding(jsObject, camelCaseJavascriptNames);
+                        }
+                        else
+                        {
+                            Array array = jsProperty.GetValue(obj.Value) as Array;
```
amaitland: If your not doing to check for null (probably isn't required), then just straight cast rather than `as Array`

CefSharp.BrowserSubprocess.Core/TypeUtils.cpp
```diff
@@ -94,6 +95,11 @@ namespace CefSharp
         {
             return CefV8Value::CreateDate(TypeUtils::ConvertDateTimeToCefTime(safe_cast<DateTime>(obj)));
         }
+        if (type == JavascriptObjectWrapper::typeid)
```
amaitland: Is there a better place for this conversion? Something about this just doesn't quite sit right.

CefSharp.BrowserSubprocess.Core/JavascriptObjectWrapper.cpp
```diff
@@ -40,13 +62,72 @@ namespace CefSharp
...
+            response->Result = propWrapper->_javascriptObjectWrapper;
+        }
+        else if (type->IsArray && type->GetElementType() == JavascriptObject::typeid)
+        {
+            auto array = safe_cast<Array^>(response->Result);
+            CefRefPtr<CefV8Value> cefArray = CefV8Value::CreateArray(array->Length);
+            for (int i = 0; i < array->Length; i++)
+            {
+                JavascriptObject^ jsObj = safe_cast<JavascriptObject^>(array->GetValue(i));
+                if (jsObj != nullptr)
```
amaitland: switch to `jsObj == nullptr` and adjust code accordingly.

CefSharp.BrowserSubprocess.Core/JavascriptObjectWrapper.cpp
```diff
@@ -40,13 +62,72 @@ namespace CefSharp
...
+                {
+                    auto objWrapper = gcnew JavascriptObjectWrapper(jsObj, _browserProcess);
+                    objWrapper->Bind();
+                    cefArray->SetValue(i, objWrapper->V8Value.get());
+                }
+                else
```
amaitland: Missing braces

CefSharp.BrowserSubprocess.Core/JavascriptObjectWrapper.cpp
```diff
@@ -40,13 +62,72 @@ namespace CefSharp
...
+                    auto objWrapper = safe_cast<JavascriptObjectWrapper^>(propWrapper->_javascriptObjectWrapper);
+                    same = obj->Id == objWrapper->_object->Id;
+                }
+            }
+            if (!same)
+            {
+                auto jsObjectWrapper = gcnew JavascriptObjectWrapper(obj, _browserProcess);
+                jsObjectWrapper->V8Value = propWrapper->V8Value.get();
+                jsObjectWrapper->Bind();
+                if (propWrapper->_javascriptObjectWrapper != nullptr)
```
amaitland: No need for `if` block, deleting a `nullptr` is completely safe.

CefSharp.BrowserSubprocess.Core/JavascriptObjectWrapper.cpp
```diff
@@ -40,13 +62,72 @@ namespace CefSharp
...
+                    cefArray->SetValue(i, objWrapper->V8Value.get());
+                }
+                else
+                    cefArray->SetValue(i, CefV8Value::CreateNull());
+            }
+
+            JavascriptPropertyWrapper^ propWrapper;
+            if (_wrappedProperties->TryGetValue(memberName, propWrapper))
+            {
+                if (propWrapper->_javascriptObjectWrapper != nullptr)
```
amaitland: No need for `if`, just call `delete`

amaitland:
@Erguotou I spent a little time stepping through the code. I've made additional comments inline. As the complexity has increased, would you mind going through and adding comments please. As an overall I think we also need to add some notes around how late binding actually works in this case. Maybe in `JavascriptObjectRepository`?

Let me know if you have any questions/comments.

CefSharp/Internals/JavascriptObject.cs
Erguotou: It won't work because Value is always null on the browser side. It's not transmitted over the wire.

amaitland: Fair enough.

CefSharp/Internals/JavascriptObjectRepository.cs
Erguotou: Are you sure you want to do that? Type casting is a lot more expensive than just checking the flag. You did remind me that I need to check if array is null, else array.Length will throw an exception.

amaitland: Perhaps I wasn't clear, regardless if your going to check for null then the `as` is perfect.

CefSharp.BrowserSubprocess.Core/JavascriptMethodWrapper.cpp
```diff
@@ -21,6 +22,14 @@ namespace CefSharp
 
     BrowserProcessResponse^ JavascriptMethodWrapper::Execute(array<Object^>^ parameters)
     {
-        return _browserProcess->CallMethod(_ownerId, _javascriptMethod->JavascriptName, parameters);
+        auto response = _browserProcess->CallMethod(_ownerId, _javascriptMethod->JavascriptName, parameters);
+        if (response->Result->GetType() == JavascriptObject::typeid)
```
rassilon: What if Result is null?

CefSharp.BrowserSubprocess.Core/JavascriptObjectWrapper.cpp
```diff
@@ -40,13 +62,72 @@ namespace CefSharp
...
+            _wrappedProperties->Add(prop->JavascriptName, wrappedproperty);
         }
     }
 
     BrowserProcessResponse^ JavascriptObjectWrapper::GetProperty(String^ memberName)
     {
-        return _browserProcess->GetProperty(_object->Id, memberName);
+        auto response = _browserProcess->GetProperty(_object->Id, memberName);
+
+        auto type = response->Result->GetType();
```
rassilon: Can't Result be NULL here as well?

CefSharp.BrowserSubprocess.Core/JavascriptMethodWrapper.cpp
Erguotou: You are right. Need to check if success is true as well.

CefSharp.BrowserSubprocess.Core/JavascriptObjectWrapper.cpp
Erguotou: Yes, need to check it and response->Success.

CefSharp/Internals/JavascriptObjectRepository.cs
Erguotou: OK to add IsArray, but no need to serialize it. In fact, if is array, the object is not passed to browser, only value which is the array is.

CefSharp.BrowserSubprocess.Core/JavascriptObjectWrapper.cpp
Erguotou:
```c++
(V8Value.get() && !V8Value->HasValue(objectName)) is true only when called from JavascriptRootObjectWrapper.cpp.

Setting V8Value to a null version allows JavaScript to get a null value. Before I added this, I got an empty object {}.

Comment (amaitland) [inline CefSharp.BrowserSubprocess.Core/JavascriptObjectWrapper.cpp]: I think some of this becomes hard to follow, the approach you've taken might not quite suite the existing model, so should look at refactoring.

Regardless code needs comments.

File: CefSharp/Internals/JavascriptObjectRepository.cs (hunk 19 above)
Comment (amaitland) [inline CefSharp/Internals/JavascriptObjectRepository.cs]: Can you clarify, your second sentence isn't clear.

Code needs comment to document behavior.

Somewhat sounds like array should be treated differently and the model should include first class support, to make the behavior clearer.

**Comment by amaitland** [comment]
@Erguotou comments on the PR I created against your branch?

**Comment by amaitland** [comment]
I think there is a scenario this should support. So

- Access an object that's `null` in javascript
- At some point later that object is assigned
- Attempt to access the object again in javascript and it's still `null`

I don't believe this scenario worked previously, if we're going to the trouble of adding late binding then I think it should.

As part of this we should probably also look at implementing #602

**Comment by Erguotou** [comment]
Good point. That means we cannot cache the value in JavaScriptObject. It can be done by ignoring the bound flag and calling late binding every time. It should not be a performance problem because we are only retrieving one property. It has another benefit that if later on you assign the object to an object of subclass, any new method or property defined  in the subclass will be available to JavaScript as well. It’s dynamic.






From: Alex Maitland
Sent: ‎Thursday‎, ‎May‎ ‎7‎, ‎2015 ‎7‎:‎15‎ ‎PM
To: cefsharp/CefSharp
Cc: Jerome Liang





I think there is a scenario this should support. So
Access an object that's null in javascript
At some point later that object is assigned
Attempt to access the object again in javascript and it's still null 

I don't believe this scenario worked previously, if we're going to the trouble of adding late binding then I think it should.

As part of this we should probably also look at implementing #602

**Comment by Erguotou** [comment]
The check for null is no longer there. I merged your commits.

File: CefSharp/Internals/JavascriptTypeBinder.cs [hunk 30]
```diff
@@ -0,0 +1,398 @@
+﻿using System;
```
amaitland: Add license

CefSharp/Internals/JavascriptTypeBinder.cs
```diff
@@ -0,0 +1,398 @@
...
+using System.Globalization;
+using System.Linq;
+using System.Reflection;
+using System.Text;
+
+namespace CefSharp.Internals
+{
+    internal class JavascriptTypeBinder : Binder
+    {
+        private class BinderState
```
amaitland: Classes in their own files please.

CefSharp/Internals/JavascriptTypeBinder.cs
```diff
@@ -0,0 +1,398 @@
...
+        class Converter<F, T> : IConvert
+        {
+            Func<F, T> convert;
+
+            public Converter(Func<F, T> func)
+            {
+                this.convert = func;
+                // Add this to JavascriptTypeBinder.explicitConversions
+                var type = typeof(F);
+                if (!explicitConversions.ContainsKey(type))
```
amaitland: Braces

CefSharp/Internals/JavascriptTypeBinder.cs
```diff
@@ -0,0 +1,398 @@
...
+        /// <see cref="https://msdn.microsoft.com/en-us/library/y5b434w4%28v=vs.140%29.aspx"/>
+        static Dictionary<TypeCode, TypeCode[]> implicitConversions = new Dictionary<TypeCode, TypeCode[]>();
+        /// <summary>
+        /// Explicit Numeric Conversions Table
+        /// </summary>
+        static Dictionary<Type, Dictionary<Type, IConvert>> explicitConversions = new Dictionary<Type, Dictionary<Type, IConvert>>();
+
+        static JavascriptTypeBinder()
+        {
+            implicitConversions[TypeCode.SByte] = new TypeCode[] {
```
amaitland: Braces on new line)

CefSharp/Internals/JavascriptTypeBinder.cs
```diff
@@ -0,0 +1,398 @@
...
+                {
+                    // Determine whether the return type can be converted to the properties type. 
+                    if ((matchReturnType && returnType == match[i].PropertyType) || (!matchReturnType && CanConvertFrom(returnType, match[i].PropertyType)))
+                    {
+                        return match[i];
+                    }
+                    else
+                    {
+                        continue;
+                    }
```
amaitland: If statement returns, no need for this else continue block.

amaitland:
@Erguotou Out of curiosity do you use `Resharper`?

CefSharp/Internals/JavascriptTypeBinder.cs
```diff
@@ -0,0 +1,398 @@
...
+                }
+            }
+            return null;
+        }
+
+        // Determines whether type1 can be converted to type2. Check only for primitive types. 
+        private bool CanConvertFrom(Type type1, Type type2)
```
amaitland: Make `static`

CefSharp/Internals/JavascriptTypeBinder.cs
```diff
@@ -0,0 +1,398 @@
...
+                    }
+                }
+                // Determine whether the method has been found. 
+                if (count == types.Length)
+                {
+                    return match[i];
+                }
+            }
+            return null;
+        }
```
amaitland: Add line

CefSharp/Internals/JavascriptTypeBinder.cs
```diff
@@ -0,0 +1,398 @@
...
+            // Get a field for which the value parameter can be converted to the specified field type. 
+            for (int i = 0; i < match.Length; i++)
+            {
+                if (ChangeType(value, match[i].FieldType, culture) != null)
+                {
+                    return match[i];
+                }
+            }
+            return null;
+        }
```
amaitland: Add line

CefSharp/Internals/JavascriptObjectRepository.cs
```diff
@@ -259,15 +329,48 @@ private static JavascriptMethod CreateJavaScriptMethod(MethodInfo methodInfo, bo
             return jsMethod;
         }
 
-        private static JavascriptProperty CreateJavaScriptProperty(PropertyInfo propertyInfo, bool camelCaseJavascriptNames)
+        private object ConvertIfGenericList(object value, Type type, bool camelCaseJavascriptNames)
+        {
+            if (!type.IsGenericType) return value;
+
+            IList list = value as IList;
+            if (list != null)
```
amaitland: Simplify this, evaluate `null` and throw exception straight away, reduces nesting.

CefSharp/Internals/JavascriptObjectRepository.cs
```diff
@@ -234,18 +237,85 @@ private void AnalyseObjectForBinding(JavascriptObject obj, bool analyseMethods,
...
+            }
+        }
 
-                    AnalyseObjectForBinding(jsProperty.JsObject, analyseMethods, readPropertyValue, camelCaseJavascriptNames);
+        private void SetJavascriptObjectValue(JavascriptObject jsObject, object value, Type type, bool camelCaseJavascriptNames)
+        {
+            if (type.IsGenericType)
+            {
+                IList list = value as IList;
+                if (list != null)
```
amaitland: Simplify this, evaluate `null` and throw exception straight away, reduces nesting.

CefSharp/Internals/JavascriptObjectRepository.cs
```diff
@@ -30,6 +33,7 @@ public class JavascriptObjectRepository : DisposableResource
     {
         private static long lastId;
         private static readonly object Lock = new object();
+        private static readonly BindingFlags DefaultBindingFlags = BindingFlags.Instance | BindingFlags.Public;
```
amaitland: Can this be a `const`?

CefSharp/Internals/JavascriptObject.cs
```diff
@@ -17,6 +18,11 @@ namespace CefSharp.Internals
     public class JavascriptObject //: DynamicObject maybe later
     {
         /// <summary>
+        /// Indicates if this <see cref="JavascriptObject"/> represents an Array
+        /// </summary>
+        private bool _isArray;
```
amaitland: Remove `_`, only use `_` in `c++` files.

CefSharp.Example/SubBoundObject.cs
```diff
@@ -2,6 +2,9 @@
 {
 	public class SubBoundObject
 	{
+        // setting Parent will cause stack overflow if not for late binding
+        public BoundObject Parent { get; set; }
```
amaitland: Spaces not tabs

CefSharp.Example/Resources/BindingTest.html
```diff
@@ -124,5 +124,59 @@
...
+            </script>
+        </ul>
+
+        bound subobject's parent before and after calling bound.toggleSubObjectParent(): <br/>
+        <ul>
+            <script type="text/javascript">
+                var so = bound.subObject;
+                for (var i = 0; i < 3; i++)
+                {
+                    if (so.parent == null)
```
amaitland: Add braces

CefSharp.Example/BoundObject.cs
```diff
@@ -262,5 +274,30 @@ public SubBoundObject GetSubObject()
...
+        {
+            return MyObjects;
+        }
+
+        public List<SubBoundObject> GetMySubObjectList()
+        {
+            return new List<SubBoundObject>(MyObjects);
+        }
+
+        public class NestedSubClass : BoundObject
```
amaitland: Classes in their own files.

amaitland:
Much easier to follow with the new changes :+1: Ideally a couple of people using `JSB` in production will evaluate this before we merge.

CefSharp/Internals/JavascriptTypeBinder.cs
```diff
@@ -0,0 +1,398 @@
...
+                TypeCode.Double,
+            };
+            implicitConversions[TypeCode.UInt64] = new TypeCode[] {
+                TypeCode.Single,
+                TypeCode.Double,
+                TypeCode.Decimal,
+            };
+
+            // signed to unsigned conversion to silence overflow exception
+            new Converter<int, byte>((v) => (byte) v);
```
amaitland: These don't seem to be assigned to anything?

amaitland: Hmm, the new instances of `Converter` assign themselves to the dictionary. Very indirect approach, a more direct approach would be cleaner and easier I think.

CefSharp/Internals/JavascriptObjectRepository.cs
```diff
@@ -30,11 +34,14 @@ public class JavascriptObjectRepository : DisposableResource
...
         private static long lastId;
         private static readonly object Lock = new object();
+        private const BindingFlags DefaultBindingFlags = BindingFlags.Instance | BindingFlags.Public;
 
         // A hash from assigned object ids to the objects,
         // this is done to speed up finding the object in O(1) time
         // instead of traversing the JavaScriptRootObject tree.
         private readonly Dictionary<long, JavascriptObject> objects = new Dictionary<long, JavascriptObject>();
+        // objects are never removed so we need to try to avoid duplicates, other wise 1000 calls to bound.getSubObject() will create 1000 entries in objects.
+        private readonly Dictionary<object, JavascriptObject> reverseLookup = new Dictionary<object, JavascriptObject>();
```
amaitland: Is there a move descriptive name? `reverseLookup` seems like it's more of a `cache`. Can we make the comment more descriptive? Need to immediately differentiate between the `objects` dictionary, and this one without having to dig through the code to work out what each is for.

CefSharp/Internals/JavascriptObjectRepository.cs
```diff
@@ -80,40 +87,58 @@ public bool TryCallMethod(long objectId, string name, object[] parameters, out o
                 return false;
             }
 
-            var method = obj.Methods.FirstOrDefault(p => p.JavascriptName == name);
-            if (method == null)
+            var overloads = obj.Methods.Where(p => p.JavascriptName == name);
```
amaitland: Call `ToList()` here to avoid multiple `enumerations` later. Then change the `.Count()` references.

amaitland:
Closing this so I can delete the `2272` branch. Happy to revisit if your interested.

PR #980, opened 2015-04-28
Update WPF example to use Mvvm Light

amaitland:
Replace our basic MVVM implementation, with a well tested, known framework.

Whilst there are quite a few different `MVVM` frameworks to choose from, I think that `MVVM Light` is fairly non invasive and fits well with our demo application.


amaitland:
Need to merge in `2272` branch before merging.

PR #981, opened 2015-04-28
Implement CefTask

amaitland:
`CefTaskWrapper` now implements `CefTask` and can be passed directly into `CefPostTask` - replace 
`NewCefRunnableMethod`


amaitland:
@rassilon Any objections to this one? I think the code is cleaner.

rassilon:
Looks fine to me.

Issue #987, opened 2015-04-29
Create CefFrameWrapper


rassilon:
I've started stubbing out a CefFrameWrapper class so that eventually, JavaScript integration can access any frame.

rassilon:
The first pass is now in 2272.

PR #988, opened 2015-04-29
CefFrameWrapper implementation

rassilon:
Start stubbing out some implementations in CefFrameWrapper.h

(This doesn't build yet, don't bother trying... )

This is the PR related to #987.


amaitland:
I was thinking this one a little bit the other day, ideally we'd

- Create an `IFrame` interface, similar to `IRequest`
- Replace `isMainFrame` bool param with `IFrame` reference

amaitland:
As `41` is turning into a bit of an overhaul, I'd even consider removing the `Cut/Copy/Paste` methods from the `ChromiumWebBrowser` instances.

rassilon:
An `IFrame` interface would make sense. I'm not sure what you mean by `isMainFrame` bit though.

While I was doing my silly copying and pasting I was looking at `MCefRefPtr` and wondering: When should I be using this vs just using `CefRefPtr<>`. Sadly, the file lacks any useful comments.

amaitland:
> I'm not sure what you mean by isMainFrame bit though.

`IRequestHandler.OnBeforeBrowse` as an example, replace the `isMainFrame` bool with the `IFrame` interface.

> While I was doing my silly copying and pasting I was looking at MCefRefPtr and wondering: When should I be using this vs just using CefRefPtr<>. Sadly, the file lacks any useful comments.

I was under the impression you use `MCefRefPtr` in `ref class's`

amaitland:
`MCefRefPtr` is one of those `black magic` things that's always been there, so I can't provide any definitive information.

rassilon:
`MCefRefPtr` seems to be to help with C++/CLI compilation weirdness.

CefFrame has a GetBrowser() method. However, CefSharp uses `IWebBrowser` that conflates the top level browser window and the browser control host.

So, should GetBrowser return an `IBrowserWindow` interface that wraps JUST `CefBrowser`?

Bill

rassilon:
Oh, ick. I just realized, ClientAdapter.cpp's FocusHandler methods doesn't verify it's the main frame for the browser control. (and I'm sure similarily with the rest of them) Blech.

Bill

amaitland:
> MCefRefPtr seems to be to help with C++/CLI compilation weirdness.

Yep, you need it to host the `unmanaged` classes.

> CefFrame has a GetBrowser() method. However, CefSharp uses IWebBrowser that conflates the top level browser window and the browser control host.

I would have left `GetBrowser()` out from the implementation. Confuses things otherwise and I don't think it's strictly necessary, having a reference to `IWebBrowser` allows you to do the basics, load/stop/back/forward etc. Thoughts?

amaitland:
> Oh, ick. I just realized, ClientAdapter.cpp's FocusHandler methods doesn't verify it's the main frame for the browser control. (and I'm sure similarily with the rest of them) Blech.

Do the focus methods pass in a reference to `CefFrame`?

amaitland:
What do you think about removing `Cut/Copy/Paste/etc` from `IWebBrowser` and using the new Frame wrapper instead? Since `41` is turning into and overhaul.

rassilon:
> I would have left GetBrowser() out from the implementation. Confuses things otherwise and I don't think it's strictly necessary, having a reference to IWebBrowser allows you to do the basics, load/stop/back/forward etc. Thoughts?

I'm leaving it out for now, but it does make life complicated if we want to factor things out more like CEF. (i.e. allow CefBrowser ids to be publicly accessible for instance)

> Do the focus methods pass in a reference to CefFrame?

No, just CefBrowser.

> What do you think about removing Cut/Copy/Paste/etc from IWebBrowser and using the new Frame wrapper instead? Since 41 is turning into and overhaul.

We could do that later, sure. There're pretty harmless overloads for now.
We could mark them as Deprecated I think in 41 if we wanted as well.

rassilon:
Due to `IFrame` needing to invoke the `BrowserServiceHost` is why the whole `IBrowserAdapter` goo got added. Thoughts?

Bill

rassilon:
> I'm leaving it out for now, but it does make life complicated if we want to factor things out more like CEF. (i.e. allow CefBrowser ids to be publicly accessible for instance)

Without a CefBrowserWrapper we also pass in the wrong context for things like `OnBeforePopup` since the CefBrowser might not be the main CefBrowser.

amaitland:
I'd somewhat forgotten supporting popups, we don't handle that scenario well at all. In light of that we should wrap CefBrowser.

rassilon:
> I'd somewhat forgotten supporting popups, we don't handle that scenario well at all. In light of that we should wrap CefBrowser.

Should we rename CefBrowserWrapper to be CefSubBrowserAdapter and then write CefBrowserWrapper for IBrowser?

rassilon:
Or if not IBrowser, then something that helps distinguish between the concepts of IWebBrowser and the CefBrowserWrapper interface name.

rassilon:
@amaitland , Think I should push this set of changes into 2272 for testing? I did a bunch of work of moving this stuff forward by creating `IBrowser` as the interface wrapper of CefBrowser.

amaitland:
Go for it :+1:

PR #996, opened 2015-05-03
Part 1 of Wrap CefBrowser.

rassilon:
Move some more method code from .h to .cpp files.
Create `IBrowser` interface for use of `CefBrowser` instances by CefSharp users.
Expose PopupBrowsers via new method in ClientAdapter


CefSharp.Core/Internals/CefFrameWrapper.cpp
```diff
@@ -0,0 +1,235 @@
...
+{
+    _frame->ViewSource();
+}
+
+///
+// Retrieve this frame's HTML source as a string sent to the specified
+// visitor.
+///
+/*--cef()--*/
+void CefFrameWrapper::GetSource(IStringVisitor^ visitor)
```
amaitland: Probably nicer to wrap these in a `Task` like the `ChromiumWebBrowser` instances do, same for `GetText()`

CefSharp.Core/Internals/CefFrameWrapper.h
```diff
@@ -9,271 +9,198 @@
...
+
+            ///
+            // Execute a string of JavaScript code in this frame. The |script_url|
+            // parameter is the URL where the script in question can be found, if any.
+            // The renderer may request this URL to show the developer the source of the
+            // error.  The |start_line| parameter is the base line number to use for error
+            // reporting.
+            ///
+            /*--cef(optional_param=script_url)--*/
+            virtual void ExecuteJavaScriptAsync(String ^code, String^ scriptUrl, int startLine);
```
amaitland: Probably need an overload or two for this one, provide some sensible defaults.

CefSharp.Core/Internals/CefFrameWrapper.h
```diff
@@ -9,271 +9,198 @@
...
+
+            ///
+            // Returns the name for this frame. If the frame has an assigned name (for
+            // example, set via the iframe "name" attribute) then that value will be
+            // returned. Otherwise a unique name will be constructed based on the frame
+            // parent hierarchy. The main (top-level) frame will always have an empty name
+            // value.
+            ///
+            /*--cef()--*/
+            virtual String^ GetName();
```
amaitland: Do we make the `API` more `C#` like and change methods like this to a property with only a getter?

Not strictly required, there are a few inconsistencies in the `API` as a whole, so nice to have a clear direction going forward. Thoughts?

CefSharp/IBrowser.cs
```diff
@@ -0,0 +1,144 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace CefSharp
+{
+    /// <summary>
+    /// CefSharp interface for CefBrowser.
+    /// </summary>
+    public interface IBrowser
```
amaitland: Any further thoughts on `IBrowser` as a name? It's somewhat of a tricky one, just seems a little too close to `IWebBrowser`

CefSharp.Core/Internals/CefFrameWrapper.h
rassilon: I was thinking the same thing as well. Just made the first pass copy/paste. :smile:

CefSharp/IBrowser.cs
rassilon: We could rename `IWebBrowser` :smile: something like `IBrowserControl` maybe?

CefSharp.Core/Internals/CefFrameWrapper.cpp
rassilon: I don't have a problem with that.

CefSharp/IBrowser.cs
amaitland: Wonder what sort of havoc that would create? lol

CefSharp.Core/Internals/CefFrameWrapper.h
amaitland: > Just made the first pass copy/paste. :smile:

Yep, yep :smile:

CefSharp.Core/Internals/CefFrameWrapper.cpp
rassilon: Done.

CefSharp.Core/Internals/CefFrameWrapper.h
rassilon: Added an additional overload that only takes 'code' as input.

CefSharp.Core/Internals/CefFrameWrapper.h
rassilon: Done for Frame.

CefSharp/IBrowser.cs
rassilon: Too much probably.

rassilon: Just want to punt on the name `IBrowser` for now?
If so. I could push this PR through, and then make part 2 being updating the handlers to include providing `IBrowser`.

amaitland: Yep, go with that for now.

PR #1003, opened 2015-05-05
WPF - Handle case where Bitmap in WpfBitmapInfo is null

jamespearce2006:
The OnPaint call arrives on a thread other than the UI thread. This means it is possible that the InvokeRenderAsync method, which is dispatched back to the UI thread, will get called after the Dispose method. In this case, a NullReferenceException is thrown, because FileMappingHandle is no longer set.

### DIFF
--- a/CefSharp.Wpf/Rendering/InteropBitmapInfo.cs
+++ b/CefSharp.Wpf/Rendering/InteropBitmapInfo.cs
@@ -42,13 +42,13 @@ namespace CefSharp.Wpf.Rendering
         {
             var stride = Width * BytesPerPixel;
 
-            if (FileMappingHandle != IntPtr.Zero)
+            if (FileMappingHandle == IntPtr.Zero)
             {
-                Bitmap = (InteropBitmap)Imaging.CreateBitmapSourceFromMemorySection(FileMappingHandle, Width, Height, PixelFormat, stride, 0);
+                ClearBitmap();
             }
             else
             {
-                ClearBitmap();
+                Bitmap = (InteropBitmap)Imaging.CreateBitmapSourceFromMemorySection(FileMappingHandle, Width, Height, PixelFormat, stride, 0);
             }
 
             return Bitmap;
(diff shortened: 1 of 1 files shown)


CefSharp.Wpf/Rendering/InteropBitmapInfo.cs
```diff
@@ -31,14 +32,24 @@ public override void ClearBitmap()
...
+                Bitmap.Invalidate();
+            }
         }
 
         public override BitmapSource CreateBitmap()
         {
             var stride = Width * BytesPerPixel;
 
-            Bitmap = (InteropBitmap)Imaging.CreateBitmapSourceFromMemorySection(FileMappingHandle, Width, Height, PixelFormat, stride, 0);
+            if (FileMappingHandle != IntPtr.Zero)
```
amaitland: Prefer to eval `==` first, flows better when reading code.

CefSharp.Wpf/Rendering/InteropBitmapInfo.cs
```diff
@@ -31,14 +32,24 @@ public override void ClearBitmap()
...
         }
 
         public override BitmapSource CreateBitmap()
         {
             var stride = Width * BytesPerPixel;
 
-            Bitmap = (InteropBitmap)Imaging.CreateBitmapSourceFromMemorySection(FileMappingHandle, Width, Height, PixelFormat, stride, 0);
+            if (FileMappingHandle == IntPtr.Zero)
+            {
+                ClearBitmap();
```
amaitland: `CreateBitmap()` is usually called after the bitmap has already been cleared. In the scenario your trying to handle hear is it possible for `bitmap != null`?

Ideally would fail fast, have a code block like this as the first operation.
```c#
// Unable to create bitmap without valid File Handle (Most likely control is being disposed)
if (FileMappingHandle == IntPtr.Zero)
{
  return null;
}
```

jamespearce2006: I think my update fulfils your request. Let me know.

amaitland: Close :smile: I'll remove the else (redundant) and move the stride calculation after the `if` when it's eventually time to merge this :+1:

amaitland:
Applied a similar change to `WritableBitmapInfo` see https://github.com/cefsharp/CefSharp/commit/561507b87bebfd9fd0ec6c684ef7ef9caa7a9ee4

Issue #869, opened 2015-03-10
Feature Request - Ability to open Popup with parent browser set


Bodekaer:
Hey,

How can we create custom popups?

I found out, that we can overwrite the `ILifeSpan.OnBeforePopup()`, and create our own browser instance. 
But the issue we have atm. is that we don't know how to tell the new CEF instance, who the parent browser is.

Do you know how to do this?

(e.g. it's important for a Facebook or Google login popup, where it needs to tell the parent browser, that the login was successful via Javascript).

amaitland:
Perhaps setting the parent property of the `CefWindowInfo`?

http://magpcss.org/ceforum/apidocs3/projects/%28default%29/CefWindowInfo..parent.html

amaitland:
`SetAsPopup` is probably the go.
http://magpcss.org/ceforum/apidocs3/projects/%28default%29/CefWindowInfo.html#SetAsPopup%28cef_window_handle_t,constCefString&%29

Bodekaer:
Yeah, that's it! 
Great thanks.

FYI @Antonyo, we just need to add support for this in CEF when we have time. 
Should be fairly straightforward.

Antonyo:
Hi, I think what @Bodekaer meant was to set the parent popup to an existing ChromiumWebBrowser so in our `ILifeSpan.OnBeforePopup()` we can stop the creation of CEF popup and create our own window with a browser embedded simulating a popup. 
What we want to achieve is to set the parent of this embedded browser so the popup can reference it's parent browser.

As far as I know the cef.sdk does not provide a way to set the parent to an already created browser, without that option I tried to create a new method `ManagedCefBrowserAdapter.CreateOffscreenBrowser(IntPtr windowHandle, BrowserSettings^ browserSettings, String^ address)` which uses the `CefWindowInfo.setAsPopup` method but why exactly is the method `ChromiumWebBrowser.CreateOffscreenBrowser`being called for? I tried to called my method instead of it when I want to create a popup browser but still the popbrowser can't reference the parent.

amaitland:
> As far as I know the cef.sdk does not provide a way to set the parent to an already created browser, without that option I tried to create a new method

It makes sense to specify the parent on creation. I wouldn't expect to specify a parent for an existing instance.

> why exactly is the method ChromiumWebBrowser.CreateOffscreenBrowser being called for?

You'll have to clarify, I'm unsure what your asking.

amaitland:
[...]

PR #1015, opened 2015-05-07
Support for Windowless/Offscreen Popups

jamespearce2006:
Initial solution for #869.

This PR isn't production-ready, but rather a starting point for a discussion. I'm new to CefSharp, so I've likely made some naive, or wrong-headed, architectural assumptions.

I've updated the WPF sample app (and struggled getting the change to fit the MVVM pattern).
To see the results, run the Wpf sample app and create a new tab. The new tab defaults to a w3schools try-it session, where you can open and close a window via Javascript.


amaitland:
Can you fix the formatting, spaces instead of tabs, it's hard to read. If you look at the `Files Changed` tab above you'll see what I mean.

CefSharp.Core/ManagedCefBrowserAdapter.h
```diff
@@ -70,7 +70,12 @@ namespace CefSharp
             DisposableResource::DoDispose(isDisposing);
         };
 
+	internal:
+		MCefRefPtr<ClientAdapter> GetClientAdapter();
+
     public:
+
```
amaitland: Remove extra white space, keep changes to a minimum.

CefSharp.Wpf.Example/LifespanHandler.cs
```diff
@@ -0,0 +1,86 @@
...
+using System.Threading.Tasks;
+using System.Windows;
+using System.Windows.Threading;
+
+namespace CefSharp.Wpf.Example
+{
+    public class LifespanHandler : ILifeSpanHandler
+    {
+
+        MainWindow _owner;
```
amaitland: Only prefix with `_` in `c++` for consistency.

CefSharp.Wpf.Example/ViewModels/BrowserTabViewModel.cs
```diff
@@ -14,6 +14,7 @@ namespace CefSharp.Wpf.Example.ViewModels
 {
     public class BrowserTabViewModel : ViewModelBase
     {
+
```
amaitland: Remove whitespace changes, specially as no changes were made to this class.

CefSharp.Wpf.Example/Views/BrowserTabView.xaml
```diff
@@ -168,12 +168,15 @@
                     </Rectangle.Fill>
                 </Rectangle>
 
+                    <ContentControl Content="{Binding WebBrowser}"
+                                Opacity="{Binding ElementName=opacitySlider, Path=Value}" >
+                    <!--
                 <cefSharp:ChromiumWebBrowser x:Name="browser"
```
amaitland: Whilst your way of creating the `ChromiumWebBrowser` instance will obviously work, I think the extra level of indirection is not great for an example application.

CefSharp.Wpf/ChromiumWebBrowser.cs
```diff
@@ -459,11 +459,12 @@ private static void OnAddressChanged(DependencyObject sender, DependencyProperty
 
         protected virtual void OnAddressChanged(string oldValue, string newValue)
         {
-            if (ignoreUriChange)
+            if (ignoreUriChange || newValue == null)
             {
                 return;
             }
 
+            //CreateOffscreenBrowserWhenActualSizeChanged();
```
amaitland: Remove this.

CefSharp.Wpf/ChromiumWebBrowser.cs
```diff
@@ -459,11 +459,12 @@ private static void OnAddressChanged(DependencyObject sender, DependencyProperty
 
         protected virtual void OnAddressChanged(string oldValue, string newValue)
         {
-            if (ignoreUriChange)
+            if (ignoreUriChange || newValue == null)
```
amaitland: Slight change in behavior, whilst logical, will this affect existing users? To be honest I can't remember what passing `null` does, so would have to look further.

CefSharp/ILifeSpanHandler.cs
```diff
@@ -17,8 +17,10 @@ public interface ILifeSpanHandler
         /// <param name="width">width</param>
         /// <param name="height">height</param>
         /// <param name="noJavascriptAccess">value indicates whether the new browser window should be scriptable
+        /// <param name="newBrowser">A newly created browser that will host the popup</param>
         /// and in the same process as the source browser.</param>
         /// <returns>To cancel creation of the popup window return true otherwise return false.</returns>
+
```
amaitland: Remove extra lines

CefSharp/ILifeSpanHandler.cs
```diff
@@ -34,7 +36,13 @@ public interface ILifeSpanHandler
         /// the new browser window should be scriptable and in the same process as the
         /// source browser.
         /// </remarks>
-        bool OnBeforePopup(IWebBrowser browser, IFrame frame, string targetUrl, ref int x, ref int y, ref int width, ref int height, ref bool noJavascriptAccess);
+        bool OnBeforePopup(IWebBrowser browser, IFrame frame, string targetUrl, ref int x, ref int y, ref int width, ref int height, ref bool noJavascriptAccess, ref IWebBrowser newBrowser);
+
+        /// <summary>
+        /// Called after a new browser is created.
+        /// </summary>
+        /// <param name="browser"></param>
```
amaitland: Add comment (just copy and paste)

CefSharp.Wpf.Example/MainWindow.xaml
```diff
@@ -2,8 +2,20 @@
         xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
         xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
         xmlns:controls="clr-namespace:CefSharp.Wpf.Example.Controls"
+        xmlns:cefSharp="clr-namespace:CefSharp.Wpf;assembly=CefSharp.Wpf" xmlns:local="clr-namespace:CefSharp.Wpf.Example.ViewModels"
```
amaitland: Wrap lines.

CefSharp.Core/ManagedCefBrowserAdapter.cpp
```diff
@@ -20,6 +20,7 @@ void ManagedCefBrowserAdapter::CreateOffscreenBrowser(IntPtr windowHandle, Brows
     }
 }
 
+
```
amaitland: Remove extra line

CefSharp.Core/Internals/ClientAdapter.h
```diff
@@ -62,6 +62,8 @@ namespace CefSharp
             void CloseDevTools();
             void CloseAllPopups(bool forceClose);
 
+			virtual bool IsOffscreen() { return false; }
```
amaitland: Should be a property.

CefSharp.Core/Internals/ClientAdapter.cpp
```diff
@@ -72,19 +72,42 @@ namespace CefSharp
...
+
+			return returnVal;
+
         }
 
         void ClientAdapter::OnAfterCreated(CefRefPtr<CefBrowser> browser)
-        {
-            if (browser->IsPopup())
+        {			
+            if (browser->IsPopup() && !IsOffscreen())
```
amaitland: Without running the code the appearance of `IsOffscreen()` seems rather odd. Which `ChromiumWebBrowser` instances are the events been propagated to? The parent?

CefSharp.Core/Internals/ClientAdapter.cpp
```diff
@@ -72,19 +72,42 @@ namespace CefSharp
...
             {
                 return false;
             }
 
-            return handler->OnBeforePopup(_browserControl, gcnew CefFrameWrapper(frame, _browserControl->BrowserAdapter), StringUtils::ToClr(target_url),
-                windowInfo.x, windowInfo.y, windowInfo.width, windowInfo.height, *no_javascript_access);
+            bool returnVal = handler->OnBeforePopup(_browserControl, gcnew CefFrameWrapper(frame, _browserControl->BrowserAdapter), StringUtils::ToClr(target_url),
+				windowInfo.x, windowInfo.y, windowInfo.width, windowInfo.height, *no_javascript_access, newBrowser);
+
+			IWebBrowserInternal^ newBrowserInternal = dynamic_cast<IWebBrowserInternal^>(newBrowser);
```
amaitland: Without looking at this too closely, it appears you've cast more times than is required. What's stopping you from directly casting to `IRenderWebBrowser`?

Should probably check for `newBrowser` to be null before attempting anything. Also code needs comments to document the behavior.

You may not even need to cast as it doesn't appear your using the result.

amaitland:
Does this work in `WinForms`?

CefSharp.Wpf.Example/Views/BrowserTabView.xaml
jamespearce2006: I completely agree - it's not nice. But I could find no other way of delay-loading the control instance, and conform the the MVVM pattern to some degree. I'm not really happy with how the ChromiumWebBrowser gets created first, then added to the visual tree later - it seems the underlying control should probably manage that.

CefSharp.Wpf/ChromiumWebBrowser.cs
jamespearce2006: It seems passing null would crash LoadURL()

CefSharp.Core/Internals/ClientAdapter.cpp
jamespearce2006: The events are raised against the created instance (the popup). As it's now being displayed as windowless, we can't treat it as a popup.

CefSharp.Core/Internals/ClientAdapter.cpp
jamespearce2006: Probably - it's been a long while since I've worked in C++, and managed C++ never.

jamespearce2006:
It doesn't not work in Winforms...
I've not broken the Winforms example, but I've not specifically tested the offscreen popup feature... 
I'm selfishly concentrating my efforts on WPF at the moment.

Like I said though, I really just wanted this to be a starting point...

CefSharp.Wpf/ChromiumWebBrowser.cs
amaitland: Fair enough then :+1:

amaitland:
> Like I said though, I really just wanted this to be a starting point...

And that's fine. It will need to support `WinForms` before merge.

I think to keep the `Example` changes to a minimum, it's probably easiest to spawn a new `Window` that hosts the new `ChromiumWebBrowser` instance. Enough to demo the feature whilst keeping the `Example` simple.

CefSharp.Core/Internals/ClientAdapter.cpp
amaitland: Makes sense, add comments where it seems logical.

CefSharp.Core/Internals/ClientAdapter.cpp
amaitland: Check for `nullptr` and use the `IsOffScreen` check where appropriate, most likely remove two casts.

CefSharp.Core/Internals/ClientAdapter.cpp
```diff
@@ -72,19 +72,42 @@ namespace CefSharp
...
+            IWebBrowser^ newBrowser = nullptr;
 
             if (handler == nullptr)
             {
                 return false;
             }
 
-            return handler->OnBeforePopup(_browserControl, gcnew CefFrameWrapper(frame, _browserControl->BrowserAdapter), StringUtils::ToClr(target_url),
-                windowInfo.x, windowInfo.y, windowInfo.width, windowInfo.height, *no_javascript_access);
+            bool returnVal = handler->OnBeforePopup(_browserControl, gcnew CefFrameWrapper(frame, _browserControl->BrowserAdapter), StringUtils::ToClr(target_url),
```
amaitland: if `returnValue` is true then then opening the popup will be cancelled, so only attempt to assign browser if `false`

rassilon:
Oh hrm.... @amaitland , this interacts in a subtle way with the move to expose IBrowser and IFrame as well as the changes we discussed about adding the IPopupHandler interface.

Would it be too weird for the offscreen popup browser control (i.e. `IWebBrowser`) to only call the `IPopupHandler` methods?

Thoughts?

CefSharp.Core/Internals/ClientAdapter.h
```diff
@@ -62,6 +62,8 @@ namespace CefSharp
             void CloseDevTools();
             void CloseAllPopups(bool forceClose);
 
+            virtual bool IsOffscreen() { return false; }
```
rassilon: This name is  slightly misleading. Shouldn't it rather be something like `IsPopupInSeparateWebBrowserControl` or something? The parameter documentation for `newBrowser` just says its in a different control, `newBrowser` doesn't need to necessarily be an Offscreen control.

rassilon:
@jamespearce2006 , thank you very much for the launching off point for this issue by creating this starting point for everybody.

amaitland:
> Oh hrm.... @amaitland , this interacts in a subtle way with the move to expose IBrowser and IFrame as well as the changes we discussed about adding the IPopupHandler interface.

Unfortunately :frowning: I was hopeful that we might be able to tell `CEF` that it's not actually a popup, though haven't looked into that yet.

> Would it be too weird for the offscreen popup browser control (i.e. IWebBrowser) to only call the IPopupHandler methods?

Can you elaborate a little more?

rassilon:
> > Would it be too weird for the offscreen popup browser control (i.e. IWebBrowser) to only call the IPopupHandler methods?

> Can you elaborate a little more?

Well, since it's a popup window, IWebBorwser wouldn't call the event handlers that only apply to  MainFrame browsers.

amaitland:
> Well, since it's a popup window, IWebBorwser wouldn't call the event handlers that only apply to MainFrame browsers

So the `Dependency Properties` wouldn't work anymore?

amaitland:
(From memory there are a few bound to the `CEF` events in question).

amaitland:
On a side note, we should look at exposing `CefPopupFeatures` and `CefBrowserSettings` when opening a popup.

rassilon:
> So the Dependency Properties wouldn't work anymore?

Not unless we create a special check for this case. (Which is what I think this PR's IsOffscreen`method is doing.

amaitland:
I think the combination of `IsPopup` and `IsOffScreen` combined should be enough to differentiate.

amaitland:
Probably need to conditionally call `CefBrowserHost::CreateBrowser` now, doubt it will be needed for this popup scenario.

CefSharp.Core/Internals/ClientAdapter.cpp
```diff
@@ -72,19 +72,42 @@ namespace CefSharp
...
+                windowInfo.x, windowInfo.y, windowInfo.width, windowInfo.height, *no_javascript_access, newBrowser);
+
+            IWebBrowserInternal^ newBrowserInternal = dynamic_cast<IWebBrowserInternal^>(newBrowser);
```
amaitland: Should really check `newBrowser != null` first up, then only proceed of that's the cast.

amaitland:
@rassilon I'm somewhat inclined to resolve the `IPopupHandler/whatever we're going to call it` changes first then look at merging this once there's a clearer picture of how that will work. Thoughts?

CefSharp.Wpf.Example/Views/BrowserTabView.xaml
amaitland: > it seems the underlying control should probably manage that.

Which control were you thinking? I'm sure we can come up with something, just have to ponder for a bit....

jamespearce2006:
I've simplified the XAML example, as requested, and made it simply open the popup in a new window, rather than trying to shoe-horn into the existing MVVM pattern.

jamespearce2006:
I had a look at simplifying the casts in ClientAdapter::OnBeforePopup, but I'm not sure I can. I can't just check the IsOffscreen property because we're dealing with a newly created client, not the current instance. 

The first cast is to get the IWebBrowserInternal, to ensure it's something that belongs to us. This will work whether it's offscreen or not.
The next cast is to do some specific work for IRenderWebBrowser (set windowInfo set windowless).

It's quite likely I'm misunderstanding something, though.

amaitland:
> I've simplified the XAML example, as requested, and made it simply open the popup in a new window, rather than trying to shoe-horn into the existing MVVM pattern.

Nice one :+1: 

> I had a look at simplifying the casts in ClientAdapter::OnBeforePopup, but I'm not sure I can. I can't just check the IsOffscreen property because we're dealing with a newly created client, not the current instance.

If the current instance is `OffScreen` then the newly created instance should also be `OffScreen`? We will need to add a separate property as @rassilon suggested as the `IsOffScreen()` checks you added won't work for `WinForms`. Rather than littering the place with `if else` blocks we might need to consider another layer of abstraction, possibly a plug-able implementation to handle different scenarios, some more though is required I think (It maybe overkill).

rassilon:
Yeah, I agree we should merge this after the IPopupHandler changes.

jamespearce2006:
> If the current instance is OffScreen then the newly created instance should also be OffScreen?

I wouldn't have thought that's guaranteed. Presumably there's nothing to stop a windowed instance creating an offscreen popup.

arsher:
Hey guys,

I've checked this out because I'm going to need something like this soon. But I found a caveat with the current approach.
When you're using window.open and staying in the same domain chromium can put the child window into the same process and also gives you access to the parent window (and from the parent to the children) through window.opener so the js context is accessible and you can also pass dom elements. But this solution has a new CefClient for each control so this pretty neat feature won't work. I did some experiments with reusing the ClientAdapter and creating a CefRenderHandler for each child window and it seems to be working, but it would require some thinking on the architecture to fit this in with proper cleanup etc...What do you think? Is it worth it to pursuing this?

David

amaitland:
#1150 was merged into `master`, so closing this.

PR #1017, opened 2015-05-08
Honour the Background property for the ChromiumWebBrowser

jamespearce2006:
Currently, if you set the Background DependencyProperty on the ChromiumWebBrowser control, it is ignored.

This pull request fixes that.


jamespearce2006:
Never mind, I think it would be better to allow the client app to disable the transparent offscreen rendering...

Issue #1007, opened 2015-05-06
More XP problems


hvdijk:
Hello,

The CefSharp 39 packages don't work on Windows XP. I'm aware of the FAQ entry, but the problems here aren't covered there.

Most importantly, it just plain bombs. Dependency Walker shows that the problem is that CefSharp.Core.dll refers to `GetTickCount64`, which is new in Windows Vista. This dependency does not come from the source code, and if I build CefSharp myself instead of using the pre-built packages, there is no reference to `GetTickCount64` anywhere. This may be related to one of the updates to Visual Studio.

I will be using the locally built packages for now to avoid this problem, but it would be nice if the official packages could be built without this dependency.

Additionally, a milder issue, the `DependencyChecker.IsWindowsXp` property doesn't work. Its value is only ever read in the `DependencyChecker.CefOptionalDependencies` initialiser, but the initialiser will already have run by the time any code gets the chance to set the property value. It seems like it would be easy to make `DependencyChecker.IsWindowsXp` a read-only property: it can simply check `Environment.OSVersion.Version.Major < 6` (optionally in addition to a sanity check -- `Environment.OSVersion.Platform == PlatformID.Win32NT && Environment.OSVersion.Version.Major < 6` -- but clearly the other platforms aren't going to work).

I clean up `DependencyChecker.CefOptionalDependencies` manually now to work around that problem.

With those changes, it's working nicely for me on XP SP3, .NET Framework 4 and Visual C++ redistributables installed.

amaitland:
> With those changes, it's working nicely for me on XP SP3, .NET Framework 4 and Visual C++ redistributables installed.

Great :+1:

hvdijk:
Sorry, I don't understand why you closed this? If your position is that as you no longer support XP, you  are not willing to make changes to get it working on XP again, that is an unfortunate but understandable position. However, in that case, the FAQ entry is just plain wrong. "We no longer officially support Windows XP. That being said CefSharp should work in most cases" -- The current releases of CefSharp shouldn't work on any Windows XP machine for anyone. Perhaps the entry should then say something like "We no longer support Windows XP, and our CefSharp releases will not run on it. If you need to use CefSharp on Windows XP, compiling your own version from source may allow it to work."?

amaitland:
> Sorry, I don't understand why you closed this?

Your problem was solved and at no point did you ask a question. 

[...]

PR #1018, opened 2015-05-08
Automatically detect Windows XP.

hvdijk:
As requested in https://github.com/cefsharp/CefSharp/issues/1007, here is a pull request containing only the change to automatically detect Windows XP:

Previously, it was the main program's responsibility to set `IsWindowsXp` if necessary, but that was problematic because it wasn't possible to set it before its value was used. But it's easy enough to detect automatically anyway, so it's probably nicer to just do that.

### DIFF
--- a/CefSharp.Example/CefExample.cs
+++ b/CefSharp.Example/CefExample.cs
@@ -79,10 +79,6 @@ namespace CefSharp.Example
                 Cef.SetCookiePath("cookies", true);
             };
 
-            //Cef will check if all dependencies are present
-            //For special case when Checking Windows Xp Dependencies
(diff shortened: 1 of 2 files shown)


rassilon:
Looks great. Thanks!

Bill

amaitland:
@rassilon Just a reminder to update the changelog. Thanks! :smile:

rassilon:
Heh. Unless we produce a build that works on xp, doesn't seem worth the effort. (low as it is) :)

amaitland:
Hmm, still relevant for people who choose to build from source.

amaitland:
I've amended the title slight and added to the `changelog`.

PR #1019, opened 2015-05-08
Allow app to disable transparent background in windowless mode

jamespearce2006:
Currently, ManagedCefBrowserAdapter always sets transparent to TRUE when setting windowless mode.

Transparency isn't always desired, and attempting to place the browser control in a painted outer control can have ugly side-effects when resizing the app, if its colour is different from the containing window.

This PR lets the ChromiumWebBrowser specify whether transparency should be set.

### DIFF
--- a/CefSharp.Core/ManagedCefBrowserAdapter.h
+++ b/CefSharp.Core/ManagedCefBrowserAdapter.h
@@ -72,14 +72,10 @@ namespace CefSharp
 
         void CreateOffscreenBrowser(IntPtr windowHandle, BrowserSettings^ browserSettings, String^ address)
         {
-            CreateOffscreenBrowser(windowHandle, browserSettings, address, FALSE);
-        }
-
-        void CreateOffscreenBrowser(IntPtr windowHandle, BrowserSettings^ browserSettings, String^ address, bool transparent)
-        {
             auto hwnd = static_cast<HWND>(windowHandle.ToPointer());
 
             CefWindowInfo window;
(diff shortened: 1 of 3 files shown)


rassilon:
Does it make sense to make the same change to the Offscreen control as well?

CefSharp.Core/ManagedCefBrowserAdapter.h
```diff
@@ -72,10 +72,15 @@ namespace CefSharp
 
         void CreateOffscreenBrowser(IntPtr windowHandle, BrowserSettings^ browserSettings, String^ address)
```
amaitland: No need for an overload as this is an internal method.

amaitland:
> Does it make sense to make the same change to the Offscreen control as well?

Yes

This topic was discussed a while ago. I think the general idea was to add the property to the `BrowserSettings`. Thoughts?

jamespearce2006:
I've made the suggested change - to use BrowserSettings instead. It seems a lot cleaner.

amaitland:
> I've made the suggested change - to use BrowserSettings instead. It seems a lot cleaner

Cool :smile: 

I made a few changes, changed to using `GetValueOrDefault` to simplify the code. Uppercased the `S` in `OffScreen` to match other project references. Changed to using an auto property in `BrowserSettings`

https://github.com/cefsharp/CefSharp/commit/3b27034b3539e6d209c421dc606f54389a882e6d https://github.com/cefsharp/CefSharp/commit/6805c188e7eb655e8cc66f8ae758e400a74efedd

Thanks for the contribution :+1:

PR #1030, opened 2015-05-17
More CefSharp BrowserWrapper work

rassilon:
Here's some more work on wrapping CefBrowser/CefFrame.

Added IPopupHandler and some support for calling it.

I'll write up what look like breaking changes in a subsequent comment for special attention.

Bill


amaitland:
Forget to add` IPopupHandler.cs`?

rassilon:
Bah. I sure did. I've pushed it up now.

amaitland:
Any thoughts on `OnLoadStart/OnLoadEnd`? We currently ignore all popup notifications, if we change to call the event handlers in all circumstances it will be a subtle change in behavior. Do we push those notifications to `IPopupHandler`?

rassilon:
I thought about them, but weren't those the ones that weren't reliable?

CefSharp.Core/Internals/ClientAdapter.cpp
```diff
@@ -107,14 +110,18 @@ namespace CefSharp
...
+                    IPopupHandler^ handler = _browserControl->PopupHandler;
+                    if (handler != nullptr)
                     {
-                        _popupBrowsers.erase(it);
-                        break;
+                        handler->OnBeforeClose(_browserControl, entry);
                     }
+                    _popupBrowsers->Remove(browserId);
+                    // Dispose the CefSharpBrowserWrapper
+                    (dynamic_cast<CefSharpBrowserWrapper^>(entry))->~CefSharpBrowserWrapper();
```
amaitland: Is it possible to just call `delete entry`?

CefSharp.Core/Internals/ClientAdapter.cpp
```diff
@@ -131,7 +138,24 @@ namespace CefSharp
...
         void ClientAdapter::OnLoadingStateChange(CefRefPtr<CefBrowser> browser, bool isLoading, bool canGoBack, bool canGoForward)
         {
-            _browserControl->SetLoadingStateChange(canGoBack, canGoForward, isLoading);
+            if (browser->IsPopup())
+            {
+                auto browserId = browser->GetIdentifier();
+                IBrowser^ entry;
+                IPopupHandler^ handler = _browserControl->PopupHandler;
+
+                if (_popupBrowsers->TryGetValue(browserId, entry))
```
amaitland: Check `handler` for `null` first? Slightly more efficient when there's no `handler` assigned.

CefSharp.Core/Internals/ClientAdapter.cpp
```diff
@@ -159,10 +183,25 @@ namespace CefSharp
         void ClientAdapter::OnFaviconURLChange(CefRefPtr<CefBrowser> browser, const std::vector<CefString>& iconUrls)
         {
             auto handler = _browserControl->RequestHandler;
-
-            if(handler != nullptr)
+            auto popupHandler = _browserControl->PopupHandler;
```
amaitland: Move both `handler` assignments into their relevant `if` branches

CefSharp.Core/Internals/ClientAdapter.cpp
rassilon: I'll give it a shot. Managed c++ takes some getting used to.

CefSharp.Core/Internals/ClientAdapter.cpp
rassilon: No problem.

CefSharp.Core/ManagedCefBrowserAdapter.h
```diff
@@ -76,12 +76,12 @@ namespace CefSharp
             if (offScreenRendering)
             {
                 _clientAdapter = new RenderClientAdapter(webBrowserInternal,
-                    gcnew Action<int>(this, &ManagedCefBrowserAdapter::OnAfterBrowserCreated));
+                    gcnew Action<int>(this, &ManagedCefBrowserAdapter::OnAfterBrowserCreated), this);
```
amaitland: We should ditch the `Action` and add a method to the interface now that `ClientAdapter` has a reference to `ManagedCefBrowserAdapter`.

amaitland:
> I thought about them, but weren't those the ones that weren't reliable?

Yep, they're somewhat unreliable as an overall marker of when the page has started/stopped loading. Unfortunately `OnLoadingStateChange` doesn't pass in a `CefFrame` reference, so if you really wanted to know info for a specific frame, then `OnLoadStart/End` are the only relevant methods. 

I'm happy to leave them as is just though it was worth asking the question.

amaitland:
Was thinking when your happy with these changes we might look at pushing out the first `-pre` for `41`.

rassilon:
Shouldn't we fold in the osr changes that creates osr controls for pop-ups, in order to get rid of most of these breaking changes at once?

amaitland:
> Shouldn't we fold in the osr changes that creates osr controls for pop-ups, in order to get rid of most of these breaking changes at once?

Possibly, I'd like to see the `WinForms` implementation working before merge. There will also be a bit of rework required to support the `IPopupHandler` changes. As it's only one additional param on the public facing `API` I wasn't too worried about pushing it to the second `-pre`. I was also hoping to find time to dig through the `CEF` code, see if there's a cleaner/nicer way to spawn a new instance, it just seems messy.

rassilon:
> Yep, they're somewhat unreliable as an overall marker of when the page has started/stopped loading. Unfortunately  OnLoadingStateChange  doesn't pass in a  CefFrame  reference, so if you really wanted to know info for a specific frame, then  OnLoadStart/End  are the only relevant methods. 

Yeah, I noticed that as well. *sigh* I have no problem putting them in, I just didn't do it yet due them being unreliable.

I hope to put in some time tonight to address your feedback.

Bill

amaitland:
> I hope to put in some time tonight to address your feedback.

Let me know if there is anything I can do to help. If you want to merge this as is I'm happy to make the follow up changes.

rassilon:
If this build succeeds, we can merge it in, and then we can look at what remains.
Some of the functions listed in the issue are recipes for back compatibility nightmares if we call IPopupHandler on them.

amaitland:
> Some of the functions listed in the issue are recipes for back compatibility nightmares if we only call the IPopupHandler method.

Might have to head off some of the issues with a fairly detailed `wiki` entry or something similar.....

rassilon:
Just call it CefSharp 4. :smile:

rassilon:
> delete work as expected?

Tried a popup window test, it called the managed C++ destructor correctly.

amaitland:
> Just call it CefSharp 4. :smile:

Haha, maybe we should :smile:

Issue #1031, opened 2015-05-18
Question: Blank Popups when downloading


jamespearce2006:
When clicking on a link to download a file, some websites open a popup to service that request. The popup ends up being blank.

Chrome will automatically close these popup windows once the download is complete. How might I do this in CefSharp? Obviously, I know from the download handler when the download is done, but I'm not sure how to reconcile the download with the window I need to close.
I considered just returning true from the OnBeforePopup call, but this prevents the download occurring altogether.

PR #1033, opened 2015-05-18
Auto-close blank popups when download is complete

jamespearce2006:
Fix for issue #1031

Automatically close blank popups that appear when downloads are started in a separate window, once the download is done (complete, failed or cancelled).

Test the fix here:
https://jsfiddle.net/yLk7d3z8/

### DIFF
--- a/CefSharp.Core/Internals/DownloadAdapter.cpp
+++ b/CefSharp.Core/Internals/DownloadAdapter.cpp
@@ -38,6 +38,11 @@ namespace CefSharp
         void DownloadAdapter::OnDownloadUpdated(CefRefPtr<CefBrowser> browser, CefRefPtr<CefDownloadItem> download_item,
             CefRefPtr<CefDownloadItemCallback> callback) 
         {
+            if (!download_item->IsInProgress() && browser->IsPopup() && !browser->HasDocument())
+            {
+                browser->GetHost()->CloseBrowser(false);
+            }
+
             if (_handler->OnDownloadUpdated(TypeConversion::FromNative(download_item)))
             {
                 callback->Cancel();
--- a/CefSharp.Core/Internals/DownloadAdapter.cpp
+++ b/CefSharp.Core/Internals/DownloadAdapter.cpp
@@ -38,6 +38,7 @@ namespace CefSharp
         void DownloadAdapter::OnDownloadUpdated(CefRefPtr<CefBrowser> browser, CefRefPtr<CefDownloadItem> download_item,
             CefRefPtr<CefDownloadItemCallback> callback) 
         {
+
             if (!download_item->IsInProgress() && browser->IsPopup() && !browser->HasDocument())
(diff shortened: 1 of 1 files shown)


CefSharp.Core/Internals/DownloadAdapter.cpp
```diff
@@ -38,6 +38,12 @@ namespace CefSharp
         void DownloadAdapter::OnDownloadUpdated(CefRefPtr<CefBrowser> browser, CefRefPtr<CefDownloadItem> download_item,
             CefRefPtr<CefDownloadItemCallback> callback) 
         {
+
```
jankurianski: Remove extra blank line.

amaitland:
If we merge this into `master`, I'd probably remove it in `2272` and give the user control to close their own browsers since we've introduced a wrapper. (I don't think we expose the close method yet, we should though).

jankurianski:
Ah, you're talking about the working being done in #1030? So the `IDownloadHandler` method could turn into:

    bool OnDownloadUpdated(CefSharpBrowserWrapper browser, DownloadItem downloadItem)

And then users can do the `browser.Close()` if they chose?  I think :+1:

amaitland:
> And then users can do the browser.Close() if they chose?

Yep, that's the one.

CefSharp.Core/Internals/DownloadAdapter.cpp
jamespearce2006: Done - I added the line to force AppVeyor to rebuild, but forgot to remove it.

jamespearce2006:
@amaitland I agree, 2272 gives the user more control over such things. Do you think it'd be worth keeping this behavior as default, though, if the user doesn't implement the appropriate handlers?

amaitland:
> Do you think it'd be worth keeping this behavior as default, though, if the user doesn't implement the appropriate handlers?

For downloads you have to implement `IDownloadHandler` or they're silently ignored. The default makes sense, I just think it's a little difficult to implement conditionally.

amaitland:
I have toyed with the idea of implementing a generic set of handlers that provide some sensible defaults, make all the methods `virtual` allowing the user to override only the pieces they require. Best of both worlds in some ways, if you require the flexibility you can implement all the low lying stuff yourself, if you only need to implement one method on `IRequestHandler` then it safes a lot of effort (specially in the fact that you have to `return false` to implement the default behavior in a lot of cases). Just one of those low priority ideas that I'd never gotten around too.

fabionardelli1982:
I checked the issue with the version 89.0.17 and the blank popup is not closed at the end of the Download of file. Then the issue is still reproducible

amaitland:
To give users full control this change was reverted as there was no way to disable the default behaviour. 

The relevant APIs for closing the popup were added years ago so you can easily close the popup in your IDownloadHandler implementation.

Issue #1037, opened 2015-05-20
ChromiumWebBrowser Code Duplication


amaitland:
Look ways to reduce the duplication between the three flavors.

- Can possibly use extension methods to handle repeating code blocks internally, like clearing handlers and events that are common to `IWebBrowser`
- Make `Cut/Copy/Paste/etc` as obsolete so they can be removed in `43`

If anyone else has thoughts or suggestions I'd be interested to hear them.

rassilon:
Close this one?

Issue #1016, opened 2015-05-08
Cleanup RenderClientAdapter/ClientAdapter


amaitland:
Things are getting a little bit messy, with all the intertwined layers, need to review and see if a cleaner separation of concerns is possible.

Now that `ClientAdapter` has access to `IBrowserAdapter` we can remove the `onAfterBrowserCreated` action, possibly cleanup the other layers as well.

PR #1041, opened 2015-05-21
Continued 2272 Code Cleanup

amaitland:
Some Resharper inspired code cleanups. Couple of code simplifications.

Start to look at move common blocks of code into extension methods where possible/sensible.


amaitland:
Start of #1037 and #1016

amaitland:
Since we are going for an overhaul of `CefSharp`, I'm wondering if we might do some other major changes (Maybe we will need to call it `CefSharp 4` as @rassilon  suggested :wink:)

- Remove shortcut methods like `GetSource`, `GetText` that call the underlying frame wrapper from `IWebBrowser`
- Add `IWebBrowser` extension methods, that implement the same behavior and can be used across the three `flavours`.

For `GetSourceAsync` for instance would move to being an extension method that got the `IFrame` instance for `MainFrame` and did the same thing.

Would be a small amount of effort, and a reusable pattern, would greatly reduce the amount of duplication. Long as the extensions were in the `CefSharp` namespace, you really shouldn't know much has changed.

I'll probably implement a few methods to see what it looks like and the practicality.

amaitland:
Still outstanding

- [ ] Finish `ResourceHandler` async
- [ ] Expose more `CefBrowserHost` methods (For now I think it's fine to access them indirectly through `IBrowser`)
- [ ] Add `IBrowser` to `IRequestHandler` methods

amaitland:
Whilst this builds the examples crash shortly after running.

amaitland:
Basics should be working now.

- Implemented async ResourceHandler (uses the same implementation as `Scheme Handler` for now, can look at using `StreamAdapter` and enhancing that later
- Move more methods into Extensions, the underlying `API` is a lot more like `CEF` now, so we just provide some convenience extension methods to make things easier.
- Might reconsider adding a wrapper for `CefBrowserHost`, would mean more code could be made into extension methods.

rassilon:
Had a look see: (this is some awesome stuff btw)

* Print should probably be an extension method.
* Ditto your `StreamAdapter` comment, at minimum we should at least not spawn a task if it's a `MemoryStream`.
* Should `DefaultResourceHandlerFactory` implement `IRegisterResourceHandlerFactory` or something so that Cef.Example doesn't need an as `DefaultResourceHandlerFactory` cast?

> Might reconsider adding a wrapper for `CefBrowserHost`, ...

That would mostly be for popup related reasons? (offscreen/wpf/odd winforms hosting ui/etc?)

Up for merging in what you have so far?

Bill

rassilon:
Testing this stuff really is a pain, maybe we should see about creating more of cefclient's built in test resources into Cef.Example or something...

amaitland:
> Print should probably be an extension method.

Done :+1: 

> Ditto your StreamAdapter comment, at minimum we should at least not spawn a task if it's a MemoryStream

Sounds reasonable.

> Should DefaultResourceHandlerFactory implement IRegisterResourceHandlerFactory or something so that Cef.Example doesn't need an as DefaultResourceHandlerFactory cast?

Yeah, we should look to improve that. `LoadHtml` would also benefit from something like that.

> That would mostly be for popup related reasons? (offscreen/wpf/odd winforms hosting ui/etc?)

Meant I could move `Print`, `AddWordToDictionary`, `ReplaceMisspelling`, `CloseDevTools`, `ShowDevTools`, `StopFinding`, `Find` into extension methods. Hadn't considered popups, though now that you mention it, would probably be helpful there too.

> Up for merging in what you have so far?

I probably haven't tested things well enough, as you already pointed out it's a little difficult to cover it all. I'm happy to merge, though there are likely a few rough edges.

> Testing this stuff really is a pain, maybe we should see about creating more of cefclient's built in test resources into Cef.Example or something...

Sounds like a plan.

amaitland:
> Do you think we should have IBrowser extension methods for some of these?

Which methods were you thinking of?

rassilon:
> Which methods were you thinking of?

I was thinking most of `IBrowserHost` for starters. From a users perspective, it's weird they're on the `BrowserHost` object.

Of course the need to care could disappear if we're creating IWebBrowser instances for each created popup.

Bill

amaitland:
> I was thinking most of IBrowserHost for starters. From a users perspective, it's weird they're on the BrowserHost object.

True, the `CefBrowserHost` is a little strange from a users point of view, so we could easily create some extensions.

amaitland:
So your in favor of merging this now?

rassilon:
Yep. Looking good.

Bill

amaitland:
I see you just pushed a new commit, so I'll let you merge whenever your finished :+1:

rassilon:
Heh. I think that's it for now (assuming the build passes). For the `StreamAdapter.MayBlock` concept, have any thoughts?  

Bill

amaitland:
> For the StreamAdapter.MayBlock concept, have any thoughts?

Looking at the implementation of `CefStreamResourceHandler`, do we need an actual `StreamAdapter` or can we modify the existing `ResourceHandlerWrapper` to post the reads onto the `File` thread? Assuming we reuse the `CefStreamResourceHandler::Buffer` implementation that is.

rassilon:
We don't need an actual StreamAdapter persay... just consider whether we care about providing the `MayBlock` like behavior to our users. So we could modify `ResourceHandlerWrapper` appropriately with an extra `MayBlock` property to do whatever other thing we want. If it `MayBlock` we should probably be paranoid and let it run on the .Net thread pool.

amaitland:
> If it MayBlock we should probably be paranoid and let it run on the .Net thread pool.

You wouldn't let it run on `TID_FILE`?

rassilon:
It's an arbitrary .Net `Stream`, I'm paranoid. :smile:

amaitland:
> It's an arbitrary .Net Stream, I'm paranoid. :smile:

That just complicates things then :stuck_out_tongue_winking_eye: Are you happy to kick that off when you get a chance? I'm not quite sure where to start.

rassilon:
I can give it a shot.

amaitland:
https://bitbucket.org/chromiumembedded/cef/issue/1640/add-new-cefresourcemanager-class-for maybe one to keep an eye on for future reference.

PR #1050, opened 2015-05-29
Add some more methods on IPopupHandler and call them.

rassilon:
Don't put the following handlers on IPopupHandler due to the underlying security related purpose, but add IBrowser parameters:

* OnCertificateError
* OnQuotaError
* OnPluginCrashed
* OnRenderProcessTerminated
* OnProtocolExecution
* OnBeforePluginLoad
* GetAuthCredentials

### DIFF
--- a/CefSharp.Core/Internals/CefFrameWrapper.cpp
+++ b/CefSharp.Core/Internals/CefFrameWrapper.cpp
@@ -234,10 +234,24 @@ Int64 CefFrameWrapper::Identifier::get()
 IFrame^ CefFrameWrapper::Parent::get()
 {
     ThrowIfDisposed();
-    auto parent = _frame->GetParent();
-    if (parent != nullptr)
+    if (_parentFrame != nullptr)
     {
-        return gcnew CefFrameWrapper(parent, _browserAdapter);
+        // Be paranoid about creating the cached IBrowser.
+        lock sync(_syncRoot);
+
+        auto parent = _frame->GetParent();
+        if (parent != nullptr && _parentFrame != nullptr)
+        {
+            _parentFrame = gcnew CefFrameWrapper(parent, _browserAdapter);
+        }
+        else if (parent == nullptr)
+        {
+            return nullptr;
+        }
+    }
+    else
+    {
+        return _parentFrame;
     }
     return nullptr;
 }
@@ -259,7 +273,21 @@ String^ CefFrameWrapper::Url::get()
 IBrowser^ CefFrameWrapper::Browser::get()
 {
     ThrowIfDisposed();
-    return dynamic_cast<IBrowser^>(gcnew CefSharpBrowserWrapper(_frame->GetBrowser(), _browserAdapter));
+    if (_owningBrowser != nullptr)
+    {
+        return _owningBrowser;
+    }
+
+    // Be paranoid about creating the cached IBrowser.
+    lock sync(_syncRoot);
+
+    if (_owningBrowser != nullptr)
+    {
+        return _owningBrowser;
+    }
+
+    _owningBrowser = gcnew CefSharpBrowserWrapper(_frame->GetBrowser(), _browserAdapter);
+    return _owningBrowser;
 }
 
 void CefFrameWrapper::ThrowIfDisposed()
--- a/CefSharp.Core/Internals/ClientAdapter.cpp
+++ b/CefSharp.Core/Internals/ClientAdapter.cpp
@@ -38,6 +38,7 @@ namespace CefSharp
             CefRefPtr<CefClient>& client, CefBrowserSettings& settings, bool* no_javascript_access)
         {
             ILifeSpanHandler^ handler = _browserControl->LifeSpanHandler;
+            bool createdWrapper = false;
 
             if (handler == nullptr)
             {
@@ -47,10 +48,16 @@ namespace CefSharp
             if (!_popupBrowsers->TryGetValue(browser->GetIdentifier(), browserWrapper))
             {
                 browserWrapper = gcnew CefSharpBrowserWrapper(browser, _browserAdapter);
+                createdWrapper = true;
             }
+            CefFrameWrapper frameWrapper(frame, _browserAdapter);
             auto result = handler->OnBeforePopup(_browserControl, browserWrapper,
-                gcnew CefFrameWrapper(frame, _browserAdapter), StringUtils::ToClr(target_url),
+                %frameWrapper, StringUtils::ToClr(target_url),
                 windowInfo.x, windowInfo.y, windowInfo.width, windowInfo.height, *no_javascript_access);
+            if (createdWrapper)
+            {
+                delete browserWrapper;
+            }
             return result;
         }
 
@@ -98,6 +105,10 @@ namespace CefSharp
                     // Dispose the CefSharpBrowserWrapper
                     delete entry;
                 }
+                else
+                {
+                    ThrowUnknownPopupBrowser(gcnew String(L"ClientAdapter::OnBeforeClose"));
+                }
             }
             else if (_browserHwnd == browser->GetHost()->GetWindowHandle())
             {
@@ -123,6 +134,10 @@ namespace CefSharp
                     if (_popupBrowsers->TryGetValue(browserId, entry))
                     {
                         handler->OnLoadingStateChange(_browserControl, entry, isLoading, canGoBack, canGoForward);
+                    }
+                    else
+                    {
+                        ThrowUnknownPopupBrowser(gcnew String(L"ClientAdapter::OnLoadingStateChange"));
                     }
                 }
             }
@@ -167,6 +182,10 @@ namespace CefSharp
                     {
                         popupHandler->OnFaviconUrlChange(_browserControl, entry, StringUtils::ToClr(iconUrls));
                     }
+                    else
+                    {
+                        ThrowUnknownPopupBrowser(gcnew String(L"ClientAdapter::OnFaviconUrlChange"));
+                    }
                 }
             }
             else
@@ -183,6 +202,7 @@ namespace CefSharp
         {
             String^ tooltip = StringUtils::ToClr(text);
 
+            // TODO: Deal with popuup browsers properly...
             if (tooltip != _tooltip)
             {
                 _tooltip = tooltip;
@@ -215,6 +235,10 @@ namespace CefSharp
                     {
                         popupHandler->OnStatusMessage(_browserControl, entry, valueStr);
                     }
+                    else
+                    {
+                        ThrowUnknownPopupBrowser(gcnew String(L"ClientAdapter::OnStatusMessage"));
+                    }
                 }
             }
             else
@@ -225,13 +249,6 @@ namespace CefSharp
 
         bool ClientAdapter::OnKeyEvent(CefRefPtr<CefBrowser> browser, const CefKeyEvent& event, CefEventHandle os_event)
         {
-            IKeyboardHandler^ handler = _browserControl->KeyboardHandler;
-
-            if (handler == nullptr)
-            {
-                return false;
-            }
-
             // TODO: windows_key_code could possibly be the wrong choice here (the OnKeyEvent signature has changed since CEF1). The
             // other option would be native_key_code.
             if (browser->IsPopup())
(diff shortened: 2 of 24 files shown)


CefSharp.Core/Internals/ClientAdapter.cpp
```diff
@@ -323,15 +350,36 @@ namespace CefSharp
...
+                auto popupHandler = _browserControl->PopupHandler;
+                if (popupHandler != nullptr)
+                {
+                    auto browserId = browser->GetIdentifier();
+                    IBrowser^ entry;
+                    if (_popupBrowsers->TryGetValue(browserId, entry))
+                    {
+                        auto frameWrapper = gcnew CefFrameWrapper(frame, _browserAdapter);
+                        CefRequestWrapper^ wrapper = gcnew CefRequestWrapper(request);
+                        return popupHandler->OnBeforeBrowse(_browserControl, entry, wrapper, isRedirect, gcnew CefFrameWrapper(frame, _browserAdapter));
```
rassilon: What's the ownership contract here do you think? Should we own disposing this instance right after the callbacks return, or should we just cache them all thus allowing the handlers the least amount of pain and suffering? Thoughts, @amaitland ?

(I'm talking about the `frameWrapper` here that gets created in this chunk of code.)

amaitland: Is it practical to cache the `CefFrameWrapper` instance? What's the trigger to dispose of an instance? 

Obviously the simplest approach from our end is to dispose straight after callback. How limiting would that be? We've got the same problem with `GetFocusedFrame` and others. Then we have even less control. 

Does `OnLoadingStateChange` (or similar) give us a clear boundary to dispose of old instances?

CefSharp.Core/Internals/ClientAdapter.cpp
```diff
@@ -232,10 +232,25 @@ namespace CefSharp
...
+            if (browser->IsPopup())
+            {
+                auto popupHandler = _browserControl->PopupHandler;
+                if (popupHandler != nullptr)
+                {
+                    auto browserId = browser->GetIdentifier();
+                    IBrowser^ entry;
+                    if (_popupBrowsers->TryGetValue(browserId, entry))
+                    {
+                        popupHandler->OnKeyEvent(_browserControl, entry, (KeyType)event.type, event.windows_key_code, (CefEventFlags)event.modifiers, event.is_system_key == 1);
```
amaitland: Missing a `return`?

CefSharp.Core/Internals/ClientAdapter.cpp
amaitland: I've added more wrappers around some of the `async` callbacks, in a lot of ways they have the same problem, so ideally we have a strategy that's consistent for those.

rassilon: The overhead of `CefFrameWrapper` is obviously very small.  The easiest thing for us to do is to dispose of the wrappers after we finish the callback. 

Do we get notifications when frames/browsers get destroyed? If we did, we could do something more complicated:

* Have an internal .Net friendly `CefFrameWrapper` that we cache, and is fronted by a weak reference owning `ClientCefFrameWrapper` like object.
i.e. `ClientCefFrameWrapepr` would contain a GC weak reference to our `CefFrameWrapper` instance. 

This way, we can own the lifetimes of the important CEF ref counts via our `CefFrameWrapper`, and if our users hang around to the `ClientCefFrameWrapper`s too long, then they'll get the appropriate exception afterwards when the weak reference is no longer valid.

It is even spookier action at a distance, but at least it lets us think more clearly about how we're managing CEF object lifetimes. (which clearly has given us trouble in the past, i.e. shutdowns)

Thoughts?

I'm not sure `OnLoadingStateChange` is the correct context for disposing of the instances. (That would surprise me, since we don't cache any information in our .Net wrappers, we always ask CEF what the current values are.)

Bill

amaitland: > Do we get notifications when frames/browsers get destroyed?

For `CefBrowser` it's pretty clear and I think that's pretty straight forward to reuse the wrapper. For `CefFrame`, I'm not sure that there is such a notification.

> Have an internal .Net friendly CefFrameWrapper that we cache, and is fronted by a weak reference owning ClientCefFrameWrapper like object. i.e. ClientCefFrameWrapepr would contain a GC weak reference to our CefFrameWrapper instance.

What does the extra layer get us in this instance? If we Have a reference to the wrapper and simply release the `unamanged` resource, is there anything more required? (I'm not opposed to the idea, I'm just unsure what the extra overhead gets us)

> I'm not sure OnLoadingStateChange is the correct context for disposing of the instances

I'm not sure either, it was just a though in regards to the `Frame Wrapper`. Like when the next page had finished loading we dispose of all the previous frame instances. This might end up being one of those that's left to the user to properly manage (which is bound to cause leaks). The `async` callbacks are equally as problematic.

I'm totally open to suggestions, I know what your trying to achieve, and it makes sense :smile: Still no clearer on how we'd go about it.

rassilon: > For  CefBrowser  it's pretty clear and I think that's pretty straight forward to reuse the wrapper. For  CefFrame , I'm not sure that there is such a notification.

Well, there's good news and bad news. There seems to be a notification, but it's only in the rendering process as best I can see looking through the .h files: `CefRenderProcessHandler::OnContextReleased`.

Ugh.

> What does the extra layer get us in this instance? If we Have a reference to the wrapper and simply release the  unamanged  resource, is there anything more required? (I'm not opposed to the idea, I'm just unsure what the extra overhead gets us)

I suppose as long as we have some kind of: `IsDisposed` property on the object, the effect would be the same. We'd set it to `true` when CEF tells us the underlying thing is no longer useful. Then we would add all of those fun `ThrowIfDisposed` calls to the wrappers. Whee! :smilie:

Bill

amaitland: > Then we would add all of those fun ThrowIfDisposed calls to the wrappers.

Least they're simple to implement and follow, no extra layer of indirection :smile:

CefSharp.Core/Internals/ClientAdapter.cpp
```diff
@@ -247,8 +262,23 @@ namespace CefSharp
...
+            if (browser->IsPopup())
+            {
+                auto popupHandler = _browserControl->PopupHandler;
+                if (popupHandler != nullptr)
+                {
+                    auto browserId = browser->GetIdentifier();
+                    IBrowser^ entry;
+                    if (_popupBrowsers->TryGetValue(browserId, entry))
+                    {
+                        popupHandler->OnPreKeyEvent(_browserControl, entry, (KeyType)event.type, event.windows_key_code, event.native_key_code, (CefEventFlags)event.modifiers, event.is_system_key == 1, *is_keyboard_shortcut);
```
amaitland: Another missing `return`?

CefSharp.Core/Internals/ClientAdapter.cpp
```diff
@@ -285,15 +315,15 @@ namespace CefSharp
...
-                        popupHandler->OnFrameLoadEnd(_browserControl, entry, gcnew FrameLoadEndEventArgs(frameWrapper, httpStatusCode));
+                        CefFrameWrapper frameWrapper(frame, _browserAdapter);
+                        popupHandler->OnFrameLoadEnd(_browserControl, entry, gcnew FrameLoadEndEventArgs(%frameWrapper, httpStatusCode));
                     }
                 }
             }
             else
             {
-                _browserControl->OnFrameLoadEnd(gcnew CefFrameWrapper(frame, _browserAdapter), httpStatusCode);
+                CefFrameWrapper frameWrapper(frame, _browserAdapter);
```
amaitland: Can you explain the subtle change here? What's the advantage of passing in `%frameWrapper`?

rassilon: Can skip the delete since the type doesn't end in ^. 
They're the same otherwise.

amaitland: Interesting. I take it the lifespan is only valid for the life of the call? Should this be the standard approach for all those that we would otherwise immediately dispose of?

amaitland: > I take it the lifespan is only valid for the life of the call?

For future reference, looks like the answer to my question.
http://stackoverflow.com/a/10233773

CefSharp.Core/Internals/ClientAdapter.cpp
```diff
@@ -262,9 +292,9 @@ namespace CefSharp
                     IBrowser^ entry;
                     if (_popupBrowsers->TryGetValue(browserId, entry))
                     {
-                        auto popupHandler = _browserControl->PopupHandler;
                         auto frameWrapper = gcnew CefFrameWrapper(frame, _browserAdapter);
                         popupHandler->OnFrameLoadStart(_browserControl, entry, gcnew FrameLoadStartEventArgs(frameWrapper));
```
amaitland: Should `IBrowser` be exposed through `FrameLoadStartEventArgs` as well maybe?

amaitland:
Now that `IPopupHandler` is almost finished, what do you think overall? Is the approach sound? Would we have been better to call the default handlers?

rassilon:
> Now that  IPopupHandler  is almost finished, what do you think overall? Is the approach sound? Would we have been better to call the default handlers? 

I think `IPopupHandler` itself is ok. I'm still concerned about our lifetime management of the `CefFrameWrapepr/CefBrowserWrapper` instances we hand out to callers to CefSharp from `IFrame/IBrowser` i.e. IFrame's get parent, IFrame's get IBrowser, etc...

Since for now the `IPopupHandler` method calls can at least have those instances disposed after the call returns.

amaitland:
Well for `IFrame.GetParent()` and `IFrame.GetBrowser()` we can cache the instances. So calling `IFrame.Dispose()` would release any of the child resources. That should improve things a little. Apart from documentation and clear examples I'm not sure we can neatly take care of all aspects of the `umanaged` resource management.

I've modified the callbacks slightly so they get released after the call, I think that should be fairly practical.
https://github.com/amaitland/CefSharp/commit/95fcf1623f740204d6b7e161f5461380ced9efe3

CefSharp.Core/Internals/ClientAdapter.cpp
rassilon: Works for me.

rassilon:
Ok, I think the disposes for `IFrame\IBrowser` in  `CefFrameWrapper` and `ClientAdapter.cpp` generally look good so far. `CefBrowserWrapper` still can create non-obviously disposable IFrame instances, atm.

Bill

CefSharp.Core/Internals/ClientAdapter.cpp
rassilon: Yeah, I've pushed a fix for this I think.

rassilon:
Btw, I'm done pushing commits tonight if you want to push through the branch shuffling.

CefSharp.Core/Internals/CefFrameWrapper.cpp
```diff
@@ -234,10 +234,24 @@ Int64 CefFrameWrapper::Identifier::get()
 IFrame^ CefFrameWrapper::Parent::get()
 {
     ThrowIfDisposed();
-    auto parent = _frame->GetParent();
-    if (parent != nullptr)
+    if (_parentFrame != nullptr)
```
amaitland: Should this be `==`? Always going to return `nullptr` otherwise.

rassilon: Yeah, I've cleaned it up locally to look more like the IBrowser returning method in this class.

amaitland: No prob :+1:

amaitland:
> Btw, I'm done pushing commits tonight if you want to push through the branch shuffling.

I've pushed the changes directly to `master` for now. I've rearranged the `milestones`, though I've yet to tackle the rearranging of branches and trying to re-target the outstanding PR's.

Maybe going forward we go with a `development` branch and `master`, or even a couple of branches, `master`, `vNext` and `development`. Thoughts?

rassilon:
Using a development branch and master might indeed make the most sense.

amaitland:
Was thinking about this more, perhaps we should adopt a similar strategy to `CEF`, all development is done against `master` and create a release branch for every matching `CEF` version. Think I'm in favor of this approach. Any preference?

rassilon:
I have no particular preference. The CEF approach probably makes slightly more sense.

Bill

rassilon:
I've merged master into the contents of the cef/2272 branch into a local branch,  what should we name the release branches: cefsharp/41, cefsharp/43, etc? 

Once we decide I can push the merged master/cef/2272 stuff there. 

Bill

amaitland:
> what should we name the release branches: cefsharp/41, cefsharp/43, etc?

I'm happy with that. I was originally thinking of following `CEF` and naming them `2272`, etc. though your idea is clearer.

I added `37`, `39` and `41`

https://github.com/cefsharp/CefSharp/branches/active

Let me know when your done then I'll merge in #1053

rassilon:
Ok, i merged the old cef/2272 into master since you created cefsharp/41 where master was at. I've got a pretty big sinus headache, you can merge in #1053 when you're ready. If I'm feeling better I can merge this huge chunk in tomorrow.

Bill

amaitland:
Merged this directly into `master` see https://github.com/cefsharp/CefSharp/commit/801bebbbd088a2948a9a7330c4d19cb772cb9c77

I'll double check later, it builds. Don't think I made any silly mistakes with the merge

Jayman1305:
On Wed Dec to w

Sent by Outlook<http://taps.io/outlookmobile> for Android



On Tue, Jun 9, 2015 at 8:36 AM -0700, "Bill Tutt" <notifications@github.com<mailto:notifications@github.com>> wrote:


Using a development branch and master might indeed make the most sense.

-

Jayman1305:
Sent by Outlook<http://taps.io/outlookmobile> for Android



On Tue, Jun 9, 2015 at 8:36 AM -0700, "Bill Tutt" <notifications@github.com<mailto:notifications@github.com>> wrote:


Using a development branch and master might indeed make the most sense.

-

Issue #1067, opened 2015-06-09
IKeyboardHandler doesn't allow you to easily propagate the event into WinForms from CEF


rassilon:
It sure would be nice to show our users how to forward keyboard events from CEF to WinForms.

I'm not sure if OSR/WPF have this problem because OSR/WPF feed events into CEF, so they ought to be able to do their message preprocessing themselves. (I'm guessing, I don't enough about WPF to be sure.)

Bill

rassilon:
For WinForms this seems as simple as:

* Pass the `MSG` though to `IKeyboardHandler` using the WinForms `Message` structure as an extra parameter to both `OnKeyEvent\OnPreKeyEvent`.
* Change the supplied `HWND` to `WinForms.ChromiumWebBrowser.Handle`
* Supply this altered `MSG` instance to the user.
* Expose an API or callback in order to call `PostMessage` in order for the supplied `MSG` to be delivered. 
* While we're at it, we're really putting too many parameters on these functions, we should probably consolidate into a struct/class or something.

Based on how the WinForms message loops work, the normal WinForms logic of searching parent ContainerControls for the first one to handle the message (if applicable) would just happen.

@amaitland , thoughts? The idea with including this change for `OnPreKeyEvent` is they want to ensure the key sequence is handled by WinForms specifically.

Sadly, I still need to look into how Chromium passes the different KeyType key events through CEF, because CEF doesn't have ANY example keyboard handlers that are even mildly interesting.

I also would like to know more about the `is_keyboard_shortcut` behavior, the CEF documentation here is lacking about what setting that to `true` *actually* does.

Bill

rassilon:
I was thinking about adding a `Ctrl-T` (or something similar) to the CefSharp example WinForms app that this could invoke.

rassilon:
#1068 makes this good enough to close for now.

PR #1068, opened 2015-06-09
Make isKeyboardShortcut actually settable from the handler, and impro…

rassilon:
…ve the IKeyboardHandler documentation.

@amaitland , thoughts for 2272?

I was thinking any more would need to wait for 2357. :frowning: 

Bill

### DIFF
--- a/CefSharp.Core/Internals/ClientAdapter.cpp
+++ b/CefSharp.Core/Internals/ClientAdapter.cpp
@@ -195,8 +195,14 @@ namespace CefSharp
             {
                 return false;
             }
+            System::Boolean isKeyboardShortcut(*is_keyboard_shortcut);
+            auto result = handler->OnPreKeyEvent(
+                _browserControl, (KeyType)event.type, event.windows_key_code, 
+                event.native_key_code, (CefEventFlags)event.modifiers, event.is_system_key == 1, 
+                isKeyboardShortcut);
 
-            return handler->OnPreKeyEvent(_browserControl, (KeyType)event.type, event.windows_key_code, event.native_key_code, (CefEventFlags)event.modifiers, event.is_system_key == 1, *is_keyboard_shortcut);
+            *is_keyboard_shortcut = isKeyboardShortcut;
+            return result;
         }
 
         void ClientAdapter::OnLoadStart(CefRefPtr<CefBrowser> browser, CefRefPtr<CefFrame> frame)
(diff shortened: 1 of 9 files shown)


amaitland:
> @amaitland , thoughts for 2272?

As in you'd like to merge this into `master`? I'm fine with that.

rassilon:
I've added a sample that works to WinForms.Example. Not utterly complete, but good enough for most WinForm accelerator interactions I would think. It was easier than I thought. 

Bill

CefSharp/Internals/BrowserProcessServiceHost.cs
```diff
@@ -52,13 +52,15 @@ public void SetOperationContext(OperationContext operationContext)
...
+            // Use TaskScheduler.Default to ensure that the work occurs
+            // on a background thread.
             return operationContextTask.ContinueWith(t =>
             {
                 var context = t.Result;
                 var renderProcess = context.GetCallbackChannel<IRenderProcess>();
                 var asyncResult = renderProcess.BeginEvaluateScriptAsync(browserId, frameId, script, timeout, null, null);
                 return Task.Factory.FromAsync<JavascriptResponse>(asyncResult, renderProcess.EndEvaluateScriptAsync);
-            }).Unwrap();
+            }, TaskScheduler.Default).Unwrap();
```
rassilon: Bah. I didn't mean to push this up. Seems pretty boring though. Thoughts? The idea is we're not guaranteed that the continuation (`ContinueWith`) *WILL NOT* execute on the current thread.

amaitland: Would have been ideal as a separate commit, oh well :smile: 

Makes sense to execute on the `ThreadPool`. I think adding something to the documentation would also be worthwhile.

CefSharp.Core/Internals/ClientAdapter.cpp
```diff
@@ -195,8 +195,14 @@ namespace CefSharp
             {
                 return false;
             }
+            System::Boolean isKeyboardShortcut(*is_keyboard_shortcut);
```
amaitland: I didn't think it was necessary to do anything more than pass in `*is_keyboard_shortcut` as a `ref` param?

At least it appeared to be working when I added `no_javascript_access` to `OnBeforePopup`

https://github.com/cefsharp/CefSharp/pull/935/files#diff-2731f23ba4e8fcf96f8501a49c6d687eR53

rassilon: It isn't, updated.

rassilon:
This fixes #1067 enough for me for the time being.

rassilon:
Updated 41 WIP in Changelog.

Fyi,
Bill

amaitland:
Nice :+1:

PR #1072, opened 2015-06-11
Continued API Rewrite

amaitland:
Trying to reduce the code complexity a little bit. 

More changes to come.

### DIFF
--- a/CefSharp.Core/Internals/ClientAdapter.cpp
+++ b/CefSharp.Core/Internals/ClientAdapter.cpp
@@ -672,14 +672,13 @@ namespace CefSharp
         void ClientAdapter::OnBeforeContextMenu(CefRefPtr<CefBrowser> browser, CefRefPtr<CefFrame> frame,
             CefRefPtr<CefContextMenuParams> params, CefRefPtr<CefMenuModel> model)
         {
-
             auto handler = _browserControl->MenuHandler;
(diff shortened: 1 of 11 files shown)


rassilon:
I was doing just that and I'm debugging a MCefRefPtr related garbage hitting the finalizer or indeed not at all and CEF assert during CEF.Shutdown.

Bill

amaitland:
Merge this `PR` so you can remove `DisposableResource`?

rassilon:
Sure.

amaitland:
Done. Still quite a few cleanups to go. Will look at creating another `PR` later. Need to review all the callbacks, make sure weren't not disposing the wrappers too quickly for them to be used an an `async` fashion.

rassilon:
The callbacks don't clean themselves up.  However, the fix for them looks
simple.

amaitland:
I was thinking more along the lines of when you pass in a `CefFrameWrapper` instance, making sure that it can be accessed from within a `Task` when using a callback. I'm not familiar with the `%frameWrapper` syntax enough to know exactly when the wrapper will be disposed. So I wanted to double check.

amaitland:
I also changed the callbacks so after you execute they set the underlying `MCefRefPtr` instance to `NULL`. If you've got a cleaner way then I'm all for it :+1:

PR #1077, opened 2015-06-13
Reduce objects being finalized

rassilon:
Some changes that help reduce the # of objects are called from the finalization thread.


amaitland:
Other than a few talking points I think this is a major improvement :+1: Fantastic work! Thanks :smile: 

As there are a large number of subtleties with `VC++` (I learn something new every other week!). What do you think about collecting a few reference articles? Maybe a wiki entry?

I know I would have benefited from reading something like the following MSDN article earlier.
https://msdn.microsoft.com/en-us/library/ms177197%28v=vs.80%29.aspx

PR #1079, opened 2015-06-15
Continued Work on 43

amaitland:
Start with wrapping `GetFocusedFrame` calls in `using`

More changes to come...


amaitland:
I've rewritten the `Scheme/ResourceHandler` code again. It's more in line with the `CEF API` now. It's probably `85%` finished. The code should be a lot easier to follow now, it was pretty mind bending before. Let me know what you think.

rassilon:
I merged my ReduceObjectsBeingFinalized PR into this one. I haven't looked at the Scheme/ResourceHandler updates you've made yet, but the conflicts were really tiny.

Bill

rassilon:
> Scheme/ResourceHandler changes

They look good to me.

Bill

rassilon:
> Scheme/ResourceHandler changes

Err, the SchemeHandlerFactory lifetime management of IBrowser^ and IFrame^ as well as the similar parameters for ResourceHandlerFactory need documentation about how the data is only good for that function call or something. Otherwise, we need to allow those instances to last longer before being disposed.

This is of course similar to the problem that CefSharpBrowserWrapper has for all of the CefFrameWrapper creations that don't have a definitive lifetime associated with them.

I wouldn't let it block you merging this PR though.

amaitland:
> I merged my ReduceObjectsBeingFinalized PR into this one. I haven't looked at the

Great, thanks :+1: 

> They look good to me.

Exposing the callback makes things a lit simpler, the code is much more hierarchical now, was very hard to follow.

Does returning a `Stream` for `IResourceHandler.GetResponse()` make sense? I know you had a rough idea about reading the stream on a different `Thread`. Can you see any changes that would need to be made to accommodate that?

amaitland:
@rassilon Thanks for merging your code in :+1:

PR #1088, opened 2015-06-20
Add possibility to avoid using WCF

arsher:
Hi,

I'm creating this based on the advice from @amaitland as it's a pretty complex changeset and I'm not even sure if you'd consider merging something like this. So it's not prod ready and also missing some things I imagined.

The story behind this is that I was using the lib in an extreme edge case where I rapidly created and destroyed over 15 browser instances and was injecting a lot of JS as well. The outcome of this was some occasional (but consistent) crash because of WCF and I just couldn't figure out what's going on and needed it working really quick, so I hacked together a solution where everything was going over CEF IPC instead of WCF and that made it pretty stable for my purposes, and also it seemed to be a huge leap in performance (I'm doing a lot c#/js interop). This was pretty hacky but worked, so I decided to make a presentable version in my spare time, and try to merge it :-).

During the development of my solution I ran into a major obstacle, namely that the cef guys didn't expose the sync IPC solution from chromium. This means that sync JS bindings are not possible this way, which also means that I couldn't implement property bindings as the accessors have to run sync, but ipc is happening on the thread where js runs...this is fine with me as I don't use it, but of course this means that it's a loss of feature. That's why I decided to make wcf switchable in my less hacky version. As you can see currently I moved as much comm to CEF IPC as possible and currently also working on a method bindings which of course would be async. I actually find that better because there is no way to block the render thread unnecessarily (hence the perf leap in my case), but it also means having to code against the bound csharp object in an async fashion with callbacks.

Uhh this became long, but this is the whole story....so feel free to comment on it, I can take it :-D. Also I know it's extremely big, but I just couldn't figure out how to do this in small steps as everything is connected through wcf :-(.

David


amaitland:
Thanks @arsher! 

> it's a pretty complex changeset and I'm not even sure if you'd consider merging something like this. So it's not prod ready and also missing some things I imagined.

I've pondered replacing the `WCF` implementation with `IPC` and it's quite a daunting task! I think the concept is sound in general :+1: 

> The outcome of this was some occasional (but consistent) crash because of WCF

Your not the first to experience this, and as you've found out it's a difficult case to debug :frowning: 

With the overall complexity of this I'm hopeful we can split the task up into a few smaller features. The starting point I think would simply be to make the `WCF` service optional, for those that don't requite any js binding or execution. Thoughts?

amaitland:
I think the second step would be to switch `EvaluateScriptAsync` from `WCF` to `IPC` as this can likely happen a self contained task.

A while ago `CEF` introduced a new message router class, while I don't think it fits our purpose exactly I think it's worth taking a look at for inspiration. @rassilon may have some ideas here as he's the one who alerted me to it's existence.

http://magpcss.org/ceforum/apidocs3/projects/%28default%29/cef_message_router.h.html

rassilon:
Ooh. This could be very cool. I probably won't get a chance to look at this until Tuesday, but I'm all in general favor of ditching wcf for the javascript binding.

arsher:
Ok, I think I'm going to create multiple branches for this then. Is there any better/other way to separate this out into multiple PRs?

amaitland:
> Is there any better/other way to separate this out into multiple PRs?

Multiple branches is the only way that I know of :+1:

arsher:
Ok I started off with the first separate PR. I think I will close this one.

PR #1094, opened 2015-06-24
Add Ability to Disable WCF

arsher:
added switch to turn off wcf (this disables js bindings and eval script)


amaitland:
Thanks @arsher I'll look at merging this in shortly.

rassilon:
Looks good to me.

amaitland:
I made a few additional commits to `master`, most important being https://github.com/cefsharp/CefSharp/commit/ff695dbf6d3f9c24f1db041c3cce9dde23822fe7

amaitland:
I had also considered moving the `WcfEnabled` flag onto the `Cef` object, so it could be changed at runtime (under certain circumstances). Thoughts?
(There are some technical limitations to this as the `VC++` compiler complains, so we'll likely have to create a new object to store the bool).

PR #1095, opened 2015-06-25
CEF IPC based serialization and communication for EvaluateScriptAsync

arsher:
Ok, here goes the second set of changes, it's a bit more complex. :).
The changes include:
* Added the necessary tools for de/serialization from/to ceflistvalue/cefdictionaryvalue.
* JS Binding data is transfarred through CEF IPC
* EvaluateScriptAsync is done through CEF IPC
  * ProcessMessageHandler to delegate process messages to different handlers from ClientAdapter and CefAppUnmanagedWrapper.
  * Handler implementation for the above mentioned functions implemented on cef ipc
  * Removed unnecessary methods from CefRenderProcess, BrowserProcessServiceHost, CefBrowserWrapper 
  * Tasks for eval script generated with TaskCompletionSources, all of them gets cancelled upon dispose

+ all the necessary stuff to acommodate this


amaitland:
> EvaluateScriptAsync is done through CEF IPC

How hard is it to split just this part out? I think the practical approach is to split this set of changes into smaller PR's still. Otherwise in all likely hood this will just sit here for weeks if not months.

amaitland:
Just general feedback on a whole
- Add license disclaimer to new files
- Spaces not tabs
- Prefer using statements over namespace qualified classes (unless there's a conflict)

As a general rule it's nicer to have smaller more granular commits, makes the external review a lot easier. Also slightly more descriptive comments where possible.

amaitland:
Also need to consider how we can handle the proposed changes in #972

CefSharp.BrowserSubprocess.Core/Messaging/EvaluateScriptDelegate.h
```diff
@@ -0,0 +1,29 @@
...
+
+namespace CefSharp
+{
+    class CefAppUnmanagedWrapper;
+
+    namespace Internals
+    {
+        namespace Messaging
+        {
+            class EvaluateScriptDelegate : public ProcessMessageDelegate
```
rassilon: Additoinally, could new classes have nice short summary contents in their .h files explaining how the class fits into the overall picture. Like for this one: Recieves IPC requests from browser process and forwards adapts them into calls against the correct V8 JS context. (or something)

arsher:
I probably should've spent a bit more time on factoring things out. I'm working on this now, also the missing parts like license disclaimers :). I'm trying to minimize the changes but unfortunately I think I'm gonna have to create new commits instead of these ones to do it right.

arsher:
ok, so new commits, bit smaller ones, and also this only contains the evaluatescriptasync feature

amaitland:
@arsher Fantastic :+1: So much easier to review now :smile: Thanks!

amaitland:
For those that are interested, my comments are in the form of a `PR`, see https://github.com/arsher/CefSharp/pull/2

arsher:
Cool :). I'll continue with the js callbacks stuff. I have to revise it a bit in the light of the comments in this thread, but should be done soon.

PR #1099, opened 2015-07-01
Add CefWindowInfoWrapper

amaitland:
Add Wrapper for CefWindoInfo (OnBeforePopup)

### DIFF
--- a/CefSharp/IWindowInfo.cs
+++ b/CefSharp/IWindowInfo.cs
@@ -1,0 +1,25 @@
+﻿// Copyright © 2010-2015 The CefSharp Authors. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style license that can be found in the LICENSE file.
+
+using System;
+
+namespace CefSharp
+{
+    public interface IWindowInfo : IDisposable
+    {
+        int X { get; set; }
+        int Y { get; set; }
+        int Width { get; set; }
+        int Height { get; set; }
+        uint Style { get; set; }
+        IntPtr ParentWindowHandle { get; set; }
+        bool TransparentPaintingEnabled { get; set; }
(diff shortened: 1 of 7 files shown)


CefSharp/IWindowInfo.cs
```diff
@@ -0,0 +1,25 @@
...
+        int Y { get; set; }
+        int Width { get; set; }
+        int Height { get; set; }
+        uint Style { get; set; }
+        IntPtr ParentWindowHandle { get; set; }
+        bool TransparentPaintingEnabled { get; set; }
+        bool WindowlessRenderingEnabled { get; set; }
+        IntPtr WindowHandle { get; set; }
+
+        void SetAsChild(IntPtr parentHandle, int x, int y);
```
rassilon: This stuff looks good, but we should probably put an edited version of CEF's documentation for these methods here.  As I would have to go look at CEF to figure out exactly what the difference between `SetAsChild` and `SetAsPopup` are. Ditto for Windowless, i.e. is that CEF's API pseudonym for OSR here? (as opposed to windowless as in doesn't have HWND for non-OSR apps)

amaitland: Fair point. 

> Ditto for Windowless, i.e. is that CEF's API pseudonym for OSR here? (as opposed to windowless as in doesn't have HWND for non-OSR apps)

`Create the browser using windowless (off-screen) rendering.`.

Technically you doesn't have to pass in a handle for `OSR`,

> The |parent| value will be used to identify monitor info and to act as the parent window for dialogs, context menus, etc. If |parent| is not provided then the main screen monitor will be used and some functionality that requires a parent window may not function correctly.

http://magpcss.org/ceforum/apidocs3/projects/%28default%29/CefWindowInfo.html

PR #1100, opened 2015-07-01
Throw Exceptions rather than Fail quietly

amaitland:
So previously `CefSharp` would just silently do nothing for the bulk of it's functions when the browser hadn't initialized. For the most part the major rewrite going on in `master` has meant that a lot of code is now run using extension methods.

This PR is an attempt to either move the final methods into extensions and reuse the `Throw` methods we have, or explicitly throw an exception.

The issue that sparked this is http://stackoverflow.com/questions/31155779/cefsharp-execute-javascript


amaitland:
Still likely a few that have been missed, so won't merge this straight away.

I also removed some code duplication (using extension methods).

rassilon:
Looks good to me. I haven't looked for things you've missed though, it's clearly an improvement.

amaitland:
There's still a number of the `CefBrowserHost` methods that could be moved to the `IBrowserHost` implementation to remove the duplication, will add the additional checks there.

amaitland:
Hopefully a nice error message will reduce the number of questions we get.

amaitland:
I've moved some more methods and added an exception for calling `RegisterJsObject` after the browser has been initialized.

amaitland:
@rassilon I've merged the small outstanding PR"s into `master`, so feel free to make any followup changes regarding locking :+1:

PR #1102, opened 2015-07-01
BrowserSettings map to Enabled/Disabled/Default Enum

amaitland:
Decompiled code showed `ValueType` before and how shows the correct type

Need to double check over the `cef_state_t` conversions. I've simplified them, though they can likely be simplified again.

### DIFF
--- a/CefSharp.Core/BrowserSettings.h
+++ b/CefSharp.Core/BrowserSettings.h
@@ -107,19 +107,19 @@ namespace CefSharp
             void set(CefState value) { _browserSettings->javascript = (cef_state_t)value; }
         }
 
-        property CefState JavaScriptOpenWindows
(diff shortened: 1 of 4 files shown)


rassilon:
It showed `ValueType` before and now shows `Nullable<...>`?

Bill

rassilon:
Never mind, I saw the issue that spawned this PR.

amaitland:
> It showed ValueType before and now shows Nullable<...>?

It shows `bool?` now, which is much nicer :smile: 

I've been thinking about this a little more and I'm not sure that `bool?` accurately represents the `Default/Enabled/Disabled` state, it would be a lot cleaner in my option just to use a simple `enum` that has three values. Thoughts?

PR #1105, opened 2015-07-03
Display Release Mode Message Box In Examples

amaitland:
Add Message Box into WinForms and WPF examples to alert users they need to compile in Release mode when running outside of Visual Studio


jankurianski:
Nice one :+1:  Will hopefully help stop this common error.

amaitland:
Hopefully :smile: It doesn't stop you from running the app. I thought about throwing an exception though that would limit your ability to run it outside of `VS`, so though it was an ok compromise.

PR #1107, opened 2015-07-04
Sign the OffScreen project (using existing snk file)

mkosieradzki:
Cherry-picked commit adding missing strong name to CefSharp.Offscreen. If possible update nuget package after integrating.


amaitland:
New package pushed to `nuget.org`

Issue #1110, opened 2015-07-06
v41: NullReferenceException at startup in WPF ChromiumWebBrowser


jamespearce2006:
I've just switched my app from v39 to v41, but ChromiumWebBrowser always seems to throw a NullReferenceException at startup.

In OnApplyTemplate hasn't been called yet because the browser hasn't yet been added to the Visual Tree, any attempt to set the Address (manually, or via binding) will fail.

The NullReferenceException is thrown in Load() - the call to GetMainFrame() returns null.

Since it's relatively likely that an address will be set before the control appears (particularly if the MVVM  pattern is being used), this seems like a problem...

amaitland:
> I've just switched my app from v39 to v41, but ChromiumWebBrowser always seems to throw a NullReferenceException at startup.

Your saying this worked in `39`?

Your welcome to submit a `PR` against `41`, I'm reluctant to spend too much time on that branch as `master` is a huge rewrite still requiring a lot more effort and refinement before it's released.

jamespearce2006:
OK - I think it's a relatively easy change.
Note - I'm looking at master... I'm confused by your comment about the 41 branch.

amaitland:
`master` is very much a work in progress and not ready for prime time yet. So best to base your `PR` on https://github.com/cefsharp/CefSharp/tree/cefsharp/41

It's unlikely that `41` will see many more commits as it was simply intended to upgrade the `CEF` version with a couple of small `API` fixes along the way.

jamespearce2006:
@amaitland I based on master before I saw your comment...
Still, it's a simple change.

amaitland:
> Still, it's a simple change.

That's fine.

PR #1111, opened 2015-07-06
NullReferenceException when Address is set before OnApplyTemplate

jamespearce2006:
Fix NullReferenceException in WPF ChromiumWebBrowser when Address is set before OnApplyTemplate is called

See #1110

### DIFF
--- a/CefSharp.Wpf/ChromiumWebBrowser.cs
+++ b/CefSharp.Wpf/ChromiumWebBrowser.cs
@@ -905,6 +905,12 @@ namespace CefSharp.Wpf
             Content = image = CreateImage();
 
             popup = CreatePopup();
+
+            // If Address was previously set, only now can we actually do the load
+            if (!string.IsNullOrEmpty(Address))
+            {
+                Load(Address);
+            }
         }
 
         private Image CreateImage()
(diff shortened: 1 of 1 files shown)


CefSharp.Wpf/ChromiumWebBrowser.cs
```diff
@@ -1243,8 +1249,9 @@ private void OnMouseButton(MouseButtonEventArgs e)
 
         public void Load(string url)
         {
-            //Added null check -> binding-triggered changes of Address will lead to a nullref after Dispose has been called.
-            if (managedCefBrowserAdapter != null)
+            // Added null check -> binding-triggered changes of Address will lead to a nullref after Dispose has been called
+            // or before OnApplyTemplate has been called
+            if (managedCefBrowserAdapter != null && GetMainFrame() != null)
```
amaitland: Better to get a reference to the frame, check if it's null, and then use that reference.
The `managedCefBrowserAdapter != null` check becomes redundant as `GetMainFrame` will return null if `managedCefBrowserAdapter == null`.

amaitland:
Minor changes to merge this in see https://github.com/cefsharp/CefSharp/commit/3773edb986ccf9e84e78d7d0cf84b00e8ed17044

amaitland:
I've actually had to reorder things a little as the page wasn't always loading correctly. I think this is likely a result of the refactoring.

Change is here
https://github.com/cefsharp/CefSharp/commit/9798772694fd3e6479557e369811659af043a8c4

Issue #1106, opened 2015-07-04
IWebBrowser Load/LoadHtml url formatting issue


lamerbug:
I was checking out latest builds and ran into issue with url formatting when using Load/LoadHtml methods. I've implemented custom resource handler just like in void RegisterTestResources(IWebBrowser browser)
and noticed that "http://test/resource/load" example url works fine, but when I change it to "http://test" it doesn't load later, because when Load fires url turns into "http://test/" with additional slash at the end, and looks like it makes sense. This happens only with short URLs:
http://test becomes http://test/ when loading
http://te.st/test becomes http://te.st/test/
http://test/test/okay works ok.
and because of that LoadHtml doesn't work for short urls too. 
The same story happens when loading regular web urls.

not sure if I'm able to fix it now (just started to learn changes in latest cefsharp since v37) so decided to let u know...

amaitland:
I think the class your looking for is `DefaultResourceHandlerFactory`. It just uses a basic dictionary (needs to be changed to a `ConcurrentDictionary` at some point). Will likely need to use `Uri.TryCreate` to turn the `string url` into a proper `Uri` including the trailing slash on the `hostname`

https://github.com/cefsharp/CefSharp/blob/master/CefSharp/DefaultResourceHandlerFactory.cs#L38

lamerbug:
Yeah, since incoming IRequest (at GetResourceHandler) already contains canonical URL, we just turn string into canonical URL when registering it. Works well so far... 
```C#
public virtual bool RegisterHandler(string url, IResourceHandler handler)
        {
            try
            {
                Uri uri;
                if (Uri.TryCreate(url, UriKind.Absolute, out uri))
                {
                    Handlers.TryAdd(uri.ToString(), handler);
                    return true;
                }
                else
                {
                    return false;
                }
            }
            catch
            {
                return false;
            }
        }
```
(I decided to return bool to be sure if it registered or not, just in case of bad string)

amaitland:
You can scrap the `try...catch` block and should probably return `Handlers.TryAdd`. Also the `else` block is redundant since returning in the `if` block.

lamerbug:
Sure, I just had some other code in that blocks for testing purposes,

amaitland:
Feel free to submit a `PR` against `master` if you'd like to see this changed :+1:

lamerbug:
Not sure if I did a PR correctly, be careful )

amaitland:
[...]

PR #1112, opened 2015-07-06
Update DefaultResourceHandlerFactory.cs

lamerbug:
- using ConcurrentDictionary
- create canonical URL from any string
- update value if key exists
- return bool result when register and unregister


CefSharp/DefaultResourceHandlerFactory.cs
```diff
@@ -3,27 +3,34 @@
...
+        public virtual bool RegisterHandler(string url, IResourceHandler handler)
         {
-            Handlers[url] = handler;
+            Uri uri;
+            if (Uri.TryCreate(url, UriKind.Absolute, out uri) || Uri.TryCreate("http://" + url, UriKind.Absolute, out uri))
+            {
+                Handlers.AddOrUpdate(uri.ToString(), handler, (k, v) => handler);
+                return true;
+            }
+                return false;
```
amaitland: Minor formatting issue.

CefSharp/DefaultResourceHandlerFactory.cs
```diff
@@ -3,27 +3,34 @@
...
-            Handlers = new Dictionary<string, IResourceHandler>(comparer ?? StringComparer.OrdinalIgnoreCase);
+            Handlers = new ConcurrentDictionary<string, IResourceHandler>(comparer ?? StringComparer.OrdinalIgnoreCase);
         }
 
-        public virtual void RegisterHandler(string url, IResourceHandler handler)
+        public virtual bool RegisterHandler(string url, IResourceHandler handler)
         {
-            Handlers[url] = handler;
+            Uri uri;
+            if (Uri.TryCreate(url, UriKind.Absolute, out uri) || Uri.TryCreate("http://" + url, UriKind.Absolute, out uri))
```
amaitland: In this case if they don't provide a `scheme name` then I'd rather throw an exception than implicitly add `http`.

amaitland:
Fixes #1106

lamerbug:
>>In this case if they don't provide a scheme name then I'd rather throw an exception than implicitly add http.

I think it's ok, That's a default behaviour of UriBuilder, and browser does the same. So user can get working any string. (I have a use case when user allowed to manually register it)

rassilon:
> No, I think it's ok, That's a default behaviour of UriBuilder, and browser does the same. So user can get working any string. (I have a use case when user allowed to manually register it)

I think I'm with @amaitland here, if you'd like to do something specific for "urls" without schemes, I would suggest doing so at the application level. Sort of like Chrome/IE change the request into a google/bing/whatever search for such URLs if its not obviously domain name like.

So, I think you'd have to explain to us a bit more why CEFSharp prevents you from doing something like that at the application level.  My gut instinct is to infer that while your application might be able to be tweaked to do such a thing, that for various other reasons its just cheaper/less risk/etc... to make this change in CEFSharp.

The ConcurrentDictionary part is just fine of course. I've even commented about needing to getting around to doing it in the past.

lamerbug:
>So, I think you'd have to explain to us a bit more why CEFSharp prevents you from doing something like that at the application level.

Forgot to mention that while resolving #1106 I just was implementing custom ResourceHandlerFactory so of course any changes to DefaultResourceHandlerFactory is a subject to review from your side.

> that for various other reasons its just cheaper/less risk/etc... to make this change in CEFSharp.

We have to canonize url anyway to get it working. If there would be a choice to do so or not, then of course we would move all the logic to application level.

amaitland:
> We have to canonize url anyway to get it working. If there would be a choice to do so or not, then of course we would move all the logic to application level.

We'll keep that change of course :+1:

PR #1113, opened 2015-07-07
Add IPostData to IRequest

### LINKED ISSUE #1038: Feature Request - IRequest Access CefPostData
When implementing `IRequestHandler` there is no way to edit post data of the request in `OnBeforeBrowse` method

amaitland:
Resolve #1038

It's not very well tested.

TODO:
- `throw NotImplementedException` for methods that don't do anything yet.
 - Dispose of `IPostDataElement's`if possible
- Check `IsReadonly` flag before attempting to change data


amaitland:
Also remove commented out code block

amaitland:
@jankurianski I know you made some changes to the `Body` method, which I've just removed. If you have a chance you might want to take a look, see if I've forgot to implement something your dependent on?

jankurianski:
All good - I appreciate you leaving in the important bits for me (sniffing of charset, converting post data byte[] to string). :+1:

rassilon:
The bits that you removed seem good enough to do the correct thing and seem a sensible default.
 
If we had a logging system at the CefSharp layer, I'd add a verbose event for an unknown charset.

The extra cool thing with this apporach is that the user can use an alternative implementation if they like since nothing internal needs a GetBody implementation. 

We should at least add the doc comments about how it works though.

The only other possible implementation choice for GetBody I can think up right now would be to throw an exception about the unknown charset and force the consumer to figure it out with the raw byte array.

We could have utilize a default parameter value to allow the API consumer such a choice or something.

Bill

PR #1114, opened 2015-07-07
Javascript callbacks moved to cef ipc

arsher:
Aside from moving the js callbacks to cef ipc I fixed a potential problem in the serialization code.


CefSharp.BrowserSubprocess.Core/CefAppUnmanagedWrapper.h
```diff
@@ -19,6 +19,28 @@ namespace CefSharp
     private class CefAppUnmanagedWrapper : CefApp, CefRenderProcessHandler
     {
     private:
+        //Simple v8context enter/exit to avoid code duplication
+        struct V8ContextScope
+        {
+            V8ContextScope(CefRefPtr<CefV8Context> context)
```
amaitland: Prefer classes/structs to be in their own files.

CefSharp.Core/Internals/JavascriptCallbackImplFactory.h
```diff
@@ -0,0 +1,31 @@
...
+
+#pragma once
+
+using namespace System;
+
+namespace CefSharp
+{
+    namespace Internals
+    {
+        ref class CefSharpBrowserWrapper;
```
amaitland: Pass in the `WeakReference` directly to avoid this. May as well just use `BrowserWrapper` as a `get/set` property.

CefSharp.Core/Internals/JavascriptCallbackImpl.cpp
```diff
@@ -0,0 +1,95 @@
...
+        {
+            CefSharpBrowserWrapper^ result = nullptr;
+            if (_browserWrapper->IsAlive)
+            {
+                result = static_cast<CefSharpBrowserWrapper^>(_browserWrapper->Target);
+            }
+            return result;
+        }
+
+        void JavascriptCallbackImpl::DisposedGuard()
```
amaitland: Looks unused, is this required?

CefSharp.Core/Internals/ClientAdapter.cpp
```diff
@@ -836,7 +836,7 @@ namespace CefSharp
                     {
                         response->Result = DeserializeV8Object(argList, 2, callbackFactory);
                     }
-                    else
+                     else
```
amaitland: Remove stray space

CefSharp.BrowserSubprocess.Core/CefAppUnmanagedWrapper.cpp
```diff
@@ -91,64 +91,105 @@ namespace CefSharp
...
+
+        auto browserId = argList->GetInt(0);
+        if (browser->GetIdentifier() != browserId)
         {
-            auto argList = message->GetArgumentList();
-            auto browserId = argList->GetInt(0);
-            auto frameId = GetInt64(argList, 1);
-            auto callbackId = GetInt64(argList, 2);
-            auto script = argList->GetString(3);
+            throw gcnew InvalidOperationException(String::Format("Request BrowserId : {0} does not match browser Id : {1}", browserId, browser->GetIdentifier()));
```
amaitland: Will this crash the `BrowserProcess`? I think in general we're better to return an error response than than an a `Exception`

CefSharp.BrowserSubprocess.Core/CefAppUnmanagedWrapper.cpp
```diff
@@ -91,64 +91,105 @@ namespace CefSharp
...
+                callbackId = GetInt64(argList, 2);
+                auto script = argList->GetString(3);
 
-                if (context.get() && context->Enter())
+                auto frame = browser->GetFrame(frameId);
+                if (frame.get())
                 {
-                    try
+                    auto context = frame->GetV8Context();
+                    V8ContextScope scope(context);
```
amaitland: I think a big part of the `try...finally` was actually the `try`, which those takes away. If there's an exception then the process will likely crash. Also the check for `if(context->Enter())` is gone, so I'd prefer the small amount of code repartition in this case.

CefSharp.BrowserSubprocess.Core/CefAppUnmanagedWrapper.cpp
```diff
@@ -91,64 +91,105 @@ namespace CefSharp
...
+
+                    success = context->Eval(script, result, exception);
+                    response = CefProcessMessage::Create(kEvaluateJavascriptResponse);
```
amaitland: Should always create the message I think, that way if there's an error we can still respond. It looks no response would be sent if a problem occurred.

CefSharp.BrowserSubprocess.Core/CefAppUnmanagedWrapper.cpp
```diff
@@ -91,64 +91,105 @@ namespace CefSharp
...
+                    params.push_back(DeserializeV8Object(parameterList, static_cast<int>(i)));
+                }
+
+                auto callbackRegistry = browserWrapper->CallbackRegistry;
+                auto callbackWrapper = callbackRegistry->FindWrapper(jsCallbackId);
+                auto context = callbackWrapper->GetContext();
+                auto value = callbackWrapper->GetValue();
+
+                {
+                    V8ContextScope scope(context);
```
amaitland: Same goes with this `V8ContextScope` - prefer `try..finally`

CefSharp.BrowserSubprocess.Core/CefAppUnmanagedWrapper.cpp
```diff
@@ -91,64 +91,105 @@ namespace CefSharp
...
+                    if (success)
+                    {
+                        auto argList = response->GetArgumentList();
+                        SerializeV8Object(result, argList, 2, browserWrapper->CallbackRegistry);
                     }
-                    finally
+                    else
                     {
-                        context->Exit();
+                        exception = value->GetException();
```
amaitland: Rather than store the exception it might be better to store a `CefString`, then set that if an exception occurs. Cases when the matching frame wasn't found etc could also be dealt with.

CefSharp.BrowserSubprocess.Core/JavascriptCallbackRegistry.cpp
```diff
@@ -48,6 +48,13 @@ namespace CefSharp
             return nullptr;
         }
 
+        JavascriptCallbackWrapper^ JavascriptCallbackRegistry::FindWrapper(int64 id)
```
amaitland: Can the `Execute` method be removed from this class now?

CefSharp.BrowserSubprocess.Core/JavascriptCallbackWrapper.cpp
```diff
@@ -5,11 +5,14 @@
 #include "stdafx.h"
 #include "TypeUtils.h"
 #include "JavascriptCallbackWrapper.h"
+#include "Serialization/V8Serialization.h"
 
 namespace CefSharp
 {
     namespace Internals
     {
+        using namespace Serialization;
```
amaitland: Prefer `using` statements to follow directly after `include`

CefSharp.Core/Internals/JavascriptCallbackImpl.cpp
```diff
@@ -0,0 +1,95 @@
...
+#include "JavascriptCallbackImpl.h"
+#include "Messaging/Messages.h"
+#include "Serialization/Primitives.h"
+#include "Serialization/V8Serialization.h"
+
+namespace CefSharp
+{
+    namespace Internals
+    {
+        using namespace Messaging;
```
amaitland: Move usings to directly after includes

CefSharp.Core/Internals/JavascriptCallbackImpl.cpp
```diff
@@ -0,0 +1,95 @@
...
+        using namespace Serialization;
+
+        JavascriptCallbackImpl::JavascriptCallbackImpl(JavascriptCallback^ callback, PendingTaskRepository<JavascriptResponse^>^ pendingTasks, WeakReference^ browserWrapper)
```
amaitland: Going forward, I think we're basically adopting the convention that `Constructors/Destructors/Finalizers` all go in the header file.

amaitland: There are a number of places where this isn't the case. I'll clean this up later.

CefSharp.Core/Internals/JavascriptCallbackImplFactory.h
```diff
@@ -0,0 +1,31 @@
...
+
+        ref class JavascriptCallbackImplFactory : public IJavascriptCallbackFactory
```
amaitland: `Impl` in the name isn't consistent with the rest of the codebase. I know that causes some conflicts, so I think a name reshuffle is in order.

amaitland: Also prefer to include explicit `accessor` deceleration, so either `private` or `public` before `ref class`.

amaitland:
@arsher In general, great progress :+1: Comments inline. Feel free to discuss.

CefSharp.BrowserSubprocess.Core/JavascriptCallbackRegistry.cpp
arsher: Yep, I missed this

CefSharp.Core/Internals/JavascriptCallbackImpl.cpp
arsher: Ok, got it. Is it a good idea though? Wouldn't it be cleaner to have all implementation in the cpp files?

CefSharp.BrowserSubprocess.Core/CefAppUnmanagedWrapper.cpp
arsher: You're right with the try finally, I was just trying to minimize code duplication. I think maybe I should reintroduce msg handler classes to make the message handling more manageable, what do you think?

CefSharp.Core/Internals/JavascriptCallbackImpl.cpp
amaitland: Think it comes down to consistency and preference. Obviously it's not strictly required to even have a `.cpp` file. For super simply classes I'd probably be in favor of doing away with the `.cpp` files.

CefSharp.BrowserSubprocess.Core/CefAppUnmanagedWrapper.cpp
amaitland: Let's get this squared away first then take a step back, look at how the code can be best structured. I'm all for adding a class in general.

CefSharp.BrowserSubprocess.Core/CefAppUnmanagedWrapper.cpp
arsher: I believe if this happens that's a very bad sign, I'm not sure what we should do in this case.

amaitland: Any reason we can't just send a response message containing an error?

arsher: We can do that, but what would that contain? I mean should we put the callbackid into it, so  we can fail the task on the other side? If so what happens for a message when there is no callback (like destroy callback or messages yet to add)? Should we have multiple different error messages? Or should there be a unique correlation id kind of thing in each and every message so we can report any kind of error? So my only concern is that it raises a set of problems in a context which shouldn't really happen. Anyway I'll try to cook up something tonight, let's see how that goes :)

amaitland: So I made a minor change which makes things a little more explicit. I think some of it still needs a little refactoring.

https://github.com/amaitland/CefSharp/commit/d766cc826e8caecc025a3db8d48d2bd49c3c305f#diff-2731f23ba4e8fcf96f8501a49c6d687eL865

Basically the call to `SendProcessMessage` should happen on the browser instance that it was intended for, currently it's executing on the main browser process. When can then remove the `browserId` param.

https://github.com/amaitland/CefSharp/commit/d766cc826e8caecc025a3db8d48d2bd49c3c305f#diff-a27ca20dcc4f76e2dde0c785dd2abb9cR104
For the the `kJavascriptCallbackDestroyRequest` message, if it can't find the browser wrapper then It's pretty much safe to ignore as it's pretty much guaranteed to have been disposed already (unless there's a memory leak somewhere else).

Some of it's still a little rough, so I think it's time to revisit the whole structure. I've been thinking how we can move a large portion of the code into `c#` land with a few simple wrappers, so I'll toy with that idea a little more.

arsher: I like the idea. Can you create a PR against my branch?

arsher:
I've implemented something similar to what @amaitland suggested for missing browserwrapper on the render process side

amaitland:
You could have just cherry picked my commit or merged the changes from my branch :smile:

arsher:
Oh, I didn't know that was possible between forks :smile: well...next time

amaitland:
You can do it from the command line without add another source, something like `git pull https://github.com/amaitland/CefSharp.git feature/replacewcf2` would have done the trick :smile:

amaitland:
Merging this now. Will add some followup changes also.

Hopefully in the next few days I'll have something to show in terms of moving some code into `c#` land.

PR #1135, opened 2015-07-21
Add basic RequestContext Implementation

amaitland:
Ability to set your own request context, browsers can share a custom set of settings

TODO create a wrapper around CefRequestContextSettings

### DIFF
--- a/CefSharp.Core/BrowserSettings.h
+++ b/CefSharp.Core/BrowserSettings.h
@@ -29,7 +29,6 @@ namespace CefSharp
         !BrowserSettings()
         {
             delete _browserSettings;
-            delete RequestContext;
         }
 
         ~BrowserSettings()
@@ -331,7 +330,5 @@ namespace CefSharp
         }		
 
         property Nullable<bool> OffScreenTransparentBackground;
-
-        property RequestContext^  RequestContext;
     };
 }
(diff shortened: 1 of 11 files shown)


amaitland:
@etomm Try this - lots of changes with the new version

etomm:
Thanks, it seems much similar to what I was trying to achieve!... I will try asap and report back to you

amaitland:
Thinking about this further, should also split `RequestContext` out from `BrowserSettings`.

etomm:
Ok, I tested it... the offscreen example works. I hope it is not just a glitch of the soft multithread used there. In our solution we are having an high multithread environment and we were having problem of zoom being resetted from one thread in the others.

About splitting out RequestContext from the BrowserSettings, I would agree with you because you can use the same BrowserSettings with different RequestContext according to the cef method. So it would be good to split them. I was just not wanting to change the code too much.

Do you have an idea when this will be released in a prerelease nuget?

amaitland:
> Do you have an idea when this will be released in a prerelease nuget?

There is quite a lot of work to be done before a `-pre` is ready unfortunately.

amaitland:
I've split `RequestContext` out and added a wrapper for `RequestContextSettings`. Basics should be working, `merging` now.

PR #1136, opened 2015-07-22
ManagedCefBrowserAdapter can now be disposed before initialization

jankurianski:
This fixes a NullReferenceException when disposing a ChromiumWebBrowser prior to initialization.

To reproduce the bug, modify `CefSharp.WinForms.Example\BrowserTabUserControl.cs` to the following:

```
line 21
            browserPanel.Controls.Add(browser);
+           browserPanel.Visible = false;
```

Then start the `CefSharp.WinForms.Example` application and close the main window.  Because the `_browserWrapper` was never initialized, a `NullReferenceException` would occur on the line `_browserWrapper->CloseBrowser(true);`.


amaitland:
Nice :+1:

Issue #570, opened 2014-11-04
Clicking on Control using Winforms does not Set Focus


RadicalLinux:
When using docking tabs menus etc. that should automatically close when they no longer have focus, clicking on the browser does not fire the event needed.  You can duplicate the issue by running the winforms example. After the browser is fully loaded click on the edit menu.  Now click back on the browser.  The edit menu will not close like it should.  Also the mouse events are not triggered.

BTW All the other issues have been resolved!  Great Work!

jankurianski:
@RadicalLinux Not sure if it helps you any more, but I found a hack to make this work for my case.  If you paste this into `BrowserTabUserControl.cs` in the WinForms example project then the `ToolStrip` will get dismissed when you click into the browser:

https://gist.github.com/jankurianski/2163d7928d3d6f839e47

RadicalLinux:
Thanks, I'm giving it a try now.

asdorius:
Hey guys, thanks for the great workaround. Almost got it working for me.. almost. My problem is that the workaround is causing Chromium to lose focus for a short time. This destroys functionality in my app because as a graphics editor it relies heavily on mouse event handling.Any ideas on how to fix this?

jankurianski:
You could try posting a different message type instead of WM_LBUTTONDOWN? See here for the messages that will cause the ToolStrip to close:
http://referencesource.microsoft.com/#System.Windows.Forms/winforms/Managed/System/WinForms/ToolStripManager.cs,1249

asdorius:
This is it! Thank you so much. When posting WM_NCLBUTTONDOWN (0x00A1), cef doesn't lose focus and handles the following mouse click correctly.

amaitland:
@jankurianski Is it worth adding your `Gist` example to the `WinForms.Example` project? Seems like there's enough interest.

jankurianski:
> This is it! Thank you so much. When posting WM_NCLBUTTONDOWN (0x00A1), cef doesn't lose focus and handles the following mouse click correctly.

@asdorius No, thank you.  :+1:   I had a bug in my app when running on VMs where the `mouseup` never got sent to CefSharp, and `WM_LBUTTONDOWN` was the cause of the problem.

I've updated my `Gist` and the `WinForms.Example` PR to use `WM_NCLBUTTONDOWN` now.

jankurianski:
> @jankurianski Is it worth adding your Gist example to the WinForms.Example project? Seems like there's enough interest.

[...]

Issue #1098, opened 2015-06-30
MouseEnter and MouseLeave events of ChromiumWebBrowser control do not fire


sarus:
Hello,

Not sure if I should be expecting this to work but the MouseLeave and MouseEnter events don't seem to work when using the ChromiumWebBrowser control for WinForms.  Here's the code I was using.  Should I be expecting these events to trigger properly on the browser control?  Thank you!

```C#
        public WebView()
        {
            InitializeComponent();
            browser = new ChromiumWebBrowser("resource://web/index.html")
            {
                Dock = DockStyle.Fill,
            };

            mainPanel.Controls.Add(browser);

            browser.MouseLeave += new System.EventHandler(this.browser_MouseLeave);
            browser.MouseEnter += new System.EventHandler(this.browser_MouseEnter);           
        }

        void browser_MouseLeave(object sender, EventArgs e)
        {
            // Never Fires
            Console.WriteLine("Browser Mouse ENter");              
        }

        void browser_MouseEnter(object sender, EventArgs e)
        {
            // Never Fires
            Console.WriteLine("Browser Mouse Enter");
        }
```

I'm using the latest CefSharp available from Nuget (41.0.0)

jankurianski:
The underlying CEF library captures all mouse events and does not bubble them up to WinForms.  There's no way around this for WinForms - if you want to handle mouse events you need to do so in JavaScript.

Your alternative is to use the `WPF` CefSharp control, as this uses the underlying CEF library's OSR (Off Screen Rendering) mode, in which `WPF` is in charge of all keyboard and mouse events, and it passes them onto CEF.  This would give you an opportunity to intercept whatever events you wanted.

sarus:
Ah interesting thank you very much.  Is the general recommendation to use the WPF control?  We're "porting" an existing application that is built on winforms which is why I started with the winforms control but it is possible for us to move to WPF if needed (although a bit painful).

amaitland:
You may be able to hook the underlying handle and capture the raw messages, e.g. `WM_MOUSELEAVE` using a `NativeWindow` implementation. Might be worth loading `Spy++` and seeing if there's a handle that receives the messages.

sarus:
[...]

PR #1138, opened 2015-07-22
Add ChromeWidgetMessageInterceptor to WinForms Example

jankurianski:
Addresses #570

### DIFF
--- a/CefSharp.WinForms.Example/BrowserTabUserControl.cs
+++ b/CefSharp.WinForms.Example/BrowserTabUserControl.cs
@@ -1,5 +1,6 @@
 ﻿using CefSharp.Example;
 using System;
+using System.Runtime.InteropServices;
 using System.Windows.Forms;
 using CefSharp.WinForms.Internals;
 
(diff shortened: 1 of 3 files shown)


jankurianski:
Anyone have corrections or thoughts before I merge this?

amaitland:
Only question I have is would it be better to explicitly release the handle?

amaitland:
If it's working then I'm happy to merge is as, few formatting things, they can be sorted later :+1:

jankurianski:
> Only question I have is would it be better to explicitly release the handle?

It is working without releasing the handle, but it may be inadvertently preventing garbage collection.  So I have made it explicitly release the handle now.

jankurianski:
> Add browser.HandleDestroyed -= BrowserHandleDestroyed;?

Good catch :+1:

amaitland:
Might be worth adding the extra example from https://github.com/cefsharp/CefSharp/issues/1098#issuecomment-118208651

Can be commented out, just as a user reference?

jankurianski:
> Might be worth adding the extra example from #1098 (comment)
> 
> Can be commented out, just as a user reference?

Good thinking.  I might actually change the `ChromeWidgetMessageInterceptor` to accept an `Action<Message>` delegate, then it can be up to the caller to do what they wish with the widget messages.

jankurianski:
I added a comment about mouse events.

I also moved moved the `ToolStrip`-closing logic into `BrowserTabUserControl`.  Any thoughts?

amaitland:
Looks fine :+1:

PR #1141, opened 2015-07-23
OnGotFocus no longer calls SetFocus if browser not initialized.

jankurianski:
Fixes following crash:
```
Exception: System.Exception
Message: 'Browser Is Not yet initialized. Use the IsBrowserInitializedChanged event and checkthe IsBrowserInitialized property to determine when the browser has been intialized.'
Source: CefSharp
Stacktrace:
at CefSharp.WebBrowserExtensions.ThrowExceptionIfBrowserNotInitialized(IWebBrowser browser)
at CefSharp.WinForms.ChromiumWebBrowser.SetFocus(Boolean isFocused)
at CefSharp.WinForms.ChromiumWebBrowser.OnGotFocus(EventArgs e)
at System.Windows.Forms.Control.WmSetFocus(Message& m)
at System.Windows.Forms.Control.WndProc(Message& m)
at System.Windows.Forms.NativeWindow.Callback(IntPtr hWnd, Int32 msg, IntPtr wparam, IntPtr lparam)
```

### DIFF
--- a/CefSharp.WinForms/ChromiumWebBrowser.cs
+++ b/CefSharp.WinForms/ChromiumWebBrowser.cs
@@ -324,7 +324,11 @@ namespace CefSharp.WinForms
 
         protected override void OnGotFocus(EventArgs e)
         {
-            SetFocus(true);
+            if (IsBrowserInitialized)
+            {
+                SetFocus(true);
+            }
+
             base.OnGotFocus(e);
         }
 


amaitland:
Merge away :+1: 

How are you finding `master` in general? Any comments on the `API` after using it for a bit?

jankurianski:
> How are you finding master in general? Any comments on the API after using it for a bit?

The API is much more understandable now.  I think making it more in line with CEF has paid off.

In terms of stability, I haven't got through all of my app's regression tests yet :smile: Will still need to post more bug fixes.

PR #1142, opened 2015-07-23
Ignore callback usage after dispose

jankurianski:
Users have no way to determine whether a callback has been disposed, resulting in the following crash if the callback is fired after disposal.  This change silently ignores the method calls.  Another approach would have been to add an `IsDisposed` flag that users must check but that seems more error-prone.

Because callbacks are typically executed asynchronously there is a good chance that they fire after the callback and/or the browser control has been disposed.  This happens occasionally in my app, and adding a `if (!browser.IsDisposed)` check prior to the callback being fired was not adequate to protect against the `NullReferenceException`.

I'm re-running my tests now to make sure this fixes it.

```
Exception: System.NullReferenceException
Message: 'Object reference not set to an instance of an object.'
Source: CefSharp.Core
Stacktrace:
at CefSharp.CefCallbackWrapper.Continue()
at ...
```

### DIFF
--- a/CefSharp.Core/Internals/CefCallbackWrapper.h
+++ b/CefSharp.Core/Internals/CefCallbackWrapper.h
@@ -33,6 +33,11 @@ namespace CefSharp
 
         virtual void Cancel()
         {
+            if (this == nullptr)
+            {
+                return;
+            }
+
             _callback->Cancel();
 
             delete this;
@@ -40,6 +45,11 @@ namespace CefSharp
 
         virtual void Continue()
         {
+            if (this == nullptr)
+            {
+                return;
+            }
+
             _callback->Continue();
 
             delete this;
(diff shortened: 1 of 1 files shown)


jankurianski:
Well, that didn't work.  My C++ understanding isn't very good.  I need to check whether `_callback` is null rather than `this` is null, however I can't figure out how to.

BTW to reproduce this in `CefSharpSchemeHandler.cs` change line `60` to:

```
                    callback.Dispose();
                    callback.Continue();
```

amaitland:
Try 

```c++
virtual void Continue()
{
	if(_callback.get())
	{
		_callback->Continue();

		delete this;
	}
}
```

jankurianski:
Thanks @amaitland, appreciate your help :+1:

amaitland:
No probs, let me know how it goes :+1:

jankurianski:
My regression test that found this crash now passes with the fix, so I'm happy.  Anything further before merging?

amaitland:
I've added #1143, so we can make a note to update the other `callbacks`.

Issue #915, opened 2015-03-30
Unable to execute javascript in non administrator while other WCF service is running 


Antonyo:
Hello,

some of our users have experience an issue that their WPF CEF browser wasn't responding, after some research we finally found a user that give us the clue, he was running voodooshield service which is a WCF service running as window service, he stated that in order to get it to work he need to either stop the voodooshield service or run our app as administrator.

I installed `https://voodooshield.com/`and I get to recreate it in Windows 7, W8.1 and W10 x64. Our WPF cef control wasn't working because we require some javascript to execute at initialization which was not being executed at all. 

I got to recreate it in your cefsharp solution examples, in the WPF example and the winform example. If you got the voodooshield service running (you don't need to enable voodooshield app actually, only get his service running) you will see that the `bound` object has not been register and you cannot execute javascript asynchronously. The bound object I refer is the one register in:

```CefSharp.Wpf.Example.Views.BrowserTabView.BrowserTabView line 19```

After some debugging (you need to run VS as not administrator to debug it) I think there must be some problem in the `BrowserProcessServiceHost`, I'm not sure but seems to be a problem with the named pipe, at least I noticed that the `BrowserProcessService`is never instantiated. Also the `BrowserProcessServiceHost.EvaluateScriptAsync` method also doesn't work because the `operationContextTask`never seems to terminate and the `ContinueWith` block is not executed.

The voodooshield team tell the user that they thought that this issue will occur while any generic WCF/Windows service is running but I've created a WCF running as windows service myself and there was no problem at all so we've contacted voodooshield to see what could be happening.

Thanks for the help

**Comment by amaitland** [comment]
As them if they specify a absolute address when creating their `WCF` service e.g. `net.pipe://localhost/CefSharpSubProcessProxy/6688/1`

You might find this useful http://stackoverflow.com/questions/10362246/two-unique-named-pipes-conflicting-and-invalidcredentialexception

**Comment by mol** [comment]
I talked to them and this is how the endpoint is defined in the config file:

```
<system.serviceModel>
    <bindings>
      <netNamedPipeBinding>
        <binding name="VoodooShieldServiceNetNamedPipeEndPoint" />
      </netNamedPipeBinding>
    </bindings>
    <client>
      <endpoint address="net.pipe://vswcfpipe/" binding="netNamedPipeBinding"
        bindingConfiguration="VoodooShieldServiceNetNamedPipeEndPoint"
[...]

PR #1147, opened 2015-07-27
Async method bindings using CEF IPC

cefsharp-ms:
This branch contains an async method binding implementation which doesn't require the usage of WCF. The bound methods always return promises when called from javascript so can be integrated into any code that uses the builtin Promise objects in chromium.

Please note that the additional license info has been added upon request from our legal department and would be great if you'd consider accepting the pull request in it's current form.


amaitland:
@cefsharp-ms Welcome! Very interesting contribution, thanks!

Can you please update to include some examples of the new code in action. Simple additions to `BindingTest.html` should be fine.

CefSharp/Internals/MethodInvocationCompleteArgs.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
```
amaitland: Missing license disclaimer, please add (there's other cases as well)

CefSharp/IWebBrowser.cs
```diff
@@ -59,6 +59,8 @@ public interface IWebBrowser : IDisposable
         /// <param name="camelCaseJavascriptNames">camel case the javascript names of properties/methods, defaults to true</param>
         void RegisterJsObject(string name, object objectToBind, bool camelCaseJavascriptNames = true);
 
+        void RegisterAsyncJsObject(string name, object objectToBind, bool camelCaseJavascriptNames = true);
```
amaitland: Need to add comment including limitations.

CefSharp.Core/Internals/Messaging/Messages.h
```diff
@@ -24,6 +24,12 @@ namespace CefSharp
             const CefString kJavascriptCallbackDestroyRequest = "JavascriptCallbackDestroyRequest";
             //Message containing the result of a given js function call
             const CefString kJavascriptCallbackResponse = "JavascriptCallbackDoneResponse";
+            //Message containing a js root object for async bindings
+            const CefString kJavascriptRootObjectRequest = "JavascriptRootObjectRequest";
```
amaitland: Should have async in the name.

Likely we'll port standard JSB over to using native IPC.

CefSharp.Core/Internals/ClientAdapter.cpp
```diff
@@ -847,6 +854,26 @@ namespace CefSharp
 
                 handled = true;
             }
+            else if (name == kJavascriptMethodCallRequest)
+            {
+                if (!browser->IsPopup())
```
amaitland: We should look at supporting `popups`

CefSharp.BrowserSubprocess.Core/CefAppUnmanagedWrapper.cpp
```diff
@@ -261,7 +269,38 @@ namespace CefSharp
...
+                //dispose
+                delete callback;
+            }
+            handled = true;
+        }
 
         return handled;
     };
+
+    DECL void CefAppUnmanagedWrapper::OnWebKitInitialized()
```
amaitland: Remove `DECL`

CefSharp.BrowserSubprocess.Core/Async/JavascriptAsyncMethodWrapper.h
```diff
@@ -0,0 +1,35 @@
...
+// Use of this source code is governed by a BSD-style license that can be found in the LICENSE file.
+
+#pragma once
+
+#include "JavascriptAsyncMethodHandler.h"
+#include "JavascriptCallbackRegistry.h"
+
+namespace CefSharp
+{
+    ref class CefBrowserWrapper;
```
amaitland: Is this required? I don't see any reference to `CefBrowserWrapper` in this class

CefSharp.BrowserSubprocess.Core/Async/JavascriptAsyncMethodHandler.h
```diff
@@ -0,0 +1,40 @@
...
+
+#pragma once
+
+#include "include/cef_v8.h"
+#include "JavascriptCallbackRegistry.h"
+#include "JavascriptAsyncMethodCallback.h"
+
+namespace CefSharp
+{
+    ref class CefBrowserWrapper;
```
amaitland: Is this required?

CefSharp.BrowserSubprocess.Core/Async/JavascriptAsyncMethodHandler.cpp
```diff
@@ -0,0 +1,52 @@
...
+                    SerializeV8Object(arguments[i], params, i, _callbackRegistry);
+                }
+
+                SetInt64(_objectId, argList, 0);
+                SetInt64(callbackId, argList, 1);
+                argList->SetString(2, name);
+                argList->SetList(3, params);
+
+                browser->SendProcessMessage(CefProcessId::PID_BROWSER, request);
+                return true;
```
amaitland: Is there a case where this should `return false`? Probably no point returning anything if it's always true.

CefSharp.BrowserSubprocess.Core/Async/JavascriptAsyncMethodHandler.cpp
```diff
@@ -0,0 +1,52 @@
+// Copyright � 2010-2015 The CefSharp Project. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style license that can be found in the LICENSE file.
+
+#include "stdafx.h"
+#include "JavascriptAsyncMethodHandler.h"
+#include "../CefSharp.Core/Internals/Messaging/Messages.h"
+#include "../CefSharp.Core/Internals/Serialization/Primitives.h"
+#include "Serialization/V8Serialization.h"
+#include "CefBrowserWrapper.h"
```
amaitland: Is this required?

CefSharp.BrowserSubprocess.Core/Async/JavascriptAsyncObjectWrapper.cpp
```diff
@@ -0,0 +1,35 @@
...
+namespace CefSharp
+{
+    namespace Internals
+    {
+        namespace Async
+        {
+            void JavascriptAsyncObjectWrapper::Bind(const CefRefPtr<CefV8Value> &value)
+            {
+                //V8Value that represents this javascript object - only one per complex type, no accessor
+                auto javascriptObject = value->CreateObject(nullptr);
```
amaitland: From memory `CreateObject` is a static method, so prefer to have it called as such.

CefSharp.BrowserSubprocess.Core/CefAppUnmanagedWrapper.cpp
```diff
@@ -98,7 +106,7 @@ namespace CefSharp
         //Error handling for missing/closed browser
         if (browserWrapper == nullptr)
         {
-            if (name == kJavascriptCallbackDestroyRequest)
+            if (name == kJavascriptCallbackDestroyRequest || name == kJavascriptRootObjectRequest || name == kJavascriptMethodCallResponse)
```
amaitland: Should probably wrap this as it's getting a little long.

CefSharp.BrowserSubprocess.Core/CefAppUnmanagedWrapper.cpp
```diff
@@ -261,7 +269,38 @@ namespace CefSharp
...
+            handled = true;
+        }
 
         return handled;
     };
+
+    DECL void CefAppUnmanagedWrapper::OnWebKitInitialized()
+    {
+        //we need to do this because the builtin Promise object is not accesible
+        CefRegisterExtension("promisecreator", "function cefsharp_CreatePromise() {var object = {};var promise = new Promise(function(resolve, reject) {object.resolve = resolve;object.reject = reject;});return{ p: promise, res : object.resolve,  rej: object.reject};}", nullptr);
```
amaitland: Wrap this.

CefSharp.BrowserSubprocess.Core/CefAppUnmanagedWrapper.h
```diff
@@ -10,6 +10,7 @@
 #include "include/cef_base.h"
 
 #include "CefBrowserWrapper.h"
+#include "Async/JavascriptAsyncMethodCallback.h"
```
amaitland: Can probably move this into the `cpp` file

CefSharp.BrowserSubprocess.Core/CefBrowserWrapper.h
```diff
@@ -35,6 +43,8 @@ namespace CefSharp
             CefSharp::Internals::JavascriptCallbackRegistry^ get();
         }
 
+        void SendProcessMessage(CefProcessId target_process, CefRefPtr<CefProcessMessage> message);
```
amaitland: Can likely remove the `target_process` param and just hard code it as it's pretty much guaranteed to always be sending messages to the `Browser` process. If you feel strongly enough to keep it then `target_process` should be renamed to `targetProcess`.

CefSharp/Internals/MethodRunnerQueue.cs
```diff
@@ -0,0 +1,117 @@
...
+        }
+
+        private void ConsumeTasks()
+        {
+            try
+            {
+                while (!cancellationTokenSource.IsCancellationRequested)
+                {
+                    var task = queue.Take(cancellationTokenSource.Token);
+                    task.RunSynchronously(TaskScheduler.Current);
```
amaitland: What thread does this end up being executed on? `ThreadPool`?

amaitland:
Comments inline, look forward to your response :+1:

amaitland:
Out of curiosity what was the motivation for implementing this?

cefsharp-ms:
I'll get on the modification according to the comments.
Also about the motivation, we had a lot of elusive errors with the wcf based bindings. And as CEF doesn't include the sync IPC mechanism of chromium I went with the async approach. Which in the end is actually the better solution in js as it won't block the render thread.

CefSharp/Internals/MethodRunnerQueue.cs
cefsharp-ms: Oh I missed that, I should've used the overload without a TaskScheduler parameter. So it will run on the consumer thread as expected.

amaitland:
> I'll get on the modification according to the comments.

Mostly trivial stuff. Let me know if you have any thoughts.

If you can merge in `master` that would be great, means we can `auto merge` using `GitHub` :smile:

CefSharp.BrowserSubprocess.Core/Async/JavascriptAsyncObjectWrapper.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+using namespace System::Linq;
+
+namespace CefSharp
+{
+    namespace Internals
+    {
+        namespace Async
+        {
+            void JavascriptAsyncObjectWrapper::Bind(const CefRefPtr<CefV8Value> &value)
```
amaitland: I think passing in the `CefV8Value` is better, originally there was a need to store a reference, don't think it's required anymore.

So I'm thinking we go with that for all the `Bind` methods?

I updated `master` accordingly yesterday https://github.com/cefsharp/CefSharp/commit/3d19b3a587e364b1501bfc3423185fc132047a79

mol:
@amaitland quick question(s). Is this a straight up switching from WCF to using this? I just happened upon this pull request.

We're seeing more and more users with the WCF named pipes issue, where it doesn't work unless the process is elevated, because of another elevated program in the system using WCF named pipes with a relative path. Now a program from HP is causing it as well. This issue:
https://github.com/cefsharp/CefSharp/issues/915
(We've actually developed a tiny program that will cause this issue if anyone would like to see it in action. If run, CEF won't work :) Source code included - just let me know).

I also happened upon this:
https://github.com/cefsharp/CefSharp/pull/1088
Is that a similar change to this, or?

FYI @Antonyo.

amaitland:
> @amaitland quick question(s). Is this a straight up switching from WCF to using this? I just happened upon this pull request.

Only supports `functions`, so you'd have to rewrite your `API` using `Get/Set` functions specifically.

> I also happened upon this:
> #1088
> Is that a similar change to this, or?

Most of that code has already been merged into `master`, it ports `EvaluateScriptAsync` to using native `IPC`.

amaitland:
@cefsharp-ms Formatting went astray when merging in `master` it would seem. (Spaces not tabs from what I can tell).

You've also changed the life cycle of `JavascriptCallbackRegistry` to corresponding to that of the `CefBrowserWrapper`. Subtle change, though I'd deliberately moved it into into the `JavascriptRootObjectWrapper`, I think this is important as a new instances is created and destroyed with the `Context`, which can happen many times for the lifespan of a `CefBrowserWrapper`. I think the simplest thing is to have a unique instance of `JavascriptCallbackRegistry` in both `JavascriptRootObjectWrapper` and `JavascriptAsyncRootObjectWrapper` (I would imagine most people would use one or the other typically and not both).

I tried to load `BindingTest.html` and got errors and the sub process crashed. Just an FYI

amaitland:
I stuffed up, see #1161

PR #1160, opened 2015-07-29
Exposed cookie manager wrapper

arsher:


CefSharp.Core/Internals/CookieManager.cpp
```diff
@@ -0,0 +1,105 @@
...
+            ThrowIfDisposed();
+
+            return _cookieManager->SetStoragePath(StringUtils::ToNative(path), persistSessionCookies, NULL);
+        }
+
+        void CookieManager::SetSupportedSchemes(... array<String^>^ schemes)
+        {
+            ThrowIfDisposed();
+
+            vector<CefString> schemeVector;
```
amaitland: May as well use existing `StringUtils::ToNative` here, (pass in a `List` instead).

https://github.com/cefsharp/CefSharp/blob/master/CefSharp.Core/Internals/StringUtils.cpp#L53

CefSharp.Core/Internals/CookieManager.cpp
```diff
@@ -0,0 +1,105 @@
...
+    {
+        void CookieManager::ThrowIfDisposed()
+        {
+            if (_cookieManager.get() == nullptr)
+            {
+                throw gcnew ObjectDisposedException("CookieManager");
+            }
+        }
+
+        bool CookieManager::DeleteCookies(String^ url, String^ name)
```
amaitland: What do you think about renaming this to `DeleteCookiesAsync` and returning the actual task here?

Same for `SetCookie`

CefSharp.Core/Internals/CookieManager.cpp
amaitland: I think that means removing the `params` option, so something to ponder....

arsher: Or StringUtils could be modified to accept IEnumerable instead of List

CefSharp.Core/Internals/CookieManager.cpp
arsher: probably a good idea

CefSharp.Core/Internals/CookieManager.cpp
amaitland: Nice :+1:

CefSharp.Core/Internals/CookieManager.cpp
```diff
@@ -0,0 +1,101 @@
...
+            if (CefCurrentlyOn(TID_IO))
+            {
+                auto source = gcnew TaskCompletionSource<bool>();
+                source->TrySetResult(cookieInvoker->DeleteCookies());
+                return source->Task;
+            }
+
+            auto task = Cef::IOThreadTaskFactory->StartNew(gcnew Func<bool>(cookieInvoker, &CookieAsyncWrapper::DeleteCookies));
+
+            delete cookieInvoker;
```
amaitland: Is the `CookieAsyncWrapper` disposed of before the task completes?

Might be safer for `CookieAsyncWrapper` to release the `CookieManager` reference once it's made the call, then return the result. Thoughts?

amaitland:
I made an additional change, just simplified the dipose, adding the extra threading code seemed overly complex

https://github.com/cefsharp/CefSharp/commit/6d6f095bebf3bd959deec7f6981e76c45b12a085

Let me know if you think I missed anything :+1:

PR #1161, opened 2015-07-30
Async method bindings using CEF IPC

amaitland:
I accidentally merged #1147 into `master` when I took it for a spin locally. Reverted changes and recreating `PR`.


CefSharp.BrowserSubprocess.Core/Async/JavascriptAsyncMethodHandler.h
```diff
@@ -0,0 +1,38 @@
...
+            class JavascriptAsyncMethodHandler : public virtual CefV8Handler
+            {
+            private:
+                gcroot<JavascriptCallbackRegistry^> _callbackRegistry;
+                gcroot<Func<JavascriptAsyncMethodCallback^, int64>^> _methodCallbackSave;
+                CefRefPtr<CefV8Value> _promiseCreator;
+                int64 _objectId;
+
+            public:
+                JavascriptAsyncMethodHandler(int64 objectId, JavascriptCallbackRegistry^ callbackRegistry, CefRefPtr<CefV8Value> promiseCreator, Func<JavascriptAsyncMethodCallback^, int64>^ methodCallbackSave)
```
amaitland: Missing `destructor` to cleanup resources

cefsharp-ms: I don't think it's necessary, both the gcroots and CefRefPtr will take care of it in their destructors. and callback registry shouldn't be disposed here.

amaitland: Part consistency, part pedantic. Prefer to have one :smile:

amaitland:
Copying https://github.com/cefsharp/CefSharp/pull/1147#issuecomment-126145515 to continue conversation

@cefsharp-ms Formatting went astray when merging in `master` it would seem. (Spaces not tabs from what I can tell).

You've also changed the life cycle of `JavascriptCallbackRegistry` to corresponding to that of the `CefBrowserWrapper`. Subtle change, though I'd deliberately moved it into into the `JavascriptRootObjectWrapper`, I think this is important as a new instances is created and destroyed with the `Context`, which can happen many times for the lifespan of a `CefBrowserWrapper`. I think the simplest thing is to have a unique instance of `JavascriptCallbackRegistry` in both `JavascriptRootObjectWrapper` and `JavascriptAsyncRootObjectWrapper` (I would imagine most people would use one or the other typically and not both).

I tried to load `BindingTest.html` and got errors and the sub process crashed. Just an FYI

CefSharp.BrowserSubprocess.Core/Async/JavascriptAsyncMethodHandler.h
cefsharp-ms: Sure, no problem, I'll add one :smile:

cefsharp-ms:
I'll check the formatting.

> You've also changed the life cycle of JavascriptCallbackRegistry to corresponding to that of the CefBrowserWrapper. Subtle change, though I'd deliberately moved it into into the JavascriptRootObjectWrapper, I think this is important as a new instances is created and destroyed with the Context, which can happen many times for the lifespan of a CefBrowserWrapper. I think the simplest thing is to have a unique instance of JavascriptCallbackRegistry in both JavascriptRootObjectWrapper and JavascriptAsyncRootObjectWrapper (I would imagine most people would use one or the other typically and not both).

The reason I did this is that we assume the callbackid to be unique for a browser instance so having two separate callbackregistries would cause troubles. I agree the lifecycle of the registry should be connected to the context but this way it's not going to work. 
What if I create another class to wrap the callbackregistry and the two rootobjectwrappers?

amaitland:
> What if I create another class to wrap the callbackregistry and the two rootobjectwrappers?

A simpler approach might just be to provide the initial id value when creating each instance? Like one starts at `0`, the other at `50000`. End of the day long as It's clean, readable I don't mind what solution is used.

amaitland:
Actually, may just combine `JavascriptAsyncRootObjectWrapper` and `JavascriptRootObjectWrapper`? They have the same lifecycle, rename a few variables internally to make them unique. Thoughts?

CefSharp.BrowserSubprocess.Core/Async/JavascriptAsyncMethodHandler.h
amaitland: Thanks :+1:

cefsharp-ms:
I modified JavascriptRootObjectWrapper to include async bindings as well and CallbackRegistry lifecycle is managed by that too.

cefsharp-ms:
I'm sorry for the earlier problems around the license file... :disappointed: it was a misunderstanding, I removed the commit and also wrapped up the whole PR a bit

amaitland:
Fantastic thanks :+1: 

Will take for one final spin when I'm in front of VS, will merge asap :smile:

CefSharp/Internals/MethodRunnerQueue.cs
```diff
@@ -0,0 +1,121 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace CefSharp.Internals
+{
+    public sealed class MethodRunnerQueue
+    {
+        private readonly JavascriptObjectRepository repository;
+        private readonly AutoResetEvent stopped = new AutoResetEvent(false);
+        private readonly BlockingCollection<Task<MethodInvocationResult>> queue = new BlockingCollection<Task<MethodInvocationResult>>();
```
amaitland: Just going over this one last time before `merge`. Is it worth simplifying this and just maintaining a collection of `MethodInvocation` objects, as we're basically executing them in order? That would simplify the code.

The sub tasks didn't seem to be getting a cancellation token, I may have missed something obvious as I've only stepped through this a couple of times.

amaitland:
@cefsharp-ms I've merged this (on purpose this time :smile: ), thanks for all your hard work :+1: 

At some point I'd like to review the `MethodRunnerQueue`, see if we can simplify or improve. Let me know if you have any followup thoughts?

PR #1164, opened 2015-08-01
Allowing extension registration

arsher:
Currently registering a handler for native function is not supported. I think it should probably work similarly to #1161.


CefSharp.BrowserSubprocess.Core/CefAppUnmanagedWrapper.h
```diff
@@ -22,12 +22,15 @@ namespace CefSharp
         gcroot<Action<CefBrowserWrapper^>^> _onBrowserCreated;
         gcroot<Action<CefBrowserWrapper^>^> _onBrowserDestroyed;
         gcroot<Dictionary<int, CefBrowserWrapper^>^> _browserWrappers;
+		gcroot<List<CefExtension^>^> _extensions;
```
amaitland: Formatting

CefSharp/CefExtension.cs
```diff
@@ -0,0 +1,50 @@
...
+        /// <summary>
+        /// Gets or sets the javascript extension code
+        /// </summary>
+        public string JavascriptCode { get; set; }
+
+        /// <summary>
+        /// Creates a new CwefExtension instance with a given name.
+        /// </summary>
+        /// <param name="name">Name of the CefExtension</param>
+        public CefExtension(string name)
```
amaitland: I think it probably makes sense to have `javascript` as a param of the constructor, and validate the same as `name`. Thoughts?

amaitland:
Other than a few white spaces issues and a slight change to the `CefExtension` class, looks solid :smile: Thanks :+1: 

Soon as #1161, this will be next.

arsher:
I've changed CefExtension so now it checks the js code value as well.
Also it might worth waiting for #1161 and implement the extension handler functionality based on that. What do you think? Or should I open another PR for that feature when I have it?

amaitland:
> What do you think? Or should I open another PR for that feature when I have it?

As it's a non trivial change, I think it's probably better to wait otherwise merging will become quite difficult. #1161 should be merged today or tomorrow.

amaitland:
Everything outstanding has been merged :smile: 

Any thoughts on merging the final pieces of your initial `IPC` branch? Once we've stripped out the `GetRegisteredJavascriptObjects` method, I'd like to refactor the code into a more maintainable structure.

arsher:
Cool :smile:
Actually #1161 implemented the last bits i wanted to add so I believe there is nothing left. Although I agree that it needs some refactoring, the message handling starting to be unmanagable.

amaitland:
#1161 Only transmits the objects registered using `RegisterAsync`, so there's still some code to be ported across from your original `PR`, should be pretty trivial.

> the message handling starting to be unmanagable.

Now that we have a clear picture the refactoring should be fairly trivial. I think the first step is probably to cleanup the indexing situation, that's likely to cause the most problems.

arsher:
Oh, you are right, i'll add that missing piece then :smile:

amaitland:
> Oh, you are right, i'll add that missing piece then :smile:

Awesome, thanks :+1: What do you think about sending both within the same `IPC` message?

PR #1168, opened 2015-08-04
Offscreen browser is not created for wpf control when it has parent

cefsharp-ms:
I believe the new offscreen popup feature had a little problem, namely that the offscreen browser is created for the wpf control even if it had a parent. This is just a minimal fix for that, maybe it shouldn't even be a PR :smile:.

### DIFF
--- a/CefSharp.Wpf/ChromiumWebBrowser.cs
+++ b/CefSharp.Wpf/ChromiumWebBrowser.cs
@@ -788,7 +788,8 @@ namespace CefSharp.Wpf
 
         private void CreateOffscreenBrowserWhenActualSizeChanged()
         {
-            if (browserCreated || System.ComponentModel.DesignerProperties.GetIsInDesignMode(this))
+            var webBrowserInternal = this as IWebBrowserInternal;
+            if (browserCreated || System.ComponentModel.DesignerProperties.GetIsInDesignMode(this) || webBrowserInternal.HasParent)
             {
                 return;
             }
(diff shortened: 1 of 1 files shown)


CefSharp.Wpf/ChromiumWebBrowser.cs
```diff
@@ -788,7 +788,8 @@ private void RemoveSourceHook()
 
         private void CreateOffscreenBrowserWhenActualSizeChanged()
         {
-            if (browserCreated || System.ComponentModel.DesignerProperties.GetIsInDesignMode(this))
+            var webBrowserInternal = this as IWebBrowserInternal;
```
amaitland: Rather than cast and check the `HasParent` param every time there is a resize, I think it's better to have a second `if` statement and set `browserCreated = true` if `HasParent = true`. Thoughts?

amaitland:
There's a scribbled note on my desk reminding me to fix this, so you've beaten me to it, thanks :smile:

CefSharp.Wpf/ChromiumWebBrowser.cs
cefsharp-ms: Probably a good idea.

amaitland: The whole `CreateOffscreenBrowserWhenActualSizeChanged` thing is because of a `CEF` bug, which we really should check to see it's been resolved.

PR #1173, opened 2015-08-06
All bound objects transmitted with CEF IPC

arsher:
Now the sync bound objects are going through cef ipc as well. I needed a bit more modifications than I first expected, but it's ready now :smile:


amaitland:
Nice one :+1:

PR #1175, opened 2015-08-06
GetRootScreenRect implementation

arsher:


amaitland:
So this resolves your problem?

arsher:
Yes, popups are working as expected after applying this.

amaitland:
Just need to check that it works when changing monitors with different resolutions.

amaitland:
> GetRootScreenRect represents the "screen" that contains the browser and is used for functionality that extends outside of the browser bounds (drag&drop, popup menus, etc).

For future reference
http://magpcss.org/ceforum/viewtopic.php?f=6&t=12835&p=25519&hilit=GetRootScreenRect#p25519

amaitland:
I think it's likely it'll get called when changing monitor as `NotifyScreenInfoChanged` is called when the presentation source changes.

https://github.com/cefsharp/CefSharp/blob/master/CefSharp.Wpf/ChromiumWebBrowser.cs#L769

amaitland:
Would appear that the presentation source doesn't change when switching monitors (not quite sure what I got the idea that it would...oh well).

So I think we need to do a little more before this can be merged.

amaitland:
> The width of the virtual screen, in pixels. The virtual screen is the bounding rectangle of all display monitors. The SM_XVIRTUALSCREEN metric is the coordinates for the left side of the virtual screen. 

Looking at `SM_XVIRTUALSCREEN` closer. So it's the complete virtual desktop? I guess that might work, I would have though `CEF` would be expecting details for a single display.

arsher:
Yes, as I couldn't get `GetRootScreenRect` get called again I went with the whole desktop approach. This is why I wasn't sure this is the best solution even if it works.

arsher:
The problem this should've been resolving is already gone, so closing.

Issue #1009, opened 2015-05-07
Feature Request - Add CefSSLInfo wrapper


amaitland:
Add wrapper and expose `CefSSLInfo` for use with `OnCertificateError`

amaitland:
I've added an `ISslInfo` stub class, so the outward facing `API` doesn't have to change later.

amaitland:
Bumping the full implementation to `45` as its' not really a release critical feature. The basic structure is there so the missing pieces can be exposed at any point really.

PR #1190, opened 2015-08-13
Revise Event Handlers, Add ILoadHandler, IDisplayHandler and remove IPopupHandler

amaitland:
I think it's worth releasing a `-pre` release of `master` shortly, so trying to get the last `API` changes done before then (at least all the ones that I can think of).

- `EventArgs` are now directly passed into `IWebBrowserInternal` methods so simplify things, make it much easier to add params now
- Mark `WPF CanReload` property as `Obsolete` (minor cleanup, use `IsLoading` now)
- All the `EventArgs` wrappers should now have access to the relevant underlying `CEF` object e.g. browser/frame
- Moved the `Title` property from `IWebBrowser`, to `IWpfWebBrowser`, makes more sense to subscribe to the event in `WinForms` and `OffScreen`

### DIFF
--- a/CefSharp.Core/Internals/ClientAdapter.cpp
+++ b/CefSharp.Core/Internals/ClientAdapter.cpp
@@ -8,6 +8,7 @@
 #include "ClientAdapter.h"
 #include "CefRequestWrapper.h"
 #include "CefContextMenuParamsWrapper.h"
+#include "CefMenuModelWrapper.h"
 #include "CefDragDataWrapper.h"
 #include "TypeConversion.h"
 #include "CefSharpBrowserWrapper.h"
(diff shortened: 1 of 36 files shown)


amaitland:
Still need to review the `CefDisplayHandler` implementations
http://magpcss.org/ceforum/apidocs3/projects/%28default%29/CefDisplayHandler.html

There's a few that return `bool`, that are currently had coded to `return true`, should consider providing the user access to these. Only difficultly becomes the `WPF` control and it's dependency properties.

amaitland:
I hope to squeeze the last of the `API` changes into this `PR` before we release the `-pre`
(Other than things like adding the `SSLWrapper`, which will likely get a stub implementation so the `API` is stable).

CefSharp.Core/Internals/ClientAdapter.cpp
```diff
@@ -213,33 +213,48 @@ namespace CefSharp
...
+            else
+            {
+                _browserControl->SetAddress(args);
             }
         }
 
         void ClientAdapter::OnTitleChange(CefRefPtr<CefBrowser> browser, const CefString& title)
         {
+            //TODO: Expand this handling so user has more control
+            //NOTE: This should probably be changed if as users can now provide their own popup windows
```
amaitland: This needs to be reviewed

amaitland:
So this ended up being a major change in the end. Basic summary

- Remove `IPopupHandler`
- Add `ILoadHandler`
- Add `IDisplayHandler`
- Add  super simple `CefMenuModelWrapper` (so the `API` doesn't have to change later)

So the interface now more directly reflects `CEF` (still using event args for some methods though).
Internal handlers for event handlers e.g. `FrameLoadEnd` only get called for popups. The corresponding `ILoadHandler` method will be called for both.

Hopefully this keeps things simple, gives everyone access to the underlying `CEF API` if required and is easy to maintain.

amaitland:
For  #1009 added a basic `ISslInfo` stub so the public facing `API` is less likely to change between `-pre` and official release

PR #1194, opened 2015-08-14
Pending tasks only cancelled on real timeouts

cefsharp-ms:
I found a bug in the pending task repository, tasks were always cancelled when a timeout is set.

### DIFF
--- a/CefSharp/Internals/PendingTaskRepository.cs
+++ b/CefSharp/Internals/PendingTaskRepository.cs
@@ -35,13 +35,7 @@ namespace CefSharp.Internals
 
             if (timeout.HasValue)
             {
-                Timer timer = null;
-                timer = new Timer(state =>
-                {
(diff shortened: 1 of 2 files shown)


CefSharp/Internals/PendingTaskRepository.cs
```diff
@@ -35,13 +35,7 @@ public sealed class PendingTaskRepository<TResult>
 
             if (timeout.HasValue)
             {
-                Timer timer = null;
-                timer = new Timer(state =>
-                {
-                    timer.Dispose();
-                    RemovePendingTask(id);
```
amaitland: Will the `Task` still be removed?

cefsharp-ms: Nice catch, I missed that part. It's fixed now.

PR #1207, opened 2015-08-18
High Dpi again

bjarteskogoy:
As far as I can see it works. I'll do some testing with this on our alpha release channel and see what happens.

### DIFF
--- a/CefSharp.Core/Internals/RenderClientAdapter.h
+++ b/CefSharp.Core/Internals/RenderClientAdapter.h
@@ -47,11 +47,11 @@ namespace CefSharp
                 _popupBitmapInfo = nullptr;
             }
 
-			void CreateBitmapInfo()  
-			{  
+            void CreateBitmapInfo()  
+            {  
                 _mainBitmapInfo = _renderWebBrowser->CreateBitmapInfo(false);  
                 _popupBitmapInfo = _renderWebBrowser->CreateBitmapInfo(true);  
-			}
+            }
 
             // CefClient
             virtual DECL CefRefPtr<CefRenderHandler> GetRenderHandler() OVERRIDE{ return this; };
--- a/CefSharp.Core/Internals/RenderClientAdapter.h
+++ b/CefSharp.Core/Internals/RenderClientAdapter.h
@@ -56,8 +56,6 @@ namespace CefSharp
             // CefRenderHandler
             virtual DECL bool GetScreenInfo(CefRefPtr<CefBrowser> browser, CefScreenInfo& screen_info) OVERRIDE
             {
-                return false;
-
                 if ((IRenderWebBrowser^)_renderWebBrowser == nullptr)
                 {
                     return false;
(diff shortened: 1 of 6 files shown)


amaitland:
Did you end up looking through the updated `cefclient` example?

https://bitbucket.org/chromiumembedded/cef/commits/c27da0e7b120704fe2480aec7bfb624fe52ebda3

bjarteskogoy:
I did some weeks ago, but had forgotten about it ... Thank's for reminding me :) Basically the same things we're doing, so the result is the same, I believe. I can change things a bit, so that method and terminology matches the cef sample.

amaitland:
> I can change things a bit, so that method and terminology matches the cef sample.

Do you think it's worth the effort? If you think there are some long term benefits then sure.

Few formatting issues, spaces vs tabs. Other than that it looks solid :+1:

bjarteskogoy:
Then we'll leave things as they are :) If it works, it works. We're running this code (on master) on our Alpha release. Seems to be working. We're also using quite a bit of Javascript interaction, but everything seems good this far.

bjarteskogoy:
I've just discovered that there is a bug calculating mouse positions. Seems like we're sending device coordinates to Cef. When using the WritableBitmap with correct dpi all coordinates and sizes should be logical sizes. I'll fix that.

amaitland:
> When using the WritableBitmap with correct dpi all coordinates and sizes should be logical sizes. I'll fix that.

I had a quick look at the `cefclient` source, do we need to scale the `drag` coordinates also? (I would assume we need to, haven't looked into it though)

bjarteskogoy:
Are you talking about drag for selecting content, or drag/drop? Mouse drag for selection of content seems to be working correctly with this code.

amaitland:
> Are you talking about drag for selecting content, or drag/drop? Mouse drag for selection of content seems to be working correctly with this code.

There's #1210 also

bjarteskogoy:
I think the idea is that cef offscreen uses https://github.com/cefsharp/CefSharp/pull/1207/files#diff-bfec65df46ec49f4ed79472147689599L1201 to calculate mouse positions. If so maybe it should just work.

mmamone:
I just merged your patch into mine, and everything seems to be fine (there's no adorner layer, so oddities with dragging won't be obvious).
The only thing I noticed, is that while there is text selection, you can see the single pixel borders around the HTML elements. I've attached a pic.
cefsharp-highdpi

amaitland:
> The only thing I noticed, is that while there is text selection, you can see the single pixel borders around the HTML elements. I've attached a pic.

Does this differ to the `cefclient` example in `offscreen rendering mode`?

@bjarteskogoy I see you've pushed a more commits. Are there any other scenarios that need to be tested before merging this?

mmamone:
Mea-culpa - same in the cefclient example.
Didn't realise there was a 1291 build.

bjarteskogoy:
My last commit fixed placement and size of popups. Tested with http://www.w3schools.com/tags/tryit.asp?filename=tryhtml_select.

PR #1224, opened 2015-08-27
Made sure that BitmapInfo is created for offscreen popup windows as well

cefsharp-ms:
Just a small fix for the latest changes on the master branch.


amaitland:
Probably need to step through the whole thing again, it's getting rather messy, hopefully some of the code can be simplified.

Issue #1143, opened 2015-07-23
Add Dispose checks to Callbacks


amaitland:
Followup on #1142, implementing for other callbacks.

amaitland:
@jankurianski I've added `IsDisposed` to most of the callbacks and new interfaces. What do you think about having the user check `IsDisposed` before executing the callback?

jankurianski:
> I've added IsDisposed to most of the callbacks and new interfaces. What do you think about having the user check IsDisposed before executing the callback?

I think as long as it is an established pattern it is fine - perhaps by making all the `CefSharp.Example` code have an `IsDisposed` check before callbacks?  My only concern is that the crash in #1142 happened to me rarely and it was down to timing, so if users leave out the `IsDisposed` checks they might get rare crashes that would be hard to track down.

amaitland:
> I think as long as it is an established pattern it is fine - perhaps by making all the CefSharp.Example code have an IsDisposed check before callbacks?

It's looking like that will become the default pattern, so establishing that in `CefSharp.Example` project sounds sensible :+1: 

> My only concern is that the crash in #1142 happened to me rarely and it was down to timing, so if users leave out the IsDisposed checks they might get rare crashes that would be hard to track down.

In that case we should include a `ThrowIfDisposed` method, so it's clear what's going on.

jankurianski:
> In that case we should include a ThrowIfDisposed method, so it's clear what's going on.

Makes sense :smile:

amaitland:
I've added `ThrowifDisposed` to what I hope is all the callbacks. https://github.com/cefsharp/CefSharp/commit/72b1bcab4946354c45af98d8d73376345b443f84

Do you mind running your tests to see if the change works for your app?

jankurianski:
> Do you mind running your tests to see if the change works for your app?

Definitely!  I will report back.

amaitland:
> Definitely! I will report back.

Much appreciated :smile:

jankurianski:
Not looking too good, as I am hitting a C++ breakpoint early on in my tests (Exception Code: `0x80000003`, Exception Information: `A breakpoint was encountered.`).  The stack is entirely in `libcef.dll`:

```
>	libcef.dll!`anonymous namespace'::PureCall()  Line 29	C++
 	libcef.dll!_purecall()  Line 58	C
[...]

PR #1247, opened 2015-09-04
Update comments and examples to avoid users disposing of callback objects

jankurianski:
So that users leave the disposal of these objects to CefSharp.  Related to #1143.

### DIFF
--- a/CefSharp.Example/DownloadHandler.cs
+++ b/CefSharp.Example/DownloadHandler.cs
@@ -11,7 +11,6 @@ namespace CefSharp.Example
             if (!callback.IsDisposed)
             {
                 callback.Continue(downloadItem.SuggestedFileName, showDialog: true);
-                callback.Dispose();
             }
         }
 
--- a/CefSharp.Example/RequestHandler.cs
+++ b/CefSharp.Example/RequestHandler.cs
@@ -26,12 +26,9 @@ namespace CefSharp.Example
             //NOTE: When executing the callback in an async fashion need to check to see if it's disposed
             if (!callback.IsDisposed)
             {
-                using (callback)
-                {
-                    //To allow certificate
-                    //callback.Continue(true);
-                    //return true;
-                }
+                //To allow certificate
+                //callback.Continue(true);
+                //return true;
             }
 
             return false;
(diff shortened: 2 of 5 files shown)


jankurianski:
@amaitland What do you think - is it a good approach?  I figured if we make it obvious that *all* callbacks should never be disposed manually - then users won't get into trouble in specific cases (like with the `ResourceHandler` `callback`).

amaitland:
I think the only problem will be is if users keep hold of a callback, though never actually does anything with it. In that instance it could be a problem. How likely is this to happen? Not very. Thoughts?

jankurianski:
> I think the only problem will be is if users keep hold of a callback, though never actually does anything with it. In that instance it could be a problem. How likely is this to happen? Not very. Thoughts?

I have a feeling that CEF would stall/freeze if you `return true` and then never call `callback.Continue()` or `Cancel()`.  I'll try this out with a couple of the callbacks.

Actually, that makes me think... if the user returns `false` from their handler, then the handler calling code in `C++` needs to dispose of the wrapper objects given to the handler.  As the assumption is that they are not going to handle anything so the objects are not going to be used anymore.

amaitland:
> Actually, that makes me think... if the user returns false from their handler, then the handler calling code in C++ needs to dispose of the wrapper objects given to the handler. As the assumption is that they are not going to handle anything so the objects are not going to be used anymore.

The callback should go out of scope if you never take a reference to it, they're managed wrappers, so the `GC` should free the resources. It's worth double checking though.

jankurianski:
> The callback should go out of scope if you never take a reference to it, they're managed wrappers, so the GC should free the resources. It's worth double checking though.

I looked into the specific case of `CefJSDialogCallbackWrapper` in `ClientAdapter::OnJSDialog`:
https://github.com/cefsharp/CefSharp/blob/63157b8e1bc16f60ba7f4eccc017a8913cedcd6c/CefSharp.Core/Internals/ClientAdapter.cpp#L802

- If you `return true` and never fire the `callback`, Chrome becomes frozen.  So that effectively forces `CefSharp` users to fire the `callback` and it can self-dispose.
- If you `return false` and do not keep reference to the `callback`, it does not get freed after the function is out of scope.
 - I added console `WriteLine`s to the construction, finalize and destroy methods, and when running in `Release` mode (no debugger attached), I see `construction`, but then nothing.  Once I close the program I then see `finalize`.  Also, forcing a `GC.Collect()` and `GC.WaitForPendingFinalizers()` only shows `finalize` too.

I don't know what all this means yet, just reporting it for interest's sake :wink:

amaitland:
I'll run `dotMemory` when I get a second and see information it can provide.

amaitland:
I haven't had a chance to test anything yet, will hopefully get some time later this week.

jankurianski:
> I haven't had a chance to test anything yet, will hopefully get some time later this week.

No worries at all!  Happy to let this PR sit until we figure out the rich approach :+1:

amaitland:
I haven't run a full profile on this yet, so just making some observations whilst I have a few minutes. It looks like the callbacks are being finalized (not disposed of) eventually, it may take some time. So at least they're being cleaned up eventually. This may interfere with `Cef` shutting down cleanly though I've not seen signs of that in the `WPF` example so far.

I'm thinking that I'd advise explicitly disposing of the callbacks where possible. In general I don't think we'll have any issues with calling `Dispose` multiple times, long as the calls are all on the same `Thread`. So wrapping the blocks in a `using` statement and executing a handler should be fine. We could internally add `IsDisposed` checks so subsequent calls are effectively ignored, though I'm not sure that's strictly required (may be cleaner though).

Obviously `ICallback` being a special case as it stands current. We can modify `ResourceHandlerWrapper`, keeping the usage scenario inline with the others for simplicity sake. I think that's probably the best approach.

amaitland:
@jankurianski I created #1257, let me know what you think.

jankurianski:
> I'm thinking that I'd advise explicitly disposing of the callbacks where possible. In general I don't think we'll have any issues with calling Dispose multiple times, long as the calls are all on the same Thread. So wrapping the blocks in a using statement and executing a handler should be fine.

Yep I think that will be fine.  The `AccessViolationException` I found only came up due to the race condition between `ResourceHandlerWrapper` disposing of callbacks on the `IO` thread versus user code disposing in a user thread.  I see you've removed that race condition in #1257 so now only the user or the `GC` should ever dispose of the `callback`.

> We could internally add IsDisposed checks so subsequent calls are effectively ignored, though I'm not sure that's strictly required (may be cleaner though).

I don't think it's necessary as that type of problem would be entirely in user-land code.  If they are calling `Continue` or `Cancel` twice on the same `ICallback` object it would be due to a user bug or a race condition due to the user using threads.  Also to make the internal `IsDisposed` checks truly effective against this we would need to make all `ICallback` implementers thread-safe and it seems more trouble than it's worth.

Closing this in favour of #1257.

PR #1249, opened 2015-09-04
TargetInvocationException is converted properly in JavascriptObjectRepository.TryCallMethod

cefsharp-ms:
Small change that fixes the exception handling in js binding method calls. All errors had the same target invocation error message and not the underlying one.

### DIFF
--- a/CefSharp/Internals/JavascriptObjectRepository.cs
+++ b/CefSharp/Internals/JavascriptObjectRepository.cs
@@ -134,6 +134,11 @@ namespace CefSharp.Internals
 
                 return true;
             }
+            catch(TargetInvocationException e)
+            {
+                var baseException = e.GetBaseException();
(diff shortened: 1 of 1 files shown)


amaitland:
Not much of a description there :stuck_out_tongue:

cefsharp-ms:
Not intentionally :smile:

>Small change that fixes the exception handling in js binding method calls. All errors had the same target invocation error message and not the underlying one.

cefsharp-ms:
* Also updated desc

amaitland:
> Also updated desc

Thanks :smile: 

Prob should port this to the `cefsharp/43` branch also. Will do that at some point.

cefsharp-ms:
Yep, as it's so small it shouldn't be a problem to cherry-pick it to 43.

PR #1252, opened 2015-09-07
Feature detailed exception strings

illfang:
Generated detailed exception message for errors during the invocation of injected javascript methods / properties.


amaitland:
Please provide a description.

PR #1257, opened 2015-09-08
Improve Callback Handling in Examples

amaitland:
This reverts  #1246, instead modify `ResourceHandlerWrapper` so it no longer disposes of the callback.

Also added/update how the examples deal with callbacks.

### DIFF
--- a/CefSharp.Core/Internals/CefCallbackWrapper.h
+++ b/CefSharp.Core/Internals/CefCallbackWrapper.h
@@ -42,6 +42,8 @@ namespace CefSharp
                 ThrowIfDisposed();
 
                 _callback->Cancel();
+
+                delete this;
             }
 
             virtual void Continue()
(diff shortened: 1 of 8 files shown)


jankurianski:
Good approach with `ResourceHandlerWrapper` - so all the callbacks are disposed consistently now.  I tested this change by adding a sleep prior to `callback.Continue()` and cancelling the `ResourceHandler` (which occurs a lot naturally in my application), and everything worked OK.

amaitland:
> and everything worked OK.

Thanks

Are you happy to go with this over #1247?

If your happy then I'll merge this into `master` and `43`

jankurianski:
I just noticed the scheme handler example could do with a `using (callback)`: https://github.com/amaitland/CefSharp/blob/improvement/callbacks/CefSharp.Example/CefSharpSchemeHandler.cs#L65

> Are you happy to go with this over #1247?

Yes, definitely like this approach better :+1:

amaitland:
> I just noticed the scheme handler example could do with a using (callback): https://github.com/amaitland/CefSharp/blob/improvement/callbacks/CefSharp.Example/CefSharpSchemeHandler.cs#L65

Thanks, I'll update :+1:

amaitland:
Both branches updated, thanks :+1:

PR #1268, opened 2015-09-15
Register Custom Schemes in all sub processes

### LINKED ISSUE #1221: Zoom levels don't seem to work when using a custom scheme handler
#### Issue Summary
Invoking Zoom when using a custom scheme doesn't actually zoom the document.

#### Steps to Reproduce
1. Visit a page using a custom scheme (the Cef WPF sample page is a good one)
1. Click zoom in/zoom out, and observe nothing happens
1. Try the same on a regular http page, and everything works as expected

Cannot check this against a cef build, as I cannot use a custom scheme with them.

#### Technical details
* CefSharp Version: 1691e172bd85acc65e12d58d73787c2a54de592e
* Operating System: Windows 8.1
* x86
* WPF

### LINKED ISSUE #1263: Launching popup from a page using a custom scheme causes the render process to crash
Steps to reproduce:
- Open the CefSharp WPF example application
- In the execute script pane run the script <code>window.open("http://google.com")</code>

This is because custom schemes need to be registered in both the parent and all sub processes.  When the browser subprocess is run <code>CefAppUnmanagedWrapper::OnRegisterCustomSchemes</code> is executed by Cef before <code>CefAppUnmanagedWrapper::OnRenderThreadCreated</code> so the subprocess has no access to the custom scheme settings at the time of register.

Suggested solution maybe is to share scheme settings through a named pipe before calling <code>CefExecuteProcess</code>

amaitland:
Fixes #1263
Fixes #1221

Due to ordering of events (see #1263) for details, had to pass scheme names through as command line arguments so they're registered.


amaitland:
This will be backported in `cefsharp/43`

amaitland:
Just need to confirm that multiple schemes are parsed correctly.

amaitland:
Tested with two schemes, everything seems to be working ok.

PR #1271, opened 2015-09-15
OffScreen Example saving the same Google.com image twice.

ekevoo:


amaitland:
Please include a description.

When I run the example as it is currently it works fine. When I run your modifications I get an empty image for the first image, so likely a timing issue.

Out of curiosity what hardware are you using?

amaitland:
Try this change https://github.com/cefsharp/CefSharp/commit/9d64907cdcb7ecd8f4a2062691d962acaf43cf4f

ekevoo:
That change works, too. Thanks.

My Hardware? It's a regular 64 bits PC. It runs 8.1.

PR #1278, opened 2015-09-19
PrintToPDF support

arsher:
I've created bindings for the new PrintToPDF feature in CEF.
Unfortunately the custom HeaderFooterTitle and Url are not working for some reason, no matter how I assign them, maybe it's a CEF/Chromium issue. Otherwise it looks fine.

### DIFF
--- a/CefSharp.Core/Internals/CefBrowserHostWrapper.cpp
+++ b/CefSharp.Core/Internals/CefBrowserHostWrapper.cpp
@@ -4,6 +4,7 @@
 
 #include "Stdafx.h"
 #include "CefBrowserHostWrapper.h"
+#include "CefPdfPrintCallbackWrapper.h"
 #include "WindowInfo.h"
 
 void CefBrowserHostWrapper::StartDownload(String^ url)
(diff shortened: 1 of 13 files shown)


CefSharp/CefSharp.csproj
```diff
@@ -76,6 +76,7 @@
     <DocumentationFile>bin\x64\Release\CefSharp.XML</DocumentationFile>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="PresentationFramework" />
```
amaitland: Is this required?

CefSharp.Core/Internals/CefBrowserHostWrapper.cpp
```diff
@@ -20,6 +21,40 @@ void CefBrowserHostWrapper::Print()
     _browserHost->Print();
 }
 
+void CefBrowserHostWrapper::PrintToPDF(String^ path, CefSharpPdfPrintSettings^ settings, IPrintToPdfCallback^ callback)
+{
+    CefPdfPrintSettings nativeSettings;
+    CefString(&nativeSettings.header_footer_title).FromString(StringUtils::ToNative(settings->HeaderFooterTitle));
```
amaitland: This can likely be simplified e.g.
https://github.com/cefsharp/CefSharp/blob/master/CefSharp.Core/BrowserSettings.h#L42

CefSharp/IBrowserHost.cs
```diff
@@ -11,6 +11,14 @@ public interface IBrowserHost : IDisposable
...
         void StartDownload(string url);
         void Print();
+        /// <summary>
+        /// Prints the current browser contents to the PDF file and executes a callback
+        /// on completion. The caller is responsible for deleting the file when done.
+        /// </summary>
+        /// <param name="path">PDF file location.</param>
+        /// <param name="settings">Settings.</param>
+        /// <param name="callback">The callback executed on completion.</param>
+        void PrintToPDF(string path, CefSharpPdfPrintSettings settings, IPrintToPdfCallback callback);
```
amaitland: What do you think about removing the callback and wrapping this in a Task? Can return a simple structure. Would be similar to GetTextAsync()

https://github.com/cefsharp/CefSharp/blob/bda5e9eeea52226aba3b0eb57438030e7f50d655/CefSharp.Core/Internals/CefFrameWrapper.cpp#L123

CefSharp.Core/Internals/CefBrowserHostWrapper.cpp
```diff
@@ -20,6 +21,40 @@ void CefBrowserHostWrapper::Print()
...
+    nativeSettings.header_footer_enabled = settings->HeaderFooterEnabled ? 1 : 0;
+    nativeSettings.landscape = settings->Landscape ? 1 : 0;
+    nativeSettings.selection_only = settings->SelectionOnly ? 1 : 0;
+    nativeSettings.margin_bottom = settings->MarginBottom;
+    nativeSettings.margin_top = settings->MarginTop;
+    nativeSettings.margin_left= settings->MarginLeft;
+    nativeSettings.margin_right = settings->MarginRight;
+    nativeSettings.page_height = settings->PageHeight;
+    nativeSettings.page_width = settings->PageWidth;
+    switch (settings->MarginType)
```
amaitland: Can likely cast the enum and remove the switch.

CefSharp/CefSharp.csproj
arsher: Uhh, forgot about it

CefSharp.Core/Internals/CefBrowserHostWrapper.cpp
arsher: Good point

CefSharp/IBrowserHost.cs
arsher: Sounds like a good idea, let me take a look.

arsher: To me it seems like that path param will be exactly whatever you specify in PrintToPDF so I don't think we need that at all. How about returning a simple Task<bool>, or better just a Task and signal an exception on failure?

amaitland: Exceptions are expensive and in this case unnecessary, I'd rather return a bool. Fair point about the path.

amaitland:
Merged into `master` with changes. See https://github.com/cefsharp/CefSharp/commit/18434860507085f8692ce6fc368caa48c2c6fac0

PR #1283, opened 2015-09-23
Update cef.sdk.profs import to use SolutionDir

Haraguroicha:
let cef.sdk reference from SolutionDir\packages not project relative parent

### DIFF
--- a/CefSharp.BrowserSubprocess.Core/CefSharp.BrowserSubprocess.Core.vcxproj
+++ b/CefSharp.BrowserSubprocess.Core/CefSharp.BrowserSubprocess.Core.vcxproj
@@ -1,6 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project DefaultTargets="Build" ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <Import Project="..\packages\cef.sdk.3.2454.1317\build\cef.sdk.props" Condition="Exists('..\packages\cef.sdk.3.2454.1317\build\cef.sdk.props')" />
+  <Import Project="$(SolutionDir)\packages\cef.sdk.3.2454.1317\build\cef.sdk.props" Condition="Exists('$(SolutionDir)\packages\cef.sdk.3.2454.1317\build\cef.sdk.props')" />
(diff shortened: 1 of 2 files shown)


amaitland:
In what case were you experiencing a problem?

amaitland:
Also can you include a detailed description of the problem this PR solves? In which scenario is the proposed change required?

Haraguroicha:
When I add CefSharp as a submodule, the `packages` folder is not relative parent to `CefSharp.Core` and `CefSharp.BrowserSubprocess.Core`, `packages` is always relative with `$(SolutionDir)\packages`

amaitland:
> When I add CefSharp as a submodule, the packages folder is not relative parent to CefSharp.Core and CefSharp.BrowserSubprocess.Core, packages is always relative with $(SolutionDir)\packages

As a submodule to what exactly? Whilst I'm not against the change, it's just an odd one to be making in the context of the `CefSharp` codebase.

Haraguroicha:
> As a submodule to what exactly?

Sorry for my bad explanation!

I'm create a project and references to `CefSharp`, `CefSharp.BrowserSubprocess`, `CefSharp.BrowserSubprocess.Core`, `CefSharp.Core`, `CefSharp.WinForm` to build a experimental application, but `CefSharp` was add as a submodule in my repository, so directory will be like this:

```
MyProject Solution Repo
|- MyProject
  |- MyProject.csproj
  |- ...
|- CefSharp
  |- CefSharp3.sln
  |- CefSharp.Core
    |- CefSharp.Core.vcxproj
    |- ...
  |- ...
|- packages
|- MyProject.sln
```

Then it will happened an exception when build, because `CefSharp.Core.vcxproj` is reference `cef.sdk` to `$(SolutionDir)\CefSharp\packages` not `$(SolutionDir)\packages`, so try `$(SolutionDir)\packages` instead of `..\packages` will be better.

amaitland:
Fair enough. I'd generally recommend that people use the official `Nuget` packages or generate their own `nuget` packages using the build script unless they're trying to debug a specific problem with `CefSharp`.

amaitland:
Having just upgraded the `cef.sdk` packages, it appears that `Nuget` defaults back to using the relative path, so this change may be overwritten if care isn't taken when upgrading to newer versions.

Issue #1306, opened 2015-10-01
Upgrade to CEF 3.2454.1326


amaitland:
A newer version has appeared on `cefbuilds.com`, so update `2454` to a release version of `Chromium`

https://cefbuilds.com/#branch_2454

amaitland:
Note that `IRequestHandler.OnBeforePluginLoad` has been removed. Is now part of the ` CefRequestContextHandler` and will need to be re-implemented 

http://magpcss.org/ceforum/apidocs3/projects/%28default%29/CefRequestContextHandler.html#OnBeforePluginLoad%28constCefString&,constCefString&,constCefString&,CefRefPtr%3CCefWebPluginInfo%3E,PluginPolicy*%29

PR #1310, opened 2015-10-02
Upgrade build.ps1 to reference cef.redist 3.2454.1326

illfang:
Updated the default redist version in the build script from 3.2454.1317 to 3.2454.1326

cefsharp/CefSharp#1306


amaitland:
Thanks :+1:

PR #1318, opened 2015-10-09
Feature multi binding test

illfang:
added multi binding test for testing bindings with multiple contexts and dynamic context creation / destruction

- one browser, multiple changing contexts
- also added a link to the popuptest

#1315

Issue #1317, opened 2015-10-08
Crash when adding Facebook login button in BindingTest.html


duboi381:
<H4>Issue Summary</H4>

What is the expected output? What do you see instead?

Expected: We should be able to use BoundObject without crash
Observation: Application crashed

<H4>Technical details</H4>

Version: Chromium: 43.0.2357.130, CEF: r3.2357.1287.g861c26e, CefSharp: 43.0.0.0
Windows 7
x86
WinForms *.Net framework :4.5

<H4>Steps to Reproduce</H4>

1- Add a facebook login button in BindingTest.html
<pre><code>&lt;html&gt;
    &lt;head&gt;
        &lt;title&gt;Binding Test&lt;/title&gt;
    &lt;/head&gt;
![step_to_reproduce](https://cloud.githubusercontent.com/assets/12235562/10370605/a9852996-6dad-11e5-8d05-ff397fad9283.gif)

    &lt;body&gt;
        &lt;div id="fb-root"&gt;&lt;/div&gt;
        &lt;script&gt;
        (function (d, s, id) {
            var js, fjs = d.getElementsByTagName(s)[0];

            if (!d.getElementById(id)) {
                js = d.createElement(s);
                js.id = id;
                js.src = "http://connect.facebook.net/en_US/all.js";
                fjs.parentNode.insertBefore(js, fjs);
            }
        }(document, "script", "facebook-jssdk"));
        
        window.fbAsyncInit = function () {
            FB.init({
                version: "v2.5",
                appId: "160468374291578"
            });
        };
        &lt;/script&gt;
        &lt;input id="customFbBtn" type="button" onclick='FB.login(function(response){});' value="Sign in with Facebook"&gt;&lt;/input&gt;</code></pre>
2- Run CefSharp.WinForms.Example
3- Change url to "custom://cefsharp/BindingTest.html"
4- Click the "Test Callback" button

![step_to_reproduce](https://cloud.githubusercontent.com/assets/12235562/10370624/bc6a2386-6dad-11e5-8203-3ed33ed23af1.gif)

<H4>Additional Information</H4>
If I try to reproduce with AsyncBoundObject function in a button, the application will not crash, but the "then" callback will never be called<br>
If I try to reproduce with Chromium 45 3.2454.1326, with pre-release CefSharp.Common/CefSharp.WinForm .45.0.0-CI1495, this will also crash<br>
[...]

Issue #1315, opened 2015-10-08
JavascriptCallback also has problem when the page include iframe


euler82:
In Branch Master and 43
If the page include iframe ,the JavascriptCallback not work correctly.
But in Branch 41 
It work correctly.

In the issues #1216 ,There is a detailed description for this problem.

Now the version 41 has some other problem,
such as Scroll too much with MouseWheel, it's a bug: https://bitbucket.org/chromiumembedded/cef/issue/1481 for cef.

I want upgrate to version 43,but the JavascriptCallback also not work correctly.

How to solve this problem. thanks!

amaitland:
> How to solve this problem. thanks!

To start debugging this, download the source code, add your changes to demo the problem (ideal if you can push those to your `GitHub` fork)

Add breakpoint to https://github.com/cefsharp/CefSharp/blob/master/CefSharp.BrowserSubprocess.Core/CefAppUnmanagedWrapper.cpp#L63

Enable `renderer-startup-dialog`
https://github.com/cefsharp/CefSharp/blob/master/CefSharp.Example/CefExample.cs#L48

Attach the debugger to the process that displays a message dialog on load (from memory it's got `Chromium Render` in the description).

See if `CefAppUnmanagedWrapper::OnContextCreated`, step through to see if the binding happens. The `OnContextCreated` event should be called for the main frame and `iframes` (going from memory with that one also)

Along the way you might need to compare to the `41` version
https://github.com/cefsharp/CefSharp/blob/cefsharp/41/CefSharp.BrowserSubprocess.Core/CefAppUnmanagedWrapper.cpp#L47

illfang:
@amaitland I'm working on a patch for this issue. I believe there is a flaw in the CefSharp binding logic for context creation / destruction.

amaitland:
Changes merged into `43` branch, hopefully release `43.0.1` shortly (just needs more testing).

amaitland:
`43.0.1` has been released to `Nuget.org` See https://github.com/cefsharp/CefSharp/releases/tag/v43.0.1 for more details

PR #1319, opened 2015-10-09
Feature multi context wrapper binding

illfang:
The binding of the root objects seems to be flawed. During the context creation the root object wrapper of the browser will be set:
```
wrapper->JavascriptRootObjectWrapper = gcnew JavascriptRootObjectWrapper(browser->GetIdentifier(), wrapper->BrowserProcess);
```
And during the context destruction the root object wrapper is getting released.
However, the context creation and destruction for a browser is frame related and therefore can happen multiple times with arbitrary order. e.g.
- iframe 1 added: context create (overriding the root object of the main site as context destroyed is not called, resources of site root object leaked)
- iframe 2 added: context create  (overriding the root object of frame 1 as context destroyed is not called, resources of iframe 1 root object leaked)
- remove iframe 1: context destroyed is invoked destroying the root object of iframe 2 resulting in exceptions when invoking e.g. callbacks

Without the changes here the test of #1318 will result in application crashes / render process termination / results not being passed back.
e.g.
- Invoke Callback in an iframe after another iframe has been refreshed -> render process crash
- Invoke Callback in an iframe after another iframe has been refreshed in a popup -> application crash similar to #1317 
- Async bound object results are not shown when the result is not available before an iframe is refreshed

With the changes introduced here the above errors do not occur.

```
refactored root object binding to be context aware
- as the binding of root objects occurs during context creation, root
objects are now frame bound and contained within the browser wrapper
- callbacks are browser bound as they keep track of their target for
themself
- introduced concurrent object registry storing objects with an auto
generated id, refactored method and js callback handling to use new type
```

#1315 

I would be glad to hear your feedback on these changes.

### DIFF
--- a/CefSharp.BrowserSubprocess.Core/CefAppUnmanagedWrapper.cpp
+++ b/CefSharp.BrowserSubprocess.Core/CefAppUnmanagedWrapper.cpp
@@ -39,20 +39,20 @@ namespace CefSharp
     // CefRenderProcessHandler
     void CefAppUnmanagedWrapper::OnBrowserCreated(CefRefPtr<CefBrowser> browser)
     {
+        LOG(INFO) << "OnBrowserCreated:" << browser->GetIdentifier();
         auto wrapper = gcnew CefBrowserWrapper(browser);
         _onBrowserCreated->Invoke(wrapper);
 
         //Multiple CefBrowserWrappers created when opening popups
-        _browserWrappers->Add(browser->GetIdentifier(), wrapper);
+        _browserWrappers->TryAdd(browser->GetIdentifier(), wrapper);
     }
 
     void CefAppUnmanagedWrapper::OnBrowserDestroyed(CefRefPtr<CefBrowser> browser)
     {
-        auto wrapper = FindBrowserWrapper(browser->GetIdentifier(), false);
-
-        if (wrapper != nullptr)
+        LOG(INFO) << "OnBrowserDestroyed:" << browser->GetIdentifier();
+        CefBrowserWrapper^ wrapper;
+        if (_browserWrappers->TryRemove(browser->GetIdentifier(), wrapper) && wrapper != nullptr)
         {
-            _browserWrappers->Remove(wrapper->BrowserId);
             _onBrowserDestroyed->Invoke(wrapper);
             delete wrapper;
         }
@@ -60,33 +60,44 @@ namespace CefSharp
 
     void CefAppUnmanagedWrapper::OnContextCreated(CefRefPtr<CefBrowser> browser, CefRefPtr<CefFrame> frame, CefRefPtr<CefV8Context> context)
     {
-        auto wrapper = FindBrowserWrapper(browser->GetIdentifier(), true);
-
+        LOG(INFO) << "OnContextCreated:" << browser->GetIdentifier();
+        auto browserWrapper = FindBrowserWrapper(browser->GetIdentifier(), true);
         if (!Object::ReferenceEquals(_javascriptRootObject, nullptr) || !Object::ReferenceEquals(_javascriptAsyncRootObject, nullptr))
         {
-            wrapper->JavascriptRootObjectWrapper = gcnew JavascriptRootObjectWrapper(browser->GetIdentifier(), wrapper->BrowserProcess);
-            wrapper->JavascriptRootObjectWrapper->Bind(_javascriptRootObject, _javascriptAsyncRootObject, context->GetGlobal());
+            auto rootObjectWrapperDictionary = browserWrapper->JavascriptRootObjectWrappers;
+            auto frameID = frame->GetIdentifier();
+            JavascriptRootObjectWrapper^ wrapper;
+            if (!rootObjectWrapperDictionary->TryGetValue(frameID, wrapper) || wrapper == nullptr)
+            {
+                wrapper = gcnew JavascriptRootObjectWrapper(browser->GetIdentifier(), browserWrapper->BrowserProcess);
+                rootObjectWrapperDictionary[frameID] = wrapper;
+            }
+            else {
+                LOG(WARNING) << "A context has been created for the same browser / frame without context released called previously";
+            }
+            LOG(INFO) << "Bound Root Object Wrapper:" << browser->GetIdentifier() << ":" << frame->GetIdentifier();
+            wrapper->Bind(browserWrapper, _javascriptRootObject, _javascriptAsyncRootObject, context->GetGlobal());
         }
     };
 
     void CefAppUnmanagedWrapper::OnContextReleased(CefRefPtr<CefBrowser> browser, CefRefPtr<CefFrame> frame, CefRefPtr<CefV8Context> context)
     { 
-        auto wrapper = FindBrowserWrapper(browser->GetIdentifier(), true);
-
-        if (wrapper->JavascriptRootObjectWrapper != nullptr)
+        LOG(INFO) << "OnContextReleased:" << browser->GetIdentifier();
+        auto browserWrapper = FindBrowserWrapper(browser->GetIdentifier(), true);
+        auto frameID = frame->GetIdentifier();
+        auto rootObjectWrapperDictionary = browserWrapper->JavascriptRootObjectWrappers;
+        JavascriptRootObjectWrapper^ wrapper;
+        if (rootObjectWrapperDictionary->TryRemove(frameID, wrapper) && wrapper != nullptr)
         {
-            delete wrapper->JavascriptRootObjectWrapper;
-            wrapper->JavascriptRootObjectWrapper = nullptr;
+            delete wrapper;
         }
     };
 
     CefBrowserWrapper^ CefAppUnmanagedWrapper::FindBrowserWrapper(int browserId, bool mustExist)
     {
-        CefBrowserWrapper^ wrapper = nullptr;
-
-        _browserWrappers->TryGetValue(browserId, wrapper);
-
-        if (mustExist && wrapper == nullptr)
+        CefBrowserWrapper^ wrapper;
+        bool invalid = !_browserWrappers->TryGetValue(browserId, wrapper) || wrapper == nullptr;
+        if (mustExist && invalid)
         {
             throw gcnew InvalidOperationException(String::Format("Failed to identify BrowserWrapper in OnContextCreated. : {0}", browserId));
         }
@@ -142,9 +153,8 @@ namespace CefSharp
 
             return true;
         }
-    
-        auto rootObjectWrapper = browserWrapper->JavascriptRootObjectWrapper;
-        auto callbackRegistry = rootObjectWrapper != nullptr ? rootObjectWrapper->CallbackRegistry : nullptr;
+
+        auto callbackRegistry = browserWrapper->CallbackRegistry;
         //these messages are roughly handled the same way
         if (name == kEvaluateJavascriptRequest || name == kJavascriptCallbackRequest)
         {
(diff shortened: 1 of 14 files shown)


amaitland:
Without looking at this in great detail I'm curious as to how well `41.0.1` behaved, as most of the issues created seem to suggest things worked previously. The code has changed, though the lifecycle handling for a context is only slightly different. 

I'll make a few quick comments inline, though no need to action anything just yet.

CefSharp.BrowserSubprocess.Core/CefAppUnmanagedWrapper.cpp
```diff
@@ -60,33 +60,44 @@ namespace CefSharp
...
-            wrapper->JavascriptRootObjectWrapper->Bind(_javascriptRootObject, _javascriptAsyncRootObject, context->GetGlobal());
+            auto rootObjectWrapperDictionary = browserWrapper->JavascriptRootObjectWrappers;
+            auto frameID = frame->GetIdentifier();
+            JavascriptRootObjectWrapper^ wrapper;
+            if (!rootObjectWrapperDictionary->TryGetValue(frameID, wrapper) || wrapper == nullptr)
+            {
+                wrapper = gcnew JavascriptRootObjectWrapper(browser->GetIdentifier(), browserWrapper->BrowserProcess);
+                rootObjectWrapperDictionary[frameID] = wrapper;
+            }
+            else {
```
amaitland: Brace on new line.

illfang:
> The code has changed, though the lifecycle handling for a context is only slightly different.

Maybe there has been a bug in 41 where callback wrappers would not be deleted. I have confirmed that if the constructor is not called, the master branch will also not crash.

CefSharp.BrowserSubprocess.Core/JavascriptRootObjectWrapper.h
```diff
@@ -17,6 +17,8 @@ using namespace CefSharp::Internals::Async;
 
 namespace CefSharp
 {
+    ref class CefBrowserWrapper;
```
amaitland: I dislike using forward declarations. Hopefully with a little refactoring we can do away with this.

CefSharp.BrowserSubprocess.Core/JavascriptMethodHandler.h
```diff
@@ -13,19 +13,20 @@ namespace CefSharp
     {
     private:
         gcroot<Func<array<Object^>^, BrowserProcessResponse^>^> _method;
-        gcroot<JavascriptCallbackRegistry^> _callbackRegistry;
+        gcroot<JavascriptCallbackRegistry^> _callbackRegistryReference;
```
amaitland: Any reason for the variable being renamed?

CefSharp.BrowserSubprocess.Core/ConcurrentObjectRegistry.h
```diff
@@ -0,0 +1,53 @@
...
+        generic <typename T>
+        public ref class ConcurrentObjectRegistry
+        {
+        private:
+            bool _deleteEntriesOnDestroy;
+            Int64 _lastId;
+            ConcurrentDictionary<Int64, T>^ _entries;
+
+        public:
+            ConcurrentObjectRegistry(bool deleteEntriesOnDestroy) :
```
amaitland: Is `deleteEntriesOnDestroy` ever `false`? Can the variable be removed?

CefSharp.BrowserSubprocess.Core/ConcurrentObjectRegistry.cpp
```diff
@@ -0,0 +1,33 @@
...
+        int64 ConcurrentObjectRegistry<T>::RegisterObject(T entry)
+        {
+            int64 objectId = Interlocked::Increment(_lastId);
+            _entries->TryAdd(objectId, entry);
+            return objectId;
+        }
+        generic <typename T>
+        bool ConcurrentObjectRegistry<T>::TryGetObject(int64 id, T% entry)
+        {
+            return _entries->TryGetValue(id, entry) && entry != nullptr;
```
amaitland: What purpose does the `nullptr` check serve?

CefSharp.BrowserSubprocess.Core/CefBrowserWrapper.h
```diff
@@ -26,15 +27,38 @@ namespace CefSharp
...
+        
+        property JavascriptCallbackRegistry^ CallbackRegistry
+        {
+            CefSharp::Internals::JavascriptCallbackRegistry^ get()
+            {
+                return _callbackRegistry;
+            }
+        }
+
+        property ConcurrentObjectRegistry<JavascriptAsyncMethodCallback^>^ MethodCallbackRegistry
```
amaitland: Should probably have `Async` in the name so it's clear which implementation it belongs to.

CefSharp.BrowserSubprocess.Core/CefAppUnmanagedWrapper.cpp
```diff
@@ -60,33 +60,44 @@ namespace CefSharp
...
     };
 
     void CefAppUnmanagedWrapper::OnContextReleased(CefRefPtr<CefBrowser> browser, CefRefPtr<CefFrame> frame, CefRefPtr<CefV8Context> context)
     { 
-        auto wrapper = FindBrowserWrapper(browser->GetIdentifier(), true);
-
-        if (wrapper->JavascriptRootObjectWrapper != nullptr)
+        LOG(INFO) << "OnContextReleased:" << browser->GetIdentifier();
+        auto browserWrapper = FindBrowserWrapper(browser->GetIdentifier(), true);
+        auto frameID = frame->GetIdentifier();
```
amaitland: `frameID` should be `frameId`

amaitland:
> . I have confirmed that if the constructor is not called, the master branch will also not crash.

What `constructor` exactly?

illfang:
> What constructor exactly?

sorry, the **de**structor of the registered objects / callbacks

CefSharp.BrowserSubprocess.Core/ConcurrentObjectRegistry.cpp
illfang: TryGetValue will resolve to true if the key has been found even for a null value. If the TryGetObject returns true, there is an object passed which can be used.

CefSharp.BrowserSubprocess.Core/ConcurrentObjectRegistry.h
illfang: Currently, it will never be false. But usually such data structures do not delete their children when removed so this seemed like the default behaviour and if the developer wants to automatically delete all contained objects he can set this to true.
As it is currently only used with true it can also be removed but I thought of this class as a general data structure.

CefSharp.BrowserSubprocess.Core/ConcurrentObjectRegistry.cpp
amaitland: How exactly does `null` get added to a key? Shouldn't a key be removed if the object is `null`?

CefSharp.BrowserSubprocess.Core/ConcurrentObjectRegistry.h
amaitland: > and if the developer wants to automatically delete all contained objects he can set this to true.

Which developer exactly? This code isn't exposed to the public.

> As it is currently only used with true it can also be removed but I thought of this class as a general data structure.

I'd rather not have code that's unused just lying around as a **maybe**. Please remove.

amaitland:
I've merged in your other two `PR's` can you please merge in `master` so the changes are reflected.

CefSharp.BrowserSubprocess.Core/ConcurrentObjectRegistry.h
```diff
@@ -0,0 +1,53 @@
...
+#include "JavascriptCallbackWrapper.h"
+
+using namespace System::Collections::Concurrent;
+
+namespace CefSharp
+{
+    namespace Internals
+    {
+        generic <typename T>
+        public ref class ConcurrentObjectRegistry
```
amaitland: In general, I'm not convinced this class is actually required. It only removes a small amount of code duplication and introduces what I'm wondering is an unnecessary layer of abstraction.

illfang: The same logic has been used in two places
https://github.com/cefsharp/CefSharp/blob/master/CefSharp.BrowserSubprocess.Core/JavascriptRootObjectWrapper.h#L29 and in the JavascriptCallbackRegistry

Without this object the
- int64 SaveMethodCallback(JavascriptAsyncMethodCallback^ callback);
- bool TryGetAndRemoveMethodCallback(int64 id, JavascriptAsyncMethodCallback^% callback);
methods have to be added to the browser wrapper which are now handled by the object repository where it is kind of hard to understand to what `_lastCallback` revers with the `_methodCallbacks` and `_callbackRegistry` members. 
Furthermore, it seemd inconsistent that there is JavascriptCallbackRegistry but for async callbacks the logic is contained directly in the class.

CefSharp.BrowserSubprocess.Core/ConcurrentObjectRegistry.cpp
illfang: Sorry for the ambiguous explanation. Of course the key cannot be null as the dictionary would throw an exception for this case. However, the registered value can be null which is allowed. TryGetObject would then return true but the passed value would be null

CefSharp.BrowserSubprocess.Core/ConcurrentObjectRegistry.h
illfang: removed the additional logic

CefSharp.BrowserSubprocess.Core/CefAppUnmanagedWrapper.cpp
illfang: :ballot_box_with_check: adjusted

CefSharp.BrowserSubprocess.Core/JavascriptMethodHandler.h
illfang: previously the callbackRegistry has been deleted withi nthe destructor but this is a shared instance among many instances so if one deletes this registry other callbacks will not work anymore.
```
 ~JavascriptMethodHandler()
        {
            delete _method;
            delete _callbackRegistry;
        }
```
https://github.com/cefsharp/CefSharp/blob/master/CefSharp.BrowserSubprocess.Core/JavascriptMethodHandler.h#L28

I have renamed the memeber to make clear that this is only a reference which should not be deleted by this class.

CefSharp.BrowserSubprocess.Core/CefAppUnmanagedWrapper.cpp
illfang: :ok_hand:

illfang:
> I've merged in your other two PR's can you please merge in master so the changes are reflected.

Thank you. I have merged this branch with the new head.

CefSharp.BrowserSubprocess.Core/ConcurrentObjectRegistry.h
amaitland: > Furthermore, it seemd inconsistent that there is JavascriptCallbackRegistry but for async callbacks the logic is contained directly in the class.

I'm happy for that to change :+1:

amaitland:
@illfang I've created #1321 which is based on/inspired by this `PR`. Looking at the problem in more detail and it seems like the cleanest solution is to create a `JavascriptRootObjectWrapper` per `Frame/Context`, and make each callback/method call keep a reference to the relevant `frameId`. Whilst a slightly fiddly solution as modifying the `IPC` indexes is a little tedious, I think it keeps a cleaner separation of concerns.

Thoughts/comments welcome.

PR #1321, opened 2015-10-11
Add support for JSB with frames

amaitland:
Current implementation only works with a single context per `CefBrowser` wrapper, when multiple frames are added a new `V8Context` is created per `Frame`.

This `PR` is based on/inspired by #1319

Support for a `JavascriptRootObjectWrapper` per `V8Context` has been added. Each `JS Callback` and `Async Method` call now includes a `FrameId`, so they can be associated with the correct wrapper.

I haven't yet done extensive testing, it appears to work. The changes are quite fiddly as params passed using the native `IPC` are index based. (At some point a layer of abstraction needs to be added to handle this complexity as it's a little hard to follow at the moment, It was a deliberate decision to avoid doing that as yet, I though it was more important to fully understand the problem domain before attempting to engineer a solution).

### DIFF
--- a/CefSharp.BrowserSubprocess.Core/Async/JavascriptAsyncMethodHandler.cpp
+++ b/CefSharp.BrowserSubprocess.Core/Async/JavascriptAsyncMethodHandler.cpp
@@ -38,10 +38,11 @@ namespace CefSharp
                     SerializeV8Object(arguments[i], params, i, _callbackRegistry);
                 }
 
-                SetInt64(_objectId, argList, 0);
-                SetInt64(callbackId, argList, 1);
-                argList->SetString(2, name);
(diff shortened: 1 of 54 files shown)


amaitland:
Potential fix for #1317 and #1315

illfang:
@amaitland I have tried your solution and it is still possible to crash CEF. (process crash for popups, render process crash for usual page)

I've created a PR on your repository which I believe fixes the remaining issues.
https://github.com/amaitland/CefSharp/pull/4

amaitland:
@illfang Thanks for your hard work on this :+1: I've merged into `master`, if all goes well then I'll probably back port the changes into the `cefsharp/43` branch at some point in the near future.

Issue #1324, opened 2015-10-12
Add IPluginHandler.OnBeforePluginLoad


amaitland:
`CefRequestHandler::OnBeforePluginLoad` has been moved to `CefRequestContextHandler`

https://bitbucket.org/chromiumembedded/cef/commits/b6e560e2eabbff73e49642e9e4340cfecbb09a67?at=2454#Linclude/cef_request_context_handler.hT75

Add a new `IPluginHandler` interface that can be passed into the `CefRequestContextHandler`

PR #1325, opened 2015-10-12
Implement IPluginHandler.OnBeforePluginLoad

amaitland:
Can now be used through the RequestContext
Add Example to WPF application


amaitland:
Resolves #1324

dsherret:
@amaitland IPluginHandler doesn't seem to be used anywhere... the example implements the interface, but does nothing with the class.

amaitland:
@dsherret This commit is from a year ago, the interface has changed. `PR's` are not the place to ask questions. Use https://gitter.im/cefsharp/CefSharp

PR #1344, opened 2015-10-18
Fix for #1336: IJavascriptCallback.ExecuteAsync never returns when callback no longer exists

### LINKED ISSUE #1336: IJavascriptCallback.ExecuteAsync never returns when callback no longer exists
I just upgraded to the latest version 43.0.0 of CefSharp (I believe the previous version was 41).
In 41 when an IJavascriptCallback went out of scope if you tried to call it the JavascriptResponse would be null which was my clue that it had gone out of scope and I would remove it from my callback list.

But in 43.0.0 if the IJavascriptCallback goes out of scope (i.e. the page changes) the call to IJavascriptCallback.ExecuteAsync never returns.  I haven't counted it but after somewhere between 10-20 never-returning calls to IJavascriptCallback.ExecuteAsync something crashes and the DevTools says its been disconnected and the browser becomes unresponsive.

I copied my current code below. I don't know if IJavascriptCallback.isDisposed was in 41, but I added a check for it to make sure I wasn't calling a disposed object and I'm not. As you can see I catch exceptions just in case also. In stepping through this code with the Visual Studio C# debugger I see that every call to ExecuteAsync never returns until something crashes.

Perhaps I've missed something added in 43.0.0 to handle this.  All I do know is it used to work very reliably in 41 returning a null JavascriptResponse when the callback was no longer valid.

Here's my code make the callback:

```c#
private async void CallJsDownloadCallback(IJavascriptCallback jsCallbackForDownloadStatus, int sinkCallbackId, DownloadManagerEventStatus eventType, Guid? itemGuid, string appId, ApplicationStatus appStatus, ApplicationInfo appInfo, long longValue, string stringValue) { JavascriptResponse response = null;
        try
        {
            if (!jsCallbackForDownloadStatus.IsDisposed)
            {
                string appInfoString = JsonConvert.SerializeObject(appInfo, defaultJsonSerializerSettings);

                // Here is the call that NEVER comes back if the IJavascriptCallback has went out of scope
                Task<JavascriptResponse> task = jsCallbackForDownloadStatus.ExecuteAsync(sinkCallbackId, (int)eventType, itemGuid?.ToString(), appId, (int)appStatus, appInfoString, longValue, stringValue);

                response = await task;
            }
            else
            {
                Debug.WriteLine("jsCallbackForStoreStatus was disposed");
            }
        }
        catch (Exception ex)
        {
            Debug.WriteLine("CallJsDownloadCallback exception: " + ex.Message);
        }

        if (response == null)
        {
            // Assume the Javascript callback is now invalid
            m_managerServices.DownloadManagerServiceEventSink.UnsubscribeDownloadEvents(sinkCallbackId);
        }
    }
```

gregmartinhtc:
This should fix #1336 (IJavascriptCallback.ExecuteAsync never returns when callback no longer exists) and likely fixes as-of-yet undiscovered related callback problems.

(Note the changes are actually far simpler than the file comparison on Github indicates mostly due to indentation changes looking added/deleted new lines.  I did several commits/reverts fixing whitespace and other issues (including reverting the wrong commit) trying to get the file compare down to the real changes.)

The code previously in CefAppUnmanagedWrapper.cpp's OnProcessMessageReceived method did not check for a nullptr callbackWrapper resulting in an exception.
Also note that previously the success and reponse variables were not initialized by default which I believe to have been a mistake after following all of the if..else code paths (most importantly some error conditions would never be returned back).
I added a new test to BindingTest.html entitled "Test Disposed Callback" which makes the same 1.5 second delayed callback that "Test Callback" does, but then immediately navigate the page to www.google.com.  Without the fix in CefAppUnmanagedWrapper.cpp the browser's renderer process will die in about 10 seconds.

### DIFF
--- a/CefSharp.Example/Resources/BindingTest.html
+++ b/CefSharp.Example/Resources/BindingTest.html
@@ -113,6 +113,32 @@
         </p>
 
         <p>
+            Disposed Javscript Callback Test (navigates to www.google.com before callback fires)
+            <br />
+            <script type="text/javascript">
+                function disposedCallback(s) {
+                    // This callback should be disposed and should not be called
+                    window.alert("This callback should not have been called");
+
+                    var result = document.getElementById('disposedcbresult');
+                    result.innerText += "Callback: " + s + "" + Date();
+                }
+
+                function testDisposedCallback() {
+                    bound.testCallback(callback);
+
+                    var result = document.getElementById('disposedcbresult');
+                    result.innerText = "The function has returned: " + Date() + "\n";
+
+                    window.location.assign("http://www.google.com");
+                }
+            </script>
+            <button onclick="testDisposedCallback()">Test Disposed Callback</button>
+            <br />
+            <span id="disposedcbresult"></span>
+        </p>
+
+        <p>
             Result of calling bound.repeat("hi ", 5) =
             <script type="text/javascript">
                 var result = bound.repeat("hi ", 5);
--- a/CefSharp.Example/Resources/BindingTest.html
+++ b/CefSharp.Example/Resources/BindingTest.html
@@ -1,15 +1,15 @@
 ﻿<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.0 Transitional//EN">
 <html>
-<head>
-    <title>Binding Test</title>
-</head>
-<body>
-    <p>
-        Async Binding Test
-        <span id="asyncresult"></span>
-        <script type="text/javascript">
+    <head>
+        <title>Binding Test</title>
+    </head>
+    <body>
+        <p>
+            Async Binding Test
+            <span id="asyncresult"></span>
+            <script type="text/javascript">
                 var asResult = document.getElementById('asyncresult');
-
+            
                 function writeAsyncResult(call, end)
                 {
                     var p = document.createElement('p');
@@ -18,16 +18,16 @@
                     var title = document.createTextNode('Async Call: ');
                     var callText = document.createTextNode(call);
                     var endText = document.createTextNode(end);
-
+                
                     p.appendChild(title);
                     p.appendChild(br);
                     p.appendChild(callText);
                     p.appendChild(br2);
                     p.appendChild(endText);
-
+                
                     asResult.appendChild(p);
                 }
-
+            
                 function asyncError()
                 {
                     var call = "Async call (Throw Exception): " + Date();
@@ -37,7 +37,7 @@
                         writeAsyncResult(call, end);
                     });
                 }
-
+            
                 function asyncDivOk()
                 {
                     var call = "Async call (Divide 16 / 2): " + Date();
@@ -47,7 +47,7 @@
                         writeAsyncResult(call, end);
                     });
                 }
-
+            
                 function asyncDivFail()
                 {
                     var call = "Async call (Divide 16 /0): " + Date();
@@ -62,7 +62,7 @@
                         writeAsyncResult(call, end);
                     });
                 }
-
+            
                 function asyncHello()
                 {
                     var call = "Async call (Hello): " + Date();
@@ -72,7 +72,7 @@
                         writeAsyncResult(call, end);
                     });
                 }
-
+            
                 function asyncDoSomething()
                 {
                     var call = "Async call (Long Running Task): " + Date();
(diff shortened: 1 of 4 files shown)


CefSharp.Example/Resources/BindingTest.html
```diff
@@ -113,6 +113,32 @@
         </p>
 
         <p>
+            Disposed Javscript Callback Test (navigates to www.google.com before callback fires)
```
amaitland: Need to split this out into it's own test case.

amaitland:
> (Note the changes are actually far simpler than the file comparison on Github indicates mostly due to indentation changes looking added/deleted new lines. I did several commits/reverts fixing whitespace and other issues (including reverting the wrong commit) trying to get the file compare down to the real changes.)

Maybe it's better if you squash the changes into a single commit then?

amaitland:
> The code previously in CefAppUnmanagedWrapper.cpp's OnProcessMessageReceived method did not check for a nullptr callbackWrapper resulting in an exception.

This is new code post release of `43.0.0`, which has not been extensively tested. So it's great that you picked up on that before the `43.0.1` release :+1: 

The actual code that represents `43.0.0` is https://github.com/gregmartinhtc/CefSharp/blob/12b31a7782523aeca50c073da2caa13df62e3180/CefSharp.BrowserSubprocess.Core/CefAppUnmanagedWrapper.cpp

Not that it makes much different, just important to remember that your looking at the branch, for `43.0.0`, if you wanted the exact code you'd need to look at the `tag`.

Anyways, thanks for taking the time to submit a `PR` :+1: At some point in the near future I hope that `OnProcessMessageReceived` will be refactored into smaller more manageable sections of code. (It was a deliberate choice to leave it as one giant code block up until the code had stabilized).

CefSharp.Example/Resources/BindingTest.html
amaitland: Actually, scrap that. Just leave it here is fine.

CefSharp.Example/Resources/BindingTest.html
```diff
@@ -113,6 +113,32 @@
...
+            <br />
+            <script type="text/javascript">
+                function disposedCallback(s) {
```
amaitland: Braces on new lines please :+1:

gregmartinhtc:
Ah, I see.  I didn't even think to look at any tags.  I see the v43.0.0 tag
in the history listed by Visual Studio now that you mention it.  I guess I
assumed you'd want an edit based on the newest code, but in hindsight that
also might mean I've brought in some destabilizing changes into my compiled
code made after 43.0.0.

CefSharp.Example/Resources/BindingTest.html
gregmartinhtc: How do I split it out into its own test case?  Is there any guidance or
some other test case I can look at?

gregmartinhtc:
I had pushed them to GitHub before realizing my mistakes and I am not a Git
expert and was afraid of trying to reset the head of my fork on GitHub so I
had to leave it like it was or risk screwing something up.

amaitland:
> I guess I assumed you'd want an edit based on the newest code, but in hindsight that also might mean I've brought in some destabilizing changes into my compiled code made after 43.0.0.

That's perfectly fine. It was more for point of reference. Next release will need to have other fixes included that are also relevant to callbacks.

> I had pushed them to GitHub before realizing my mistakes and I am not a Git expert and was afraid of trying to reset the head of my fork on GitHub so I had to leave it like it was or risk screwing something up.

It's never too late, you can squash then push again, overriding existing changes.

amaitland:
I've squashed this into one commit, so it's easier to merge in `master` (Which I've also done).

Relevant commit for `cefsharp/43` branch is https://github.com/cefsharp/CefSharp/commit/1a01b956190acb8cf19ac8709c62dcd706589ce5

Fixed up some formatting and made one of the error messages more verbose, subsequent commit is https://github.com/cefsharp/CefSharp/commit/a598af2a2e0ce5959a67510772fb30308e74965b

amaitland:
The https://github.com/cefsharp/CefSharp/commit/d2aa11ec75be498532fe779d5abbc637aaec0f07 commit also relates to this `PR`. Thanks to @illfang for spotting the mistake.

gregmartinhtc:
Sorry that I got that wrong in my original fix -- that's embarrassing...

amaitland:
> Sorry that I got that wrong in my original fix -- that's embarrassing...

Simple oversight :smile:

PR #1349, opened 2015-10-19
Only copy the full pixel buffer when the dirty region is non empty. Expose CefDoMessageLoopWork

bartwe:


amaitland:
Please provide a description and maybe make the title a little more succinct, it's rather long.

amaitland:
In future it's preferable that changes are grouped logically, more than happy to add the message loop change, the dirty rect obviously needs some testing.

CefSharp.Core/Internals/RenderClientAdapter.h
```diff
@@ -167,7 +167,11 @@ namespace CefSharp
                     bitmapInfo->NumberOfBytes = numberOfBytes;
                 }               
 
-                CopyMemory(backBufferHandle, (void*)buffer, bitmapInfo->NumberOfBytes);
+                if (!bitmapInfo->DirtyRectSupport || ((bitmapInfo->DirtyRect.Width != 0) && (bitmapInfo->DirtyRect.Height != 0)))
```
amaitland: I'm a little surprised that `CEF` is calling `OnPaint` with `Width == 0` and `Height == 0`. Is it called very often? Probably should check the `CEF` source, change should really be applied there, no point calling `OnPaint` if there's nothing to draw.

amaitland: @bartwe Under what scenario are you seeing `Width == 0` and `Height == 0`?

bartwe: I tried to reproduce and couldn't, the code remains correct without it,
please ignore/remove that change.

amaitland: Ok cool, thanks for following up :+1:

PR #1356, opened 2015-10-22
Feature sanitized and verbose js exception strings

illfang:
- added a test case for logging JavaScript exceptions created in a Callback
- added a CreateExceptionString to the StringUtils for creating a verbose message for Javascript exceptions including the message, filename, line number and column; this methods also checks if the exception is available
- use the util function in the app unmanaged wrapper to create the JS exception string for callbacks and js evaluations, by this the `if (exception.get())` will now also be executed in both cases.

New log message (ExceptionTest.html)
```
Uncaught Callback Error (This callback from C# was delayed 500ms)
@ custom://cefsharp/ExceptionTest.html:59:20
```

Previous message
```
Uncaught Callback Error (This callback from C# was delayed 500ms)
```

### DIFF
--- a/CefSharp.Core/Internals/StringUtils.h
+++ b/CefSharp.Core/Internals/StringUtils.h
@@ -6,7 +6,7 @@
 
 #include <vector>
 #include "vcclr_local.h"
-#include "include\internal\cef_string.h"
+#include "include\cef_v8.h"
 
 using namespace System;
 using namespace System::Collections::Generic;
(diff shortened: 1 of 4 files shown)


CefSharp.Example/Resources/ExceptionTest.html
```diff
@@ -52,5 +52,21 @@
                 }
             </script>
         </p>
+        <p>
+            Exception string for error callback:<br />
+            <script type="text/javascript">
+                function errorCallback(s) {
```
amaitland: Braces on new line.

CefSharp.Core/Internals/StringUtils.h
```diff
@@ -117,6 +117,25 @@ namespace CefSharp
...
+            /// <summary>
+            /// Creates a detailed expection string from a provided Cef V8 exception.
+            /// </summary>
+            /// <param name="exception">The exception which will be used as base for the message</param>
+            [DebuggerStepThrough]
+            static CefString CreateExceptionString(CefRefPtr<CefV8Exception> exception)
+            {
+                if (exception.get())
+                {
+                    String^ message = ToClr(exception->GetMessage()) + "\n@ " +
```
amaitland: Would be nicer to create the native `CefString` rather than convert to `String` and then back again.

amaitland:
@illfang Looks like a very useful contribution :+1:

illfang:
> Looks like a very useful contribution

Thank you and thank you for your feedback / review.

I've adjusted the formatting and used a native stringstream to create the exception message. I did not find a better way to concat CefStrings directly which would not be error prone.

amaitland:
> I've adjusted the formatting and used a native stringstream to create the exception message. I did not find a better way to concat CefStrings directly which would not be error prone.

`CEF` uses `stringstreams` within it's own examples, so I think it's probably the best approach :+1:

amaitland:
Also merged into `cefsharp/43`

PR #1358, opened 2015-10-22
Fix offscreen browser browser settings nullptr (breaking current interface)

illfang:
There is a bug with the optional browser settings in the offscreen browser project which will lead to a null poiinter exception if no settings are available.

Example:
https://github.com/illfang/CefSharp/tree/offscreen-browser-browser-settings-nullptr-test
```
System.NullReferenceException: Object reference not set to an instance of an object.
   at CefSharp.BrowserSettings.get_OffScreenTransparentBackground()
   at CefSharp.ManagedCefBrowserAdapter.CreateOffscreenBrowser(IntPtr windowHandle, BrowserSettings browserSettings, RequestContext requestContext, String address)
   at CefSharp.OffScreen.ChromiumWebBrowser.CreateBrowser(IntPtr windowHandle, String address, BrowserSettings browserSettings, RequestContext requestcontext)
   at CefSharp.OffScreen.ChromiumWebBrowser..ctor(String address, BrowserSettings browserSettings, RequestContext requestcontext, Boolean automaticallyCreateBrowser)
```

Probably related to dca52f931fcb7e0d435b2e1f46cd1d49fed5a2d3

I suggest two version to fix this. This is the first version which will always create the offscreen browser using the instance settings, request context and address. The instance browser settings are checked for null in the constructor.

The second solution is described in cefsharp/CefSharp/pull/1359

I'm looking forward to your feedback. Feel free to close one of the PRs if you decide to merge one.

### DIFF
--- a/CefSharp.OffScreen/ChromiumWebBrowser.cs
+++ b/CefSharp.OffScreen/ChromiumWebBrowser.cs
@@ -107,7 +107,7 @@ namespace CefSharp.OffScreen
 
             if(automaticallyCreateBrowser)
             {
-                CreateBrowser(IntPtr.Zero, address, browserSettings, requestcontext);
+                CreateBrowser(IntPtr.Zero);
             }
             
         }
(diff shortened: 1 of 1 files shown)


amaitland:
I think this is the cleanest approach. The changes in https://github.com/cefsharp/CefSharp/commit/dca52f931fcb7e0d435b2e1f46cd1d49fed5a2d3 were only from a week ago, so they've never been released.

amaitland:
Will need to be back ported into the `cefsharp/43` branch.

amaitland:
Setting milestone to `43.0.1` as that's the first release this change will appear in (even though it's merged initially into `master`).

PR #1372, opened 2015-10-26
ExecuteScriptAsync Execute method with params

amaitland:
Add override to ExecuteScriptAsync extension that takes in a method name and array of params, then generates the relevant script and executes.

### DIFF
--- a/CefSharp/WebBrowserExtensions.cs
+++ b/CefSharp/WebBrowserExtensions.cs
@@ -4,6 +4,7 @@
 
 using System;
 using System.Text;
+using System.Linq;
 using System.Threading.Tasks;
 using CefSharp.Internals;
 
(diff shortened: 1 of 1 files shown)


illfang:
It might be good to evalute if it is better to execute the function in a closure

amaitland:
> better to execute the function in a closure

What exactly did you have in mind?

amaitland:
> What exactly did you have in mind?

I should probably clarify, how would you suggest we quantify **better**?

illfang:
Sorry for the delay. I did think about reducing the side effects of executing functions with this helper method but with the current limitations for the object types, a user of this function would have to pass strange JS code to this function to lead to side effects.

CefSharp/WebBrowserExtensions.cs
```diff
@@ -183,6 +186,54 @@ public static Task<string> GetTextAsync(this IWebBrowser browser)
...
+                    {
+                        stringBuilder.Append("'");
+                    }
+                }
+
+                stringBuilder.Append(", ");
+            }
+            
+            //Remove the trailing comma
+            stringBuilder.Remove(stringBuilder.Length - 2, 2);
```
illfang: for the case `args.Length == 0` this would remove the last character of the method name and the (
e.g.
`ExecuteScriptAsync (browser, "console.log");` -> `"console.lo);"`

amaitland: I hadn't considered that scenario. I'm sure there will be someone that will try it, so best to guard against I suppose :+1:

Issue #1354, opened 2015-10-22
EvaluateScriptAsync Frame Is No Longer Available


bjarteskogoy:
I'm testing upgrading our software to use the master branch instead of 43.
In one particular browser (WPF), I always end up with this error when evaluating javascript: https://github.com/cefsharp/CefSharp/blob/master/CefSharp.BrowserSubprocess.Core/CefAppUnmanagedWrapper.cpp#L184

Any hints on what might lead to this?

amaitland:
The `frameId` may be incorrectly passed into that message. Not really a lot of information to go on.

Can you provide a working example?

LeMoussel:
@amaitland : Some idea when fix this?

bjarteskogoy:
I'll try to see if I can reproduce it. It's kind of inside something, inside something in a corner of the application. The two relevant differences from other places where we do exactly the same thing, as I can see now, is that the browser is not visible when the script is executed and that it happens in the FrameLoadEnd event.

amaitland:
> Some idea when fix this?

I can now reproduce the issue using the `URL` provided by @LeMoussel 

https://github.com/amaitland/CefSharp/commit/bdd922f8a7f34863d0fc146fa45e5835e77b6432

So hopefully it's not too hard to fix.

amaitland:
So it appears at least in the context of http://crawlbin.com/ `CefAppUnmanagedWrapper::OnContextCreated` is never called. From what I can tell the page has no javascript, so I guess `CEF` never creates a `CefV8Context` as it doesn't believe one is required.

If I add a call to `ExecuteScriptAsync`, a context is created. e.g.

```c#
browser.ExecuteScriptAsync("1 + 1");
```

The correct approach to fix this is yet unclear.

amaitland:
First attempt at a solution can be found at https://github.com/amaitland/CefSharp/commit/7d300d1c97c38f5bafb3e72796301fea47e615f9

amaitland:
Change merged into `master` and `cefsharp/43`, will wait for a little bit more feedback before releasing the upcoming `43.0.1`.

@bjarteskogoy If you get a change would you mind trying the change out?

bjarteskogoy:
I never experienced this on 43. I've built from master and the problem is gone for our case at least. (y)

amaitland:
> I never experienced this on 43.

Restructure to the code since the `43.0.0` release to allow full support for frames and popups.

amaitland:
[...]

PR #1380, opened 2015-10-28
Fix ensure context binding of root object

illfang:
This PR is an addition to the commit 00c926a4cc66d5ae5ed0a4c25c7b933c358e7224 for solving #1354

In the previous commit, a new root object will be created during the `OnProcessMessageReceived` when there has not yet been a JS context created by the browser. By evanluating JS in the `OnProcessMessageReceived` function, a context will be created and `OnContextCreated` will be called.
As the root object has already been created the following will be the case:
```
LOG(WARNING) << "A context has been created for the same browser / frame without context released called previously";
```
This will skip the `Bind` of the root object to this new context which this PR tries to solve.

### DIFF
--- a/CefSharp.BrowserSubprocess.Core/CefAppUnmanagedWrapper.cpp
+++ b/CefSharp.BrowserSubprocess.Core/CefAppUnmanagedWrapper.cpp
@@ -64,19 +64,23 @@ namespace CefSharp
         auto rootObjectWrappers = browserWrapper->JavascriptRootObjectWrappers;
         auto frameId = frame->GetIdentifier();
 
-        if (rootObjectWrappers->ContainsKey(frameId))
+        JavascriptRootObjectWrapper^ rootObject;
+        if (!rootObjectWrappers->TryGetValue(frameId, rootObject))
+        {
+            rootObject = gcnew JavascriptRootObjectWrapper(browser->GetIdentifier(), browserWrapper->BrowserProcess);
+            rootObjectWrappers->TryAdd(frameId, rootObject);
+        }
+
+        if (rootObject->IsBound)
         {
             LOG(WARNING) << "A context has been created for the same browser / frame without context released called previously";
         }
         else
         {
-            auto rootObject = gcnew JavascriptRootObjectWrapper(browser->GetIdentifier(), browserWrapper->BrowserProcess);
             if (!Object::ReferenceEquals(_javascriptRootObject, nullptr) || !Object::ReferenceEquals(_javascriptAsyncRootObject, nullptr))
             {
                 rootObject->Bind(_javascriptRootObject, _javascriptAsyncRootObject, context->GetGlobal());
             }
-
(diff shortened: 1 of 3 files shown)


CefSharp.BrowserSubprocess.Core/CefAppUnmanagedWrapper.cpp
```diff
@@ -64,19 +64,22 @@ namespace CefSharp
...
         auto frameId = frame->GetIdentifier();
 
-        if (rootObjectWrappers->ContainsKey(frameId))
+        JavascriptRootObjectWrapper^ rootObject;
+        if (!rootObjectWrappers->TryGetValue(frameId, rootObject))
+        {
+            rootObject = gcnew JavascriptRootObjectWrapper(browser->GetIdentifier(), browserWrapper->BrowserProcess);
+            rootObjectWrappers->TryAdd(frameId, rootObject);
+        }
+        if (rootObject->IsBound)
```
amaitland: Formatting - Add line between `}` and `if`

CefSharp.BrowserSubprocess.Core/JavascriptRootObjectWrapper.h
```diff
@@ -40,6 +41,10 @@ namespace CefSharp
         {
             CefSharp::Internals::JavascriptCallbackRegistry^ get();
         }
+        property bool IsBound
```
amaitland: Formatting - Add line between `}` and `property`

CefSharp.BrowserSubprocess.Core/JavascriptRootObjectWrapper.cpp
```diff
@@ -14,6 +14,11 @@ namespace CefSharp
 {
     void JavascriptRootObjectWrapper::Bind(JavascriptRootObject^ rootObject, JavascriptRootObject^ asyncRootObject, const CefRefPtr<CefV8Value>& v8Value)
     {
+        if (_isBound)
+        {
+            throw gcnew InvalidOperationException("This root object has already been bound.");
+        }
+        _isBound = true;
```
amaitland: Formatting - Add lines above and below

amaitland:
@illfang Do you have a use case for this? Or just adding support for the edge case?

amaitland:
Other than the formatting looks fine :+1:

illfang:
> Other than the formatting looks fine

Thank you. I've adjusted the formatting and updated the PR.


> Do you have a use case for this? Or just adding support for the edge case?

I've created two branches demonstrating the current problem:
```
Uncaught ReferenceError: testTime is not defined
@ undefined:1:21
```
https://github.com/illfang/CefSharp/tree/fix-ensure-context-binding-of-root-object-example

and one demonstrating the working fix:
```
991/7/23
```
https://github.com/illfang/CefSharp/tree/fix-ensure-context-binding-of-root-object-example-fixed

amaitland:
Thanks :+1: Merging into `cefsharp/43` as well.

Issue #1162, opened 2015-07-31
Fix/Add Missing Xml Documents


amaitland:
When building in release mode the compiler will kindly point out all the public places that don't have xml comments, we should add/fix these in preparation for releasing a symbols package.

amaitland:
`VC++` classes that need doc

- `CefSettings`
- `BrowserSettings`
- `RequestContextSettings`
- `RequestContext` (`IRequestContext` has documentations, copy from there)

Classes in the `CefSharp` project, need to be documented as well.

The `CEF API` is available at http://magpcss.org/ceforum/apidocs3/index-all.html (it's not always at the latest version, see top right corner).

Conventions from mapping the `CEF API`

- `CefSettings.browser_subprocess_path` becomes `CefSettings.BrowserSubProcessPath`
- `CursorType.CT_POINTER` becomes `CefCursorType.Pointer`

There are some inconsistencies with the naming of classes, some are prefixed with `Cef` and some aren't for historic reasons it's hard to change these now as it would break backwards compatibility. 

All of the `Handler` interfaces should be consistent in naming though, `CefResourceHandler` maps to `IResourceHandler`, that convention should apply to all of them.

amaitland:
Basic `API Doc` is now available at http://cefsharp.github.io/api/51.0.0/

nylu:
Hello,

is there a reason no xml documentation is included in the netcore nuget packages?

P.S. thank you for the good work!

amaitland:
> is there a reason no xml documentation is included in the netcore nuget packages?

We need `Microsoft` to resolve https://github.com/dotnet/sdk/issues/1458

PR #1399, opened 2015-11-03
Add missing XML documentation comments

Atvaark:
When I updated to *45.0.0-pre01* my not so forgiving StyleCop rules notified me of a missing parameter comment in the method *IContextMenuHandler.RunContextMenu*.

I did a quick search for some more empty XML comments and added them when necessary.

This fixes some warnings from #1162.


amaitland:
@Atvaark Awesome, thanks! :+1:

PR #1426, opened 2015-11-17
Add Support for executing a JavascriptCallback with a Struct as a param

### LINKED ISSUE #1407: Javascript callback - when passing a struct in as a parameter, the javascript function receives a null value.
I'm trying to return a simple struct to a javascript callback on a  method on a bound object.  However, in javascript, the parameter value always returns null.  When I pass in a string to the callback (like in the BoundObject.TestCallback method), everything works as expected.

Stepping through SerializeV8SimpleObject in V8Serialization.cpp, I see it getting the fieldNames for the properties on my object and setting those on subDict.  I don't follow how the parameter list gets updated to include subDict, however.  

Relevant code from SerializeV8SimpleObject;
```
                else if (type->IsValueType && !type->IsPrimitive && !type->IsEnum)
                {
                    auto fields = type->GetFields();
                    auto subDict = CefDictionaryValue::Create();
                    for (int i = 0; i < fields->Length; i++)
                    {
                        auto fieldName = fields[i]->Name;
                        auto strFieldName = StringUtils::ToNative(safe_cast<String^>(fieldName));
                        auto fieldVal = fields[i]->GetValue(obj);
                        SerializeV8SimpleObject(fieldVal, subDict, strFieldName, seen);
                    }
                } 
```

I've reproduced this issue on a fork of the CefSSharp.MinimalExample project that can be found here:

https://github.com/tgroff-cargas/CefSharp.MinimalExample

Thank you very much.

amaitland:
Resolves #1407
More the DeserializeV8Object call to within the `context->Enter` so a V8Object can be created
Change callback to use a struct as it's return type to demo the solution to this bug
Add missing list->SetDictionary call in V8Serialization.cpp

### DIFF
--- a/CefSharp.BrowserSubprocess.Core/CefAppUnmanagedWrapper.cpp
+++ b/CefSharp.BrowserSubprocess.Core/CefAppUnmanagedWrapper.cpp
@@ -251,12 +251,6 @@ namespace CefSharp
                 else
                 {
                     auto jsCallbackId = GetInt64(argList, 2);
-                    auto parameterList = argList->GetList(3);
-                    CefV8ValueList params;
-                    for (CefV8ValueList::size_type i = 0; i < parameterList->GetSize(); i++)
-                    {
-                        params.push_back(DeserializeV8Object(parameterList, static_cast<int>(i)));
-                    }
 
                     auto callbackWrapper = callbackRegistry->FindWrapper(jsCallbackId);
                     if (callbackWrapper == nullptr)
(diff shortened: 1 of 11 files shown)


amaitland:
Currently targeting `master` branch, will back port to `45` once tested.

amaitland:
More code cleanup, only the initial commit will be back ported.

amaitland:
Backported in `cefsharp/45`

amaitland:
Backported in `cefsharp/43` - not sure if there will be another `43.0.x` release, if there is this change will be included.

PR #1440, opened 2015-11-24
Implement IFrame.LoadRequest

### LINKED ISSUE #912: Feature Request - Load Request with Post Data
I have a form that when filled out and clicked it does a POST to a new window.

How do I catch and pass the new POST info to the new window?

     public bool OnBeforePopup(IWebBrowser browser, string sourceUrl, string url, ref int x, ref int y, ref int width, ref int height)
     {
            var popupw = width;
            var popuph = height;
          
            this.InvokeOnUiThreadIfRequired(() =>
            {
            
                 var popup = new XfBrowserPopup(url, popupw, popuph, sourceUrl);
                  popup.Show();

           });

     }

I am not using the standard default popups.  I am using my own custom winforms.

XfBrowserPopup is just a new form that I create a new Browser Instance and dock with that form.

Any ideas ?

amaitland:
Expose IFrame.LoadRequest
Add IFrame.CreateRequest to instantiate a new instance of the CefRequestWrapper
Was planning on using VC++ extensions methods, have yet to get that working, so for now just have a method on the IFrame interface directly.

Resolves #912

### DIFF
--- a/CefSharp.Core/Internals/CefFrameWrapper.cpp
+++ b/CefSharp.Core/Internals/CefFrameWrapper.cpp
@@ -297,7 +297,14 @@ IBrowser^ CefFrameWrapper::Browser::get()
     return _owningBrowser;
 }
 
-IRequest^ CefFrameWrapper::CreateRequest()
+IRequest^ CefFrameWrapper::CreateRequest(bool initializePostData)
 {
-    return gcnew CefRequestWrapper(CefRequest::Create());
+    auto request = CefRequest::Create();
+
+    if (initializePostData)
+    {
+        request->SetPostData(CefPostData::Create());
+    }
+
+    return gcnew CefRequestWrapper(request);
 }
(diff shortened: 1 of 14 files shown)


amaitland:
@rover886 Hopefully this will get you on track, I've only implemented the first part. You'll need to implement the missing post data pieces before this will work.

Original idea to use `VC++ Extension` methods didn't work, so will need to revisit that later.

rover886:
> You'll need to implement the missing post data pieces before this will work.

Dear @amaitland, Thank you so much for your guidance. Today I have modified some file to implement the creation of `PostData` inside `IRequest` variable after calling `browser.GetMainFrame().CreateRequest();`. I tested it with POST and GET method with some request parameters (`PostDataElement`)  and it's working :), but I am not able to write code to post `file` through `LoadRequest`. So can I create new `PR` for work this? and if possible please comment on it if I am missing something.

amaitland:
> As some part is not done, so can I create new PR for whatever I have done until, and if possible please comment on it if I am missing something.

@rover886 Create a `PR` with what you have so far and I'll take a look when I get a chance :+1:

PR #1460, opened 2015-12-02
Add IRenderProcessMessageHandler.OnContextCreated

amaitland:
Send an IPC message when OnContextCreated is called in the render process
Currently only called for the main frame
Should be possible to hook the DOMContentLoaded event

I've only tested using the sample code below, needs more testing before merging.


amaitland:
Not sure `IRequestHandler` is the correct place for this method yet. Will need to ponder that some more.

amaitland:
The relevant commit is https://github.com/amaitland/CefSharp/commit/6f272833fb8a8a1c2c602f3b4b06e0d14ccb54c2

Rest is just me cleaning up some other code.

PR #1466, opened 2015-12-03
Fixed incorrect case in CefSharp.Example/Properties/Resources.resx

perlun:
I've switched my Mac to use a case-sensitive file system now for my `~/git` folder so I happened to see this.

I guess we should re-apply this to the 45 branch as well, since we want that to be compilable under these circumstances also.


amaitland:
@perlun Thanks :+1: Of course you can merge any cleanup changes you like, no need to consult with me :smile:

perlun:
Hehe, thanks. :wink: Just following our official at-work GitHub workflow: make changes on a feature branch, ask someone to review before merging. Besides, I'm not a core developer of CefSharp right now so, it doesn't hurt to check with others. :smile:

amaitland:
> Besides, I'm not a core developer of CefSharp right now so, it doesn't hurt to check with others. :smile:

Fair enough :smile:

PR #1469, opened 2015-12-04
Implement OnFocusedNodeChanged

### LINKED ISSUE #1256: Feature Request - Implement OnFocusedNodeChanged
I need to detect when an input element gets focus (to show the soft keyboard) - like a mobile OS does. I wonder whether there is some supported way in there.
Currently I am thinking of polling for document.activeElement and comparing last one with current.

rlmcneary2:
Resolves #1256

Add support for when an element node in the DOM gets focus. Also invoked when focus passes from an element node to an non-element node or nothing.

Placed in IRequestHandler because that's where other CefRenderProcessHandler methods were placed.

### DIFF
--- a/CefSharp.BrowserSubprocess.Core/CefAppUnmanagedWrapper.cpp
+++ b/CefSharp.BrowserSubprocess.Core/CefAppUnmanagedWrapper.cpp
@@ -99,10 +99,15 @@ namespace CefSharp
 
     void CefAppUnmanagedWrapper::OnFocusedNodeChanged(CefRefPtr<CefBrowser> browser, CefRefPtr<CefFrame> frame, CefRefPtr<CefDOMNode> node)
     {
+        if (!_enableFocusedNodeChanged)
+        {
+            return;
+        }
+
         auto focusedNodeChangedMessage = CefProcessMessage::Create(kOnFocusedNodeChanged);
         auto list = focusedNodeChangedMessage->GetArgumentList();
 
-        // Neded in the browser process to get the frame.
+        // Needed in the browser process to get the frame.
         SetInt64(frame->GetIdentifier(), list, 0);
 
         // The node will be empty if an element loses focus but another one
--- a/CefSharp.Core/Internals/ClientAdapter.cpp
+++ b/CefSharp.Core/Internals/ClientAdapter.cpp
@@ -615,6 +615,21 @@ namespace CefSharp
             return handler->OnResourceResponse(_browserControl, browserWrapper, %frameWrapper, %requestWrapper, %responseWrapper);
         }
 
+        void ClientAdapter::OnResourceLoadComplete(CefRefPtr<CefBrowser> browser, CefRefPtr<CefFrame> frame, CefRefPtr<CefRequest> request, CefRefPtr<CefResponse> response, URLRequestStatus status, int64 receivedContentLength)
+        {
+            auto handler = _browserControl->RequestHandler;
+
+            if (handler != nullptr)
+            {
+                auto browserWrapper = GetBrowserWrapper(browser->GetIdentifier(), browser->IsPopup());;
+                CefFrameWrapper frameWrapper(frame);
+                CefRequestWrapper requestWrapper(request);
+                CefResponseWrapper responseWrapper(response);
+
+                handler->OnResourceLoadComplete(_browserControl, browserWrapper, %frameWrapper, %requestWrapper, %responseWrapper, (UrlRequestStatus)status, receivedContentLength);
+            }
+        }
+
         void ClientAdapter::OnProtocolExecution(CefRefPtr<CefBrowser> browser, const CefString& url, bool& allowOSExecution)
         {
             auto handler = _browserControl->RequestHandler;
@@ -996,7 +1011,7 @@ namespace CefSharp
 
             if (name == kOnFocusedNodeChanged)
             {
-                auto handler = _browserControl->RequestHandler;
+                auto handler = _browserControl->RenderProcessMessageHandler;
                 if (handler != nullptr)
                 {
                     // 0: frame ID (int 64)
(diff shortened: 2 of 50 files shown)


amaitland:
> Placed in IRequestHandler because that's where other CefRenderProcessHandler methods were placed.

I'm not sure this is the best place, I think there's probably better.

Few other notes/questions:
- `Spaces` not tabs please.
- Braces even for single line `if` statements
- In general I'm wondering if we should make this configurable, there will be some overhead.
- Is there any point having the `IDomNode` interface?

CefSharp.BrowserSubprocess.Core/CefAppUnmanagedWrapper.cpp
```diff
@@ -97,6 +97,42 @@ namespace CefSharp
         }
     };
 
+	void CefAppUnmanagedWrapper::OnFocusedNodeChanged(CefRefPtr<CefBrowser> browser, CefRefPtr<CefFrame> frame, CefRefPtr<CefDOMNode> node)
+	{
+		auto focusedNodeChangedMessage = CefProcessMessage::Create(kOnFocusedNodeChanged);
+		auto list = focusedNodeChangedMessage->GetArgumentList();
+
+		// Neded in the browser process to get the frame.
+		list->SetString(0, frame->GetName());
```
amaitland: Would prefer to send the `FramdId`, for consistency.

CefSharp/DomNode.cs
```diff
@@ -0,0 +1,88 @@
+﻿using System;
```
amaitland: Add license disclaimer please

rlmcneary2:
Thanks for the feedback.
 - Spaces - fixed
 - Braces - fixed
 - license - added
 - IDomNode - since the node information is made available to external clients I provided an interface, but I don't have really strong feelings about it since I don't think there will be alternate implementations.

I had a few other thoughts (including configuration):
 - our use case is the same as #1256, needing to display an on-screen keyboard for all the various text inputs, all we really need is the tag name and if an input its type
 - the tag names that generate a message could be configurable
 - src attributes could be data or object URIs containing large chunks of data
 - include the node value (if any) which is not currently part of the changes in this pull request

I'm leaning towards a simple on/off configuration switch for focus changed; on and all elements that get focus will generate a message that includes the tag name, all attributes, and the node value; off - the default - and no messages are passed from the render process.

rlmcneary2:
And should this be moved to an existing handler or a new one? If an existing handler any suggestions? Otherwise a new IDomHandler?

CefSharp.Core/Internals/ClientAdapter.cpp
```diff
@@ -994,31 +994,66 @@ namespace CefSharp
...
+                        // Node information was not provided.
+                        handler->OnFocusedNodeChanged(_browserControl, GetBrowserWrapper(browser->GetIdentifier(), browser->IsPopup()), %frameWrapper, nullptr);
                     }
-
-                    pendingTask->SetResult(response);
                 }
-
-                handled = true;
             }
+			else if (name == kEvaluateJavascriptResponse || name == kJavascriptCallbackResponse)
```
amaitland: Still some formatting that requires fixing.

CefSharp.Core/Internals/ClientAdapter.cpp
```diff
@@ -994,31 +994,66 @@ namespace CefSharp
...
+                        auto attributes = gcnew System::Collections::Generic::Dictionary<String^, String^>();
+                        CefDictionaryValue::KeyList keys;
+                        argAttributes->GetKeys(keys);
+                        for (auto key : keys)
+                        {
+                            attributes->Add(gcnew String(key.c_str()), gcnew String(argAttributes->GetString(key).c_str()));
+                        }
+
+                        auto node = gcnew DomNode(attributes);
+                        node->TagName = gcnew String(argList->GetString(2).c_str());
```
amaitland: Please pass in the `TagName` through the constructor and make the getter private.

CefSharp/IRequestHandler.cs
```diff
@@ -170,5 +169,14 @@ public interface IRequestHandler
...
+
+        /// <summary>
+        /// Invoked when an element in the UI gains focus (or possibly no element gains focus; i.e. an element lost focus).
+        /// </summary>
+        /// <param name="browserControl">The ChromiumWebBrowser control</param>
+        /// <param name="browser">the browser object</param>
+        /// <param name="frame">The frame object</param>
+        /// <param name="node">An object with information about the node that has focus.</param>
+        void OnFocusedNodeChanged (IWebBrowser browserControl, IBrowser browser, IFrame frame, IDomNode node);
+	}
```
amaitland: Formatting is still a little out here. Don't worry about fixing this one as It's unlikely that `OnFocusedNodeChanged` will remain as part of `IRequestHandler`

CefSharp.Core/Internals/ClientAdapter.cpp
```diff
@@ -994,31 +994,66 @@ namespace CefSharp
...
                     {
-                        response->Result = DeserializeObject(argList, 2, callbackFactory);
+                        // Node information was passed from the render process.
+                        auto argAttributes = argList->GetDictionary(3);
+                        auto attributes = gcnew System::Collections::Generic::Dictionary<String^, String^>();
+                        CefDictionaryValue::KeyList keys;
+                        argAttributes->GetKeys(keys);
+                        for (auto key : keys)
+                        {
+                            attributes->Add(gcnew String(key.c_str()), gcnew String(argAttributes->GetString(key).c_str()));
```
amaitland: Please use `StringUtls::ToClr` method for consistency.

CefSharp/IDomNode.cs
```diff
@@ -0,0 +1,48 @@
...
+//
+// Use of this source code is governed by a BSD-style license that can be found in the LICENSE file.
+
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+
+namespace CefSharp
+{
+    /// <summary>
+    /// Represents the data belonging to a node in the browser's DOM.
```
amaitland: Probably should mention that it only represents a subset of the DOM node's information.

amaitland:
> I'm leaning towards a simple on/off configuration switch for focus changed; on and all elements that get focus will generate a message that includes the tag name and all attributes; off - the default - and no messages are passed from the render process.

Sounds sensible :+1: Easy enough to add a setting. 

https://github.com/cefsharp/CefSharp/blob/master/CefSharp.Core/Internals/CefSharpApp.h#L48

> And should this be moved to an existing handler or a new one? If an existing handler any suggestions? Otherwise a new IDomHandler?

I think we probably need a new handler that encompasses #1460 as well. I'm open to suggestion on naming for a new interface.

rlmcneary2:
OnFocusNodeChanged is now configurable through CefSettings. I added a new interface IRenderProcessMessageHandler to contain both OnFocusNodeChanged and OnContextCreated.

Unfortunately it's going to be a little difficult to follow the history because I decided to branch off master of my fork and I had a different initial implementation of enabling/disabling the focus message.

amaitland:
Great, thanks :+1: Will review and merge `ASAP.

amaitland:
Merged with some minor changes see https://github.com/cefsharp/CefSharp/commit/d33c41ec5db7000a2ef715d4adb2221ec3152cb9

PR #1499, opened 2015-12-18
WPF: Change ChromiumWebBrowser's to be Control instead of UserControl

gayaK:
I think `Control` is more proper than `ContentControl` for Wpf.ChromiumWebBrowser's base class,
because we can't change Content value!
Content value must be Image control.
I suggest using default style.

### DIFF
--- a/CefSharp.Wpf.Example/Views/BrowserTabView.xaml
+++ b/CefSharp.Wpf.Example/Views/BrowserTabView.xaml
@@ -3,7 +3,8 @@
              xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
              xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
              xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
-             xmlns:cefSharp="clr-namespace:CefSharp.Wpf;assembly=CefSharp.Wpf" xmlns:local="clr-namespace:CefSharp.Wpf.Example.ViewModels"
+             xmlns:cefSharp="clr-namespace:CefSharp.Wpf;assembly=CefSharp.Wpf"
+             xmlns:local="clr-namespace:CefSharp.Wpf.Example.ViewModels"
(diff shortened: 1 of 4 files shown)


amaitland:
Interesting contribution! Happy to see the `BitmapScalingMode demo` changes merged.

> I think Control is more proper than ContentControl for Wpf.ChromiumWebBrowser's base class,
> because we can't change Content value!

What exactly is your use case?

gayaK:
Thank you.
I'd like to remove meaningless property for users as far as possible.
Users can write following code, but it doesn't have no effect for control.
```xml
<cef:ChromiumWebBrowser Content="foo" />
```

amaitland:
> Users can write following code, but it doesn't have no effect for control.

It's true they could, it would seem strange that anyone would actually do this. And I've not seen any evidence/user reports that anyone has actually been silly enough to attempt it.

I'll review your change in more detail when I have time, at this stage I'm unsure if it will be merged. Thanks for taking the time to contribute though :+1:

gayaK:
OK!
I'll send pull request only "BitmapScalingMode demo" again, because I want to separate problems.

amaitland:
Great, thanks :+1:

amaitland:
@gayaK I'll leave this open for now. We may end up coming back and using the template. Every so often people ask why is there no default context menu. Adding a template with a default implementation maybe worth while.

perlun:
@amaitland Do you remember why you were hesitant to merging this?

@gayaK Would be great if this could be rebased on top of latest changes in master, especially since some of it went in to `master` via #1540.

amaitland:
> @amaitland Do you remember why you were hesitant to merging this?

It's a sensible change, just never a pressing need, requires more in-depth testing than I ever found time for.

amaitland:
I've created #2239 that takes the place of this `PR`. Replaces the `Popup` used to show the drop down lists, needs a bit more testing, simplifies the implementation detail (no more `Mouse.Capture` required)

PR #1510, opened 2015-12-26
Add BitmapScalingMode binding and demo.

gayaK:
It's separated from #1499.
It makes Wpf.ChromiuWebBrowser to be able to change BitmapScalingMode dynamically.

### DIFF
--- a/CefSharp.Wpf.Example/Views/BrowserTabView.xaml
+++ b/CefSharp.Wpf.Example/Views/BrowserTabView.xaml
@@ -50,6 +50,7 @@
                         <Grid.RowDefinitions>
                             <RowDefinition Height="Auto" />
                             <RowDefinition Height="Auto" />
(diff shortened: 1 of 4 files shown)


amaitland:
@gayaK It looks like your using `Generic.xaml` just to set the default `BitmapScalingMode`, is there a reason we don't just set it in code?

My proposed changes are https://github.com/amaitland/CefSharp/commit/1f6dc69deeb90c66b4814c36f0aa011de92675e1

Thoughts, comments? Have I missed something?

gayaK:
Sorry for delay.
> Have I missed something?
No. Your proposed one is better than mine.

amaitland:
> Sorry for delay.

No problem :+1: 

> No. Your proposed one is better than mine.

It's based on your changes, just a small tweak :smile:

PR #1515, opened 2015-12-28
Evaluate, execute, and asynchronous events scripting examples.

### LINKED ISSUE #1482: Include JS scripts that can be executed or evaluated.
Over time I've written a few scripts used with EvaluateScriptAsync that seem to be requested fairly regularly by CefSharp users. Perhaps these could be included somewhere in the CefSharp repository for people to use and improve them as needed? I like the idea of having a "Scripts" directory with JS files in it but they could also be included directly in the documentation somewhere.

rlmcneary2:
Demonstrate how the .Net and Javascript runtime environments can interact through the use of Evaluate, Execute, and bound .Net classes. Resolves #1482.


amaitland:
@rlmcneary2 Thanks :+1: Merged with changes see https://github.com/cefsharp/CefSharp/commit/af05e56c12c14928b7e9a652b2dc789341f5ec36

(Mostly just formatting, making the code consistent with the rest of the project)

PR #1519, opened 2015-12-30
Add IRequestHandler.GetResourceResponseFilter

### LINKED ISSUE #1439: Feature Request - Add IRequestHandler.GetResourceResponseFilter
How can I get every response's content? Obiously, creating RequestHandler and using OnResourceResponse method doesn't work. All you can see in there is response headers. Is there any other way I could accomplish that? Let's say I navigate to "google.com", while getting in there, browser sends multiple requests (depending on the pages javascript). How to get content of responses to those requests?

amaitland:
Looks like the basic pass through implementation is working - have yet to append any data

I don't pretend to understand the whole thing as yet. I think the `API` can likely be simplified (once it's properly understood).

Relevant `CEF` commit https://bitbucket.org/chromiumembedded/cef/commits/456ea5a

Resolves #1439


amaitland:
Implement `FindReplaceResponseFilter` to get a better understanding of what's going on.
https://bitbucket.org/chromiumembedded/cef/src/456ea5a19800a4cc05d0ce7f0c38111a7e73e18b/tests/cefclient/browser/response_filter_test.cc?fileviewer=file-view-default

amaitland:
`API` is now slightly simplified, it's still on the complicated side, I guess that's just the nature of the beast. At some point later, it's probably worth providing some default filters, once this has matured a little bit more. Probably an appending filter would be handy (just append a string to the end of the data stream, useful for appending javascript to an existing file).

amaitland:
Have added a quick `AppendResponseFilter` example (figured it was easier to do it now whilst everything was fresh).

https://github.com/cefsharp/CefSharp/commit/679fad6a891c9454ab1202005fde22c134da5f86

Issue #1502, opened 2015-12-21
Cleanup ManagedCefBrowserAdapter - Remove duplicate code


amaitland:
A number of methods in ManagedCefBrowserAdapter are now exposed by `IBrowserHost`, so the duplicate code can be removed.

amaitland:
Resolved with #1542

PR #1542, opened 2016-01-12
Cleanup ManagedCefBrowserAdapater + Remove Duplicate Code

amaitland:
- Move Drag methods into IBrowserHost
- Move SendKeyEvent into IBrowserHost (overload that takes OS message params).
- Unravel some of the mess that was how IBrowser instances were access between the different layers

Resovles #1502 

Needs more testing before it can be merged.


juepiezhongren:
@amaitland ,hellow
could u give me an example of how to use sendKeyEvent?
I am totally frustrated for i have tried numerous way to implement it, but never succeeded!
Beg u man!

PR #1562, opened 2016-01-28
Make JS bindings work with subprocesses hosting multiple browser instances

cefsharp-ms:
When the popup functionality is used, more than one browser instance can be hosted in the same subprocess. As the bound JS objects are not stored on a per-browser basis they won't work properly in this scenario. I modified the storage and binding mechanism a bit to solve this.

### DIFF
--- a/CefSharp.BrowserSubprocess.Core/CefAppUnmanagedWrapper.cpp
+++ b/CefSharp.BrowserSubprocess.Core/CefAppUnmanagedWrapper.cpp
@@ -60,41 +60,11 @@ namespace CefSharp
 
     void CefAppUnmanagedWrapper::OnContextCreated(CefRefPtr<CefBrowser> browser, CefRefPtr<CefFrame> frame, CefRefPtr<CefV8Context> context)
     {
-        //Send a message to the browser processing signaling that OnContextCreated has been called
-        //only param is the FrameId. Currently an IPC message is only sent for the main frame - will see
-        //how viable this solution is and if it's worth expanding to sub/child frames.
-        if (frame->IsMain())
-        {
-            auto contextCreatedMessage = CefProcessMessage::Create(kOnContextCreatedRequest);
+        auto contextCreatedMessage = CefProcessMessage::Create(kOnContextCreatedRequest);
 
-            SetInt64(contextCreatedMessage->GetArgumentList(), 0, frame->GetIdentifier());
+        SetInt64(contextCreatedMessage->GetArgumentList(), 0, frame->GetIdentifier());
 
-            browser->SendProcessMessage(CefProcessId::PID_BROWSER, contextCreatedMessage);
-        }
-
-        auto browserWrapper = FindBrowserWrapper(browser->GetIdentifier(), true);
-
-        auto rootObjectWrappers = browserWrapper->JavascriptRootObjectWrappers;
-        auto frameId = frame->GetIdentifier();
-
-        JavascriptRootObjectWrapper^ rootObject;
-        if (!rootObjectWrappers->TryGetValue(frameId, rootObject))
-        {
-            rootObject = gcnew JavascriptRootObjectWrapper(browser->GetIdentifier(), browserWrapper->BrowserProcess);
-            rootObjectWrappers->TryAdd(frameId, rootObject);
-        }
-
-        if (rootObject->IsBound)
-        {
-            LOG(WARNING) << "A context has been created for the same browser / frame without context released called previously";
-        }
-        else
-        {
-            if (!Object::ReferenceEquals(_javascriptRootObject, nullptr) || !Object::ReferenceEquals(_javascriptAsyncRootObject, nullptr))
-            {
-                rootObject->Bind(_javascriptRootObject, _javascriptAsyncRootObject, context->GetGlobal());
-            }
-        }
+        browser->SendProcessMessage(CefProcessId::PID_BROWSER, contextCreatedMessage);
     };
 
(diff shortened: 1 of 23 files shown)


CefSharp.BrowserSubprocess.Core/CefAppUnmanagedWrapper.h
```diff
@@ -28,11 +28,15 @@ namespace CefSharp
...
 
         // The serialized registered object data waiting to be used (only contains methods and bound async).
-        gcroot<JavascriptRootObject^> _javascriptAsyncRootObject;
+        gcroot<ConcurrentDictionary<int, JavascriptRootObject^>^> _javascriptAsyncRootObject;
 
         // The serialized registered object data waiting to be used.
-        gcroot<JavascriptRootObject^> _javascriptRootObject;
+        gcroot<ConcurrentDictionary<int, JavascriptRootObject^>^> _javascriptRootObject;
 
+        static JavascriptRootObject^ Update(int key, JavascriptRootObject^ value)
```
amaitland: Where is this method used?

CefSharp.BrowserSubprocess.Core/CefAppUnmanagedWrapper.h
```diff
@@ -44,6 +48,8 @@ namespace CefSharp
             _extensions = gcnew List<CefExtension^>();
             _schemes = schemes;
             _focusedNodeChangedEnabled = enableFocusedNodeChanged;
+            _javascriptRootObject = gcnew ConcurrentDictionary<int, JavascriptRootObject^>();
```
amaitland: Variables need to be renamed e.g. `_javascriptRootObjects` as it's not a dictionary.

amaitland:
@cefsharp-ms Are you able to add a test case to one of the examples?

amaitland:
Thinking about this a little more, are you sure this is the best solution? My original idea was that once the render process had been initialized the root objects could be reused. Are you using standard popups and having problems?

cefsharp-ms:
I fixed the naming and leftover method, I'm sorry, I missed those.

> @cefsharp-ms Are you able to add a test case to one of the examples?

Yep, I'll put together something.

> Thinking about this a little more, are you sure this is the best solution? My original idea was that once the render process had been initialized the root objects could be reused. Are you using standard popups and having problems?

I wasn't clear about the exact use case here. I'm using custom popups and I wanted to have separate objects bound to the popups. And this caused a few problems, hence my change. However it just hit me that for standard popups the original PR would have changed the previous behavior, so I modified it a bit to have the bound objects from the 'main' browser as default values for all popups.

amaitland:
> I fixed the naming and leftover method, I'm sorry, I missed those.

Thanks :+1: 

> Yep, I'll put together something.

Much appreciated.

> I wasn't clear about the exact use case here. I'm using custom popups and I wanted to have separate objects bound to the popups. And this caused a few problems, hence my change.

That makes more sense. Thanks for clarifying.

Will take it for a spin when I get a few minutes (hopefully next week sometime).

CefSharp.BrowserSubprocess.Core/CefAppUnmanagedWrapper.cpp
```diff
@@ -385,8 +400,18 @@ namespace CefSharp
...
         {
-            _javascriptAsyncRootObject = DeserializeJsRootObject(argList, 0);
-            _javascriptRootObject = DeserializeJsRootObject(argList, 1);
+            auto asyncRoot = DeserializeJsRootObject(argList, 0);
+            auto syncRoot = DeserializeJsRootObject(argList, 1);
+            _javascriptAsyncRootObjects->default[browser->GetIdentifier()] = asyncRoot;
+            _javascriptRootObjects->default[browser->GetIdentifier()] = syncRoot;
+
+            //let's store the non-popup browser's root objects to have default values for the popups
+            if (!browser->IsPopup())
```
amaitland: Would it be cleaner to move the assignment for the popup objects into the branch of the `if` statement then do a simple `HasKey` check above? Remove the need for https://github.com/cefsharp/CefSharp/pull/1562/files#diff-a27ca20dcc4f76e2dde0c785dd2abb9cR50

Overall I think we can make the intent a little cleaner to follow. Open to other suggestions

cefsharp-ms:
Hi,
sorry for the delay on this. I've added an example demonstrating the problem.

Also it seems like something has changed the timing for some events since the upgrade to 49 so I needed to apply another approach to the binding, now it basically pauses the first resource load until the bindings are ready. Otherwise in case of the popups the context creation was too early, and the bound objects for the popups were not sent over before that.

I know it's a pretty big change, but please review it and let's talk about it.

amaitland:
> I know it's a pretty big change, but please review it and let's talk about it.

I've had a quick look (might be quite some time before I can go over this properly). In my opinion the only way to justify making such drastic changes would be to continue the refactor and remove `process-per-tab`.

https://github.com/cefsharp/CefSharp/blob/cefsharp/49/CefSharp.Core/CefSettings.h#L38

CefSharp.BrowserSubprocess.Core/CefAppUnmanagedWrapper.cpp
amaitland: Ideally I'd still like to simplify this.

amaitland:
You can now bind objects by name in different web pages, should also apply for `frames` and `popups` though I haven't tested yet. See https://github.com/cefsharp/CefSharp/issues/2246 for details.

Issue #1594, opened 2016-02-16
Closing popup window UnhandledException


KennyKay:
I'm currently using WinForm CefSharp 47.0.2 and when I close a popup, there is a high chance of crashing the browser.

The exception message that I am receiving is:
ClientAdapter::CefSharp.Internals.ClientAdapter.OnResourceLoadComplete couldn't find IBrowser entry! Please contact CefSharp development.

Stacktrace:
```
   at CefSharp.Internals.ClientAdapter.GetBrowserWrapper(ClientAdapter* , Int32 browserId, Boolean isPopup)
   at CefSharp.Internals.ClientAdapter.OnResourceLoadComplete(ClientAdapter* , CefRefPtr<CefBrowser>* browser, CefRefPtr<CefFrame>* frame, CefRefPtr<CefRequest>* request, CefRefPtr<CefResponse>* response, cef_urlrequest_status_t status, Int64 receivedContentLength)
```

Thanks,

turgayselvi:
I have noticed same issue CefSharp 43 also. It doesn't crash for every page in a popup. 

To reproduce the issue:

1- ) Go to: www.yahoo.com
2-) Type 'NBA' on search panel.
3-) Click NBA.com on the list. It will open as popup.
4-) Close popup window. 
5-) It will crash.

After popup browser is deleted from browser dictionary. GetBrowserWrapper is called. This causes an issue. But not for every page. That is strange.


ClientAdapter.cpp


   IBrowser^ ClientAdapter::GetBrowserWrapper(int browserId, bool isPopup)
        {
            if (_browserControl->HasParent)
            {
                return _browserAdapter->GetBrowser();
            }

            if(isPopup)
            {
                IBrowser^ browserWrapper;
                if (_popupBrowsers->TryGetValue(browserId, browserWrapper))
                {
                    return browserWrapper;
                }

                auto stackFrame = gcnew StackFrame(1);
                auto callingMethodName = stackFrame->GetMethod()->Name;

            **    ThrowUnknownPopupBrowser(gcnew String(L"ClientAdapter::" + callingMethodName));**
            }

            return _browserAdapter->GetBrowser();
        }

amaitland:
It appears to be a timing issue, would depend somewhat on the web page and network connection. If you close the popup before all it's resources have completed downloading then an exception will be thrown.

amaitland:
[...]

PR #1596, opened 2016-02-17
Change method parameter value to prevent popup window crashing.

turgayselvi:
set **returnNullIfBrowserNotFound** field value as "**true**" for **GetBrowserWrapper** method in **OnBeforeResourceLoad** in order to prevent crashing popup window, when popup window is closed.


amaitland:
Thanks :+1:

amaitland:
Fixes #1594

PR #1597, opened 2016-02-18
Rewrite IResourceHandler - Expose all methods

### LINKED ISSUE #1592: Feature Request - Rewrite IResourceHandler
So it's that time of year again, it's time to rewrite `IResourceHandler`.

Basic `Response Filtering` implementation wraps the underlying `buffer` in an `UnmanagedMemoryStream`, which makes it possible to expose the total raw, bare metal `API` now.

- Refactor `IResourceHandler` to match `CefResourceHandler`, including `CanGetCookie` and `CanSetCookie` which are currently hard cododed.
- Rewrite `ResourceHandler` so that it includes abstract methods that match the currently `IResourceHandler` implementation, for those that don't need access to the raw power.
- Currently `ResourceHandler` implementation can be renamed, the static methods can stay on the `abstract` version.

For upgrading users it should be quite simple
- Implement `ResourceHandler` instead of `IResourceHandler`
- Add `override` to the methods.

http://magpcss.org/ceforum/apidocs3/projects/%28default%29/CefResourceHandler.html

amaitland:
Expose all methods of CefResourceHandler - Resolves #1592
Previously we were passing in the frame, request wrappers which hadn't been used, so just keep things simple and remove the code.
Current users should extend ResourceHandler instead of IResourceHandler and add the override keyword

### DIFF
--- a/CefSharp.Core/Internals/ClientAdapter.cpp
+++ b/CefSharp.Core/Internals/ClientAdapter.cpp
@@ -689,17 +689,13 @@ namespace CefSharp
             }
 
             auto browserWrapper = GetBrowserWrapper(browser->GetIdentifier(), browser->IsPopup(), false);
-            auto frameWrapper = gcnew CefFrameWrapper(frame);
-            auto requestWrapper = gcnew CefRequestWrapper(request);
+            CefFrameWrapper frameWrapper(frame);
+            CefRequestWrapper requestWrapper(request);
 
-            auto handler = factory->GetResourceHandler(_browserControl, browserWrapper, frameWrapper, requestWrapper);
+            auto handler = factory->GetResourceHandler(_browserControl, browserWrapper, %frameWrapper, %requestWrapper);
 
             if (handler == nullptr)
             {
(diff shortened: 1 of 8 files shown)


amaitland:
See https://github.com/cefsharp/CefSharp/blob/560fa2b867fdfaceb2d6ed88e5a112a0ad5a1aef/CefSharp.Example/FlashResourceHandler.cs for a simplified implementation of `ResourceHandler`

HTD:
I noticed a stream passed to my ResourceHandler extension is not disposed. Disposing the original stream is not automatic even when implementing IResourceHandler, but I can dispose the stream after detecting it has been copied (number of bytes written are equal to original stream length). How can I dispose original stream using ResourceHandler class extension?

amaitland:
I've made `IResourceHandler` implement `IDisposable`, see https://github.com/cefsharp/CefSharp/commit/3061becb3f4a44f6e5556d9eafdf269a56a40da3

Change will be in the next release (not sure when that will be yet).

In future, please ask your questions on `Gitter`,

Issue #880, opened 2015-03-15
Feature Request - Publish packages to symbolsource.org


amaitland:
Create additional nuget packages and push them to http://symbolsource.org/
http://docs.nuget.org/create/creating-and-publishing-a-symbol-package

Background `Stackoverflow` topic http://stackoverflow.com/questions/13488280/best-practices-with-nuget-debug-or-release

amaitland:
http://blog.davidebbo.com/2011/04/easy-way-to-publish-nuget-packages-with.html

amaitland:
Details about `MyGet` symbols feed see http://docs.myget.org/docs/reference/symbols

The `Pre-Authenticated` symbols feed is available from https://www.myget.org/F/cefsharp/auth/49b91d8f-ee25-4906-8624-6f67d9423c63/symbols/

Hasn't been fully tested yet, anyone that's interested feel free to check it out though.

amaitland:
https://www.symbolsource.org/Public/Wiki/Using

amaitland:
Using Symbols loaded from `MyGet.org` appears to be working quite nicely. 


To load symbols for the `BrowserSubProcess`, add the following command line option. When the message box opens, attach the debugger to the matching process. (Should be named `Chromium Renderer`). Short while and the symbols should be loaded.

```c#
settings.CefCommandLineArgs.Add("renderer-startup-dialog", "1");
```

Additional download the `Release Symbols` that match the `CEF` version your using from https://cefbuilds.com/ Then extract `libcef.dll.pdb` into the executing folder (same directory as `libcef.dll`).

Another helpful debugging tip can be found at https://github.com/cefsharp/CefSharp/issues/944#issuecomment-92369129

jankurianski:
Nice one @amaitland :+1: 

I've added your tips into a new section in the Trouble Shooting wiki guide: https://github.com/cefsharp/CefSharp/wiki/Trouble-Shooting#unmanaged-crashes-when-the-process-dies

amaitland:
> I've added your tips into a new section in the Trouble Shooting wiki guide:

Cool, thanks :smile: 

[...]

PR #1610, opened 2016-02-26
Publish Symbol Packages

amaitland:
Add pdb files to nuspec - they're only included when the -symbol flag is used
Add src files to nuspec - they're only included when the -symbol flag is used
Update appveyor.yml to push symbol packages to the MyGet feed
Add -symbol flag to build.ps1 so the symbol packages are generated
WinForms project not outputs pdb files
WinForms release builds are not optimized - probably should have been already

Unfortunately this will need to be merged into `master` to test fully. The packages build locally, so hopefully they'll appears on the `MyGet` feed shrotly.


amaitland:
Hopefully resolves #880

amaitland:
`CefSharp` now uses `GitLink` to publish source code see #1680 for details

PR #1628, opened 2016-03-13
Added try-catches to CookieVisitor.cpp to avoid crashes on invalid va…

YaiEf:
…lues for DateTime.

This is the simplest possible fix. It ignores cases where the user might want to know what the invalid dates were. If that is important, I would suggest creating a new class (CefDateTime?) that directly corresponds to the values in CefCookie, and then create DateTime on demand in Cookie.cs.

### DIFF
--- a/CefSharp.Core/Internals/CookieVisitor.cpp
+++ b/CefSharp.Core/Internals/CookieVisitor.cpp
@@ -21,39 +21,54 @@ namespace CefSharp
             cookie->Secure = cefCookie.secure == 1;
             cookie->HttpOnly = cefCookie.httponly == 1;
 
-            if (cefCookie.has_expires)
-            {
-                cookie->Expires = DateTime(
-                    cefCookie.expires.year,
-                    cefCookie.expires.month,
-                    cefCookie.expires.day_of_month,
-                    cefCookie.expires.hour,
-                    cefCookie.expires.minute,
-                    cefCookie.expires.second,
-                    cefCookie.expires.millisecond
-                );
-            }
+			try {
+				if (cefCookie.has_expires)
+				{
+					cookie->Expires = DateTime(
+						cefCookie.expires.year,
+						cefCookie.expires.month,
+						cefCookie.expires.day_of_month,
+						cefCookie.expires.hour,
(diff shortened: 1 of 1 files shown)


amaitland:
> I would suggest creating a new class (CefDateTime?) that directly corresponds to the values in CefCookie, and then create DateTime on demand in Cookie.cs.

If you feel so inclined you can :+1:

amaitland:
@YaiEf In general, spaces instead of tabs, you can review your `PR` above, you'll see the formatting is out.

CefSharp.Core/Internals/CookieVisitor.cpp
```diff
@@ -21,39 +21,54 @@ namespace CefSharp
...
-                    cefCookie.expires.year,
-                    cefCookie.expires.month,
-                    cefCookie.expires.day_of_month,
-                    cefCookie.expires.hour,
-                    cefCookie.expires.minute,
-                    cefCookie.expires.second,
-                    cefCookie.expires.millisecond
-                );
-            }
+			try {
```
amaitland: Braces on new lines please.

CefSharp.Core/Internals/CookieVisitor.cpp
```diff
@@ -21,39 +21,54 @@ namespace CefSharp
...
+						cefCookie.expires.day_of_month,
+						cefCookie.expires.hour,
+						cefCookie.expires.minute,
+						cefCookie.expires.second,
+						cefCookie.expires.millisecond
+						);
+				}
+			}
+			catch (Exception^ ex){
+				// ignore invalid date as we have no way to represent it
```
amaitland: Could probably set to `MinValue` to represent an invalid `DateTime` for now.

YaiEf:
I have updated it so it hopefully conforms now :)

amaitland:
Thanks :+1:

zhyzhy782:
Yo! 

I've got that nice stuff  for  you, it is  awesome. Just take a look at it <http://favor.crowandraven.org/e7yv/11>

All the best, zhyzhy782

PR #1633, opened 2016-03-17
MethodRunnerQueue target default scheduler

### LINKED ISSUE #1630: MethodRunnerQueue can lock UI thread
Upgrading from v33 to v47 (WPF) the application froze on initializing the browser control.

Turns out that the MethodRunnerQueue ran on the UI-Thread and therefore locked up the UI Thread.

`Task.Factory.StartNew` is used to run the `ConsumeTasks()`-loop. -> https://github.com/cefsharp/CefSharp/blob/master/CefSharp/Internals/MethodRunnerQueue.cs

Depending on where you come from it is then possible that this loop runs on the UI-Thread. Which should not happen ever.

Some information related to this ["Start New is Dangerous"](http://blog.stephencleary.com/2013/08/startnew-is-dangerous.html)

`Task.Factory.StartNew` should not be used here. Instead another method of starting the task async which can never hit the UI thread.

We changed it to this overload `TaskFactory.StartNew-Methode (Action, CancellationToken, TaskCreationOptions, TaskScheduler)` (https://msdn.microsoft.com/de-de/library/dd781658(v=vs.110).aspx) handing over Default-Scheduler

dga711:
Resolves #1630

### DIFF
--- a/CefSharp/Internals/MethodRunnerQueue.cs
+++ b/CefSharp/Internals/MethodRunnerQueue.cs
@@ -37,7 +37,7 @@ namespace CefSharp.Internals
                 if (!running)
                 {
                     cancellationTokenSource = new CancellationTokenSource();
-                    Task.Factory.StartNew(ConsumeTasks, TaskCreationOptions.LongRunning);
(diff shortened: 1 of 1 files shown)


amaitland:
Nice one, thanks :+1:

PR #1644, opened 2016-03-24
Add paramArray support to Javascript Binding

ChrisBardsley:
This fix allow you to use a params array for a parameter in your C# object that is bound to your javascript functions.

### DIFF
--- a/CefSharp.Example/Resources/Home.html
+++ b/CefSharp.Example/Resources/Home.html
@@ -351,14 +351,6 @@
                         objects.
                     </p>
                     <h3>Hooks by which you can modify and/or override certain features of the web browsing</h3>
-                    <p>
-                        A fix to allow params to be used by the .NET methods called by the javascript<br />
-                        <button onclick="boundEvent.methodWithParams('test', 'hello-world')">With 1 Params</button>
-                        <button onclick="boundEvent.methodWithParams('test', 'hello-world', 'chris was here')">With 2 Params</button>
-                        <button onclick="boundEvent.methodWithParams('test')">With no Params</button>
-                        <button onclick="boundEvent.methodWithoutParams('test', 'hello')">Normal Method</button>
-                        <button onclick="boundEvent.methodWithoutAnything()">Normal Method No Params</button>
-                    </p>
                 </div>
                 <div class="bs-docs-section">
                     <div class="page-header">
--- a/CefSharp.Example/Resources/Home.html
+++ b/CefSharp.Example/Resources/Home.html
@@ -351,6 +351,14 @@
                         objects.
                     </p>
                     <h3>Hooks by which you can modify and/or override certain features of the web browsing</h3>
+                    <p>
+                        A fix to allow params to be used by the .NET methods called by the javascript<br />
+                        <button onclick="boundEvent.methodWithParams('test', 'hello-world')">With 1 Params</button>
+                        <button onclick="boundEvent.methodWithParams('test', 'hello-world', 'chris was here')">With 2 Params</button>
+                        <button onclick="boundEvent.methodWithParams('test')">With no Params</button>
+                        <button onclick="boundEvent.methodWithoutParams('test', 'hello')">Normal Method</button>
+                        <button onclick="boundEvent.methodWithoutAnything()">Normal Method No Params</button>
+                    </p>
                 </div>
                 <div class="bs-docs-section">
                     <div class="page-header">
--- a/CefSharp.Example/ScriptedMethodsBoundObject.cs
+++ b/CefSharp.Example/ScriptedMethodsBoundObject.cs
@@ -36,26 +36,5 @@ namespace CefSharp.Example
                 EventArrived(eventName, eventData);
             }
         }
-
-        /// <summary>
-        /// Demonstrates the use of params as an argument in a bound object
-        /// </summary>
-        /// <param name="name">Dummy Argument</param>
-        /// <param name="args">Params Argument</param>
-        public void MethodWithParams(string name, params object[] args)
-        {
-            var test = args.ElementAtOrDefault(0);
-            var test2 = args.ElementAtOrDefault(1);
-        }
-
-        public void MethodWithoutParams(string name, string arg2)
-        {
-            var test = arg2;
-        }
-
-        public void methodWithoutAnything()
-        {
-            var test = "";
-        }
     }
 }
--- a/CefSharp.Example/ScriptedMethodsBoundObject.cs
+++ b/CefSharp.Example/ScriptedMethodsBoundObject.cs
@@ -2,6 +2,7 @@
 //
 // Use of this source code is governed by a BSD-style license that can be found in the LICENSE file.
 using System;
+using System.Linq;
 
 namespace CefSharp.Example
 {
@@ -34,6 +35,27 @@ namespace CefSharp.Example
             {
                 EventArrived(eventName, eventData);
             }
+        }
+
+        /// <summary>
+        /// Demonstrates the use of params as an argument in a bound object
+        /// </summary>
+        /// <param name="name">Dummy Argument</param>
+        /// <param name="args">Params Argument</param>
+        public void MethodWithParams(string name, params object[] args)
+        {
+            var test = args.ElementAtOrDefault(0);
+            var test2 = args.ElementAtOrDefault(1);
+        }
+
+        public void MethodWithoutParams(string name, string arg2)
+        {
+            var test = arg2;
+        }
+
+        public void methodWithoutAnything()
+        {
+            var test = "";
         }
     }
 }
--- a/CefSharp/Internals/JavascriptMethod.cs
+++ b/CefSharp/Internals/JavascriptMethod.cs
@@ -3,6 +3,7 @@
 // Use of this source code is governed by a BSD-style license that can be found in the LICENSE file.
 
 using System;
+using System.Collections.Generic;
 using System.Runtime.Serialization;
 
 namespace CefSharp.Internals
@@ -32,6 +33,9 @@ namespace CefSharp.Internals
         /// </summary>
         [DataMember]
         public string JavascriptName { get; set; }
+
+        [DataMember]
+        public List<MethodParameter> MethodParameters { get; set; }
 
         /// <summary>
         /// Number of Params this function exepects
--- a/CefSharp/Internals/JavascriptMethod.cs
+++ b/CefSharp/Internals/JavascriptMethod.cs
@@ -34,7 +34,6 @@ namespace CefSharp.Internals
         [DataMember]
         public string JavascriptName { get; set; }
 
-        [DataMember]
         public List<MethodParameter> MethodParameters { get; set; }
 
         /// <summary>
--- a/CefSharp/Internals/JavascriptObjectRepository.cs
+++ b/CefSharp/Internals/JavascriptObjectRepository.cs
@@ -104,15 +104,33 @@ namespace CefSharp.Internals
 
             try
             {
-                // Do we have enough arguments? Add Type.Missing for any that we don't have incase of optional params
+                //Fix for C# paramArray and refactor on arguments mismatch 03/22/2016
                 var missingParams = method.ParameterCount - parameters.Length;
-                if (missingParams > 0)
+                if (missingParams > 0 || method.MethodParameters.Any(t => t.IsParamArray))
                 {
-                    var paramList = new List<object>(parameters);
+                    var paramList = new List<object>();
 
-                    for (var i = 0; i < missingParams; i++)
+                    for (int i = 0; i < method.MethodParameters.Count; i++)
                     {
-                        paramList.Add(Type.Missing);
+                        object jsParam = parameters.ElementAtOrDefault(i);
+                        if (jsParam == null && !method.MethodParameters[i].IsParamArray)
+                        {
+                            paramList.Add(Type.Missing);
+                        }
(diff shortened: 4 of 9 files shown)


CefSharp.Example/Resources/Home.html
```diff
@@ -351,6 +351,14 @@ <h3 id="features-javascript-integration">JavaScript integration</h3>
                         objects.
                     </p>
                     <h3>Hooks by which you can modify and/or override certain features of the web browsing</h3>
+                    <p>
+                        A fix to allow params to be used by the .NET methods called by the javascript<br />
```
amaitland: This should be in `BindingTest.html`. Also having ` A fix to allow params to be used by the .NET methods called by the javascript` is not really relevant to someone using `CefSharp`. Simple description of what your demoing would be better :+1:

CefSharp.Example/ScriptedMethodsBoundObject.cs
```diff
@@ -35,5 +36,26 @@ public void RaiseEvent(string eventName, object eventData = null)
                 EventArrived(eventName, eventData);
             }
         }
+
+        /// <summary>
+        /// Demonstrates the use of params as an argument in a bound object
+        /// </summary>
+        /// <param name="name">Dummy Argument</param>
+        /// <param name="args">Params Argument</param>
+        public void MethodWithParams(string name, params object[] args)
```
amaitland: These should return a result, then output that result when they're called from `BindingTest.html`. Something similar to https://github.com/cefsharp/CefSharp/blob/cefsharp/47/CefSharp.Example/Resources/BindingTest.html#L210 would be ideal.

CefSharp/Internals/JavascriptMethod.cs
```diff
@@ -33,6 +34,9 @@ public class JavascriptMethod
         [DataMember]
         public string JavascriptName { get; set; }
 
+        [DataMember]
```
amaitland: Is `DataMember` required, doesn't seem like it's used by the `WCF` client?

CefSharp/Internals/JavascriptObjectRepository.cs
```diff
@@ -104,15 +104,33 @@ public bool TryCallMethod(long objectId, string name, object[] parameters, out o
 
             try
             {
-                // Do we have enough arguments? Add Type.Missing for any that we don't have incase of optional params
+                //Fix for C# paramArray and refactor on arguments mismatch 03/22/2016
```
amaitland: Again, not sure this comment is useful, seems like something you'd add to your own project, not so much an open source one? 

If your keen on referencing an `issue`, or `pull request`, you'd either open an issue to reference this or add a subsiquent commit one the `pull request` has been opened, like `Add paramArray support #1644`.

By all means add more comments, even comment what each branch of the if statement does.

CefSharp/Internals/JavascriptObjectRepository.cs
```diff
@@ -104,15 +104,33 @@ public bool TryCallMethod(long objectId, string name, object[] parameters, out o
...
                 var missingParams = method.ParameterCount - parameters.Length;
-                if (missingParams > 0)
+                if (missingParams > 0 || method.MethodParameters.Any(t => t.IsParamArray))
```
amaitland: Unless I'm mistaken you can only have one `params` per method, and it's the last param, so this can likely be simplified using `Last`

CefSharp/Internals/JavascriptObjectRepository.cs
```diff
@@ -104,15 +104,33 @@ public bool TryCallMethod(long objectId, string name, object[] parameters, out o
...
                 {
-                    var paramList = new List<object>(parameters);
+                    var paramList = new List<object>();
 
-                    for (var i = 0; i < missingParams; i++)
+                    for (int i = 0; i < method.MethodParameters.Count; i++)
```
amaitland: Prefer `var` over `int` for consistency

CefSharp/Internals/JavascriptObjectRepository.cs
```diff
@@ -104,15 +104,33 @@ public bool TryCallMethod(long objectId, string name, object[] parameters, out o
...
+                        }
+                        else if (method.MethodParameters[i].IsParamArray)
+                        {
+                            List<object> convertedParams = new List<object>();
+                            for (int s = i; s < parameters.Count(); s++)
+                            {
+                                convertedParams.Add(parameters[s]);
+                            }
+                            paramList.Add(convertedParams.ToArray());
+                            break;
```
amaitland: Is the `break` necessary?

CefSharp/Internals/JavascriptObjectRepository.cs
```diff
@@ -104,15 +104,33 @@ public bool TryCallMethod(long objectId, string name, object[] parameters, out o
...
             try
             {
-                // Do we have enough arguments? Add Type.Missing for any that we don't have incase of optional params
+                //Fix for C# paramArray and refactor on arguments mismatch 03/22/2016
                 var missingParams = method.ParameterCount - parameters.Length;
-                if (missingParams > 0)
+                if (missingParams > 0 || method.MethodParameters.Any(t => t.IsParamArray))
                 {
-                    var paramList = new List<object>(parameters);
+                    var paramList = new List<object>();
```
amaitland: As we now know the size, let's initialize to exactly the size we required.

CefSharp/Internals/JavascriptObjectRepository.cs
```diff
@@ -104,15 +104,33 @@ public bool TryCallMethod(long objectId, string name, object[] parameters, out o
...
-                        paramList.Add(Type.Missing);
+                        object jsParam = parameters.ElementAtOrDefault(i);
+                        if (jsParam == null && !method.MethodParameters[i].IsParamArray)
+                        {
+                            paramList.Add(Type.Missing);
+                        }
+                        else if (method.MethodParameters[i].IsParamArray)
+                        {
+                            List<object> convertedParams = new List<object>();
+                            for (int s = i; s < parameters.Count(); s++)
```
amaitland: `Length` is preferable to `Count()`

amaitland: Some form of comment here is required, just to describe what's happening. e..g `Add param args to list`...I'm sure you'll think of something better :smile:

CefSharp/Internals/JavascriptObjectRepository.cs
```diff
@@ -287,6 +305,14 @@ private static JavascriptMethod CreateJavaScriptMethod(MethodInfo methodInfo, bo
             jsMethod.JavascriptName = GetJavascriptName(methodInfo.Name, camelCaseJavascriptNames);
             jsMethod.Function = methodInfo.Invoke;
             jsMethod.ParameterCount = methodInfo.GetParameters().Length;
+            jsMethod.MethodParameters = methodInfo.GetParameters()
+                .Select(t => new MethodParameter()
+                { Name = t.Name,
```
amaitland: `Name` on new line.

CefSharp/Internals/JavascriptObjectRepository.cs
```diff
@@ -287,6 +305,14 @@ private static JavascriptMethod CreateJavaScriptMethod(MethodInfo methodInfo, bo
...
+                    ParameterType = t.ParameterType,
```
amaitland: Are these properties used? seems redundant to store them if they're not, or did you have other plans?

CefSharp/Internals/MethodParameter.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Runtime.Serialization;
+using System.Text;
+
+namespace CefSharp.Internals
+{
+    [DataContract]
```
amaitland: Don't think `DataContract` is required here as we don't need to reference this in the `render process`.

CefSharp/Internals/MethodParameter.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System;
```
amaitland: - Add license disclaimer
- Remove redundant using statements
- Make sure formatting is `spaces` (not `tabs` as `VS` typically uses as the default).

NuGet.config
```diff
@@ -5,5 +5,6 @@
     </activePackageSource>
   <packageSources>
     <add key="cefsharp-myget" value="https://www.myget.org/F/cefsharp/" />
+    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" protocolVersion="3" />
```
amaitland: Please revert this as it's not relevant in the context of this `PR`.

CefSharp/Internals/MethodParameter.cs
```diff
@@ -0,0 +1,20 @@
...
+    [DataContract]
+    public class MethodParameter
+    {
+        public String Name { get; set; }
+
+        public Type ParameterType { get; set; }
+
+        public Boolean IsParamArray { get; set; }
+
+        public Int32 Position { get; set; }
```
amaitland: `int` for consistency.

amaitland: Though this will probably be removed, so likely safe to ignore.

CefSharp/Internals/MethodParameter.cs
```diff
@@ -0,0 +1,20 @@
...
+using System.Linq;
+using System.Runtime.Serialization;
+using System.Text;
+
+namespace CefSharp.Internals
+{
+    [DataContract]
+    public class MethodParameter
+    {
+        public String Name { get; set; }
```
amaitland: use `string` in lower case as the default around here :+1:

CefSharp/Internals/MethodParameter.cs
```diff
@@ -0,0 +1,20 @@
...
+
+        public Type ParameterType { get; set; }
+
+        public Boolean IsParamArray { get; set; }
```
amaitland: `bool` here please :smile:

amaitland:
@ChrisBardsley Welcome, and thanks for the contribution :smile: Comments inline, by all means ask any questions for clarification.

I'd also like to rename this to something simple like `Add paramArray support to Javascript binding`, for a few reasons

- Fix implies it was supported and didn't work, it just plain wasn't supported
- I believe at least `VB.Net` has support for `params`, so `C#` is a little too specific here.

amaitland:
@ChrisBardsley Any further comment on this? Initially I was planning on this being in the `49.0.0` release, as it stands it'll be bumped to `51`

CefSharp/Internals/JavascriptObjectRepository.cs
ChrisBardsley: Changed to LastOrDefault(IsParamArray) != null

CefSharp/Internals/JavascriptObjectRepository.cs
ChrisBardsley: Constancy is key

CefSharp/Internals/JavascriptObjectRepository.cs
ChrisBardsley: Considering that paramArrays will always be last, I don't believe so. Thanks for catching that :)

CefSharp/Internals/JavascriptObjectRepository.cs
amaitland: Sounds reasonable :+1:

CefSharp/Internals/JavascriptObjectRepository.cs
amaitland: Can you add a comment, so it's clear what's going on?

CefSharp/Internals/JavascriptObjectRepository.cs
ChrisBardsley: MethodParameters is of type List and does not support Length.  I can change the MethodParameters property to an array which will allow for Length, but I will need to do a cast to an array where the property is populated, which may or may not require more processing, but TBH, i'm not sure on that one.

CefSharp/Internals/JavascriptObjectRepository.cs
ChrisBardsley: Thanks for catching that.  I had planned on implementing those properties, but never did.  I removed those.

CefSharp/Internals/JavascriptObjectRepository.cs
amaitland: My mistake, should use the `Count`, property, there's no need to call the method (which is an `Extension` method from memory, has a small overhead).

CefSharp/Internals/JavascriptObjectRepository.cs
ChrisBardsley: Absolutely

amaitland: Thanks :+1:

amaitland:
@ChrisBardsley Merged with changes, see https://github.com/cefsharp/CefSharp/commit/669404f74cf61d8bc5b32c2ed86b87c71adfac34 (comments welcome)

I'll review one more time in the next few days before merging into the `cefsharp/49` branch.

Thanks for the contribution :+1:

ray007:
There seems to be a problem with managing missing/optional parameters.
in `JavascriptObjectRepository.cs` line 123 my null-Parameter from javascript gets matched in replaced by a `MissingParameter`.

amaitland:
@ray007 Please open a new issue. Ideally submitting a `PR` with a proposed solution.

PR #1658, opened 2016-04-14
Issue 1594 - Updated calls to GetBrowserWrapper()

edgchen1:
Changed the returnNullIfBrowserNotFound parameter to true for some calls to GetBrowserWrapper() to prevent exceptions when closing a popup window.


edgchen1:
@amaitland The change you suggested seems to work, I don't see a crash anymore when closing popup windows. Thanks!

amaitland:
Great, this needs to be targeted against master. I will then merge into the `cefsharp/49` branch. Unsure if there will be another `47` release.

PR #1659, opened 2016-04-15
Issue 1594 - Updated calls to GetBrowserWrapper()

edgchen1:
Changed the returnNullIfBrowserNotFound parameter to true for some calls to GetBrowserWrapper() to prevent exceptions when closing a popup window.

Replacing pull request 1658 - targeting against master.

### DIFF
--- a/CefSharp.Core/Internals/ClientAdapter.cpp
+++ b/CefSharp.Core/Internals/ClientAdapter.cpp
@@ -619,7 +619,7 @@ namespace CefSharp
                 return false;
             }
                 
-            auto browserWrapper = GetBrowserWrapper(browser->GetIdentifier(), browser->IsPopup(), false);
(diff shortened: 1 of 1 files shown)


edgchen1:
@amaitland 
This is the same change I made on the 47 branch.

FYI, I was not as successful in testing it on master. Not sure if this is due to issues in my environment or something else. I was trying to open then close a few popup windows again from CefSharp.Wpf.Example and noticed a few differences from the 47 version:
- popup windows sometimes don't display any content
- clicking "Show DevTools" context menu option doesn't seem to do anything

amaitland:
Thanks, if you can reproduce the problem your seeing then please create a new use. Please make sure you use the issue template and provide as much detail as possible, thanks.

amaitland:
This `PR` is a follow up to #1594, changes have also been merged into `cefsharp/49`

Issue #1582, opened 2016-02-07
Globalization bug - Javascript cannot invoke a method because JavascriptObjectRepository::GetJavascriptName uses ToLower()


akjana35:
#### Issue Summary
What is the expected output? What do you see instead?
Say a method named IsSupported is exported for javascript execution. On a Turkish Windows machine, GetJavascriptName returns isSupported with dotted unicode i. As a result javascript cannot invoke this method.

#### Steps to Reproduce
1. Get a windows machine with Turkish language pack. 
2. Export a javascript method where name starts with i.
3. Try to invoke the same method from javascript.

Does this problem also occur in the `Cef` TestApp from http://cefbuilds.com? didn't try

#### Technical details
* CefSharp Version: nuget - master - latest commit: 43.0.1 (stable)
* Operating System: Windows 8/10 Turkish (tr-TR)
* x86/x64: Both 
* WinForms/WPF/OffScreen: Winforms

amaitland:
Please submit a `PR` with your proposed change :+1:

amaitland:
Short term register your objects with `camelCaseJavascriptNames = false`

akjana35:
well we fixed it locally with char.ToLowerInvariant(). Doesn't require me to change the casing on the exported methods in javascript. It will be some time before I get back to this.

amaitland:
Please submit a `PR` when you have time.

amaitland:
@akjana35 Are you planning on submitting a `PR` for this? Whilst it's a trivial fix, It would be nice if **you** submitted a fix. After all you've asked quite a few questions, be nice if you could give something back to the project :+1: Thanks.

akjana35:
I'll send the PR sometime this week. I have permissions now to do bug fixes.

amaitland:
> I'll send the PR sometime this week.

Great, please target your `PR` against, `master`.

amaitland:
Resolved with #1660

PR #1660, opened 2016-04-18
Fix Issue #1582, Use ToLowerInvariant while converting Javascript Function names to camelCase

akjana35:
Fix Issue #1582, Use ToLowerInvariant while converting Javascript Function names to camelCase


amaitland:
@akjana35 Thanks, this will be included in the `49.0.0` release.

PR #1731, opened 2016-07-02
Fix bool, number culture, and string escaping in ExecuteScriptAsync with parameters

### LINKED ISSUE #1728: ExecuteScriptAsync with parameters does not handle booleans
The version and architecture info don't matter, the bug is in the core and still present in master branch.

49.0.0-pre2 from NuGet

x86

Win 7

WinForms

Using [WebBrowserExtensions.ExecuteScriptAsync(this IWebBrowser browser, string methodName, params object[] args)](https://github.com/cefsharp/CefSharp/blob/master/CefSharp/WebBrowserExtensions.cs#L194) to pass a boolean variable in the *args* converts it to *"True"* or *"False"* which are both truthy strings, therefore passing *false* will always result in *true* on JS side.

This would probably need a little bit of special casing, unfortunately adding the bool type to [WebBrowserExtensions.numberTypes](https://github.com/cefsharp/CefSharp/blob/master/CefSharp/WebBrowserExtensions.cs#L15) is not enough because Boolean.ToString capitalizes the first letter and JavaScript doesn't accept True/False despite being case-insensitive. So either special casing booleans or something like:

```
if (!numberTypes.Contains(obj.GetType()))
{
    stringBuilder.Append("'").Append(args[i].ToString()).Append("'");
}
else
{
    stringBuilder.Append(args[i].ToString().ToLowerInvariant());
}
```

chylex:
Tested by running:
```
ExecuteScriptAsync("window.runtest", "hello", "quote'abc", -3000000, 5.5, 8F, 7.66D, true, false, null);
```
window.runtest prints out all arguments and their type.

Old output on most european systems, note booleans and decimal numbers:
```
hello | string
(syntax error due to unescaped quote)
-3000000 | number
5 | number
5 | number
8 | number
7 | number
66 | number
True | string
False | string
null | object
```
New output:
```
hello | string
quote'abc | string
-3000000 | number
5.5 | number
8 | number
7.66 | number
true | boolean
false | boolean
null | object
```
Closes #1728.

### DIFF
--- a/CefSharp/WebBrowserExtensions.cs
+++ b/CefSharp/WebBrowserExtensions.cs
@@ -6,6 +6,7 @@ using System;
 using System.Text;
 using System.Linq;
 using System.Threading.Tasks;
+using System.Globalization;
 using CefSharp.Internals;
 
 namespace CefSharp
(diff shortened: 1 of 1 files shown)


CefSharp/WebBrowserExtensions.cs
```diff
@@ -202,24 +203,21 @@ public static void ExecuteScriptAsync(this IWebBrowser browser, string methodNam
...
-                            stringBuilder.Append("'");
-                        }
-
-                        stringBuilder.Append(args[i].ToString());
-
-                        if (encapsulateInSingleQuotes)
-                        {
-                            stringBuilder.Append("'");
-                        }
+                        stringBuilder.Append("'").Append(args[i].ToString().Replace("'","\\'")).Append("'");
```
amaitland: This in my opinion is getting a little hard to read, please reformat for readability :+1:

CefSharp/WebBrowserExtensions.cs
```diff
@@ -202,24 +203,21 @@ public static void ExecuteScriptAsync(this IWebBrowser browser, string methodNam
...
                 {
                     var obj = args[i];
-                    if(obj == null)
+                    if (obj == null)
                     {
                         stringBuilder.Append("null");
                     }
+                    else if (numberTypes.Contains(obj.GetType()))
+                    {
+                        stringBuilder.Append(string.Format(CultureInfo.InvariantCulture, "{0}", args[i]));
```
amaitland: What about `Convert.ToString`? `string.format` whilst I'm sure works seems like a strange choice here (also not particularly efficient when run in a tight loop).

chylex:
Updated.

amaitland:
> Updated.

Thanks :+1: Hopefully find time to have a look during the coming week.

PR #1734, opened 2016-07-04
Offscreen popup (dropdown list) merge with the background browser.

valentin-deschaintre:
Created to have the possibility to merge the dropdown list (select tag) popup menu with the full browser to get one single bitmap containing the whole page.

There are two commits :
-  4078663 : Simply improve what can be overriden and implement a few functions for the popup informations.  This allows a better local implementation of the InvokeRenderAsync function.

-  b20ddb7 : Implementation of a InvokeRenderAsync which does the blending of popup/background image. Added a configuration to know between split and merge of the background image and popup. Also modified what happens on popup opening/closing to match the rest of the modifications.

First review contributor : ejerome

No issue opened as seen with amaitland on Gitter.

### DIFF
--- a/CefSharp.OffScreen/ChromiumWebBrowser.cs
+++ b/CefSharp.OffScreen/ChromiumWebBrowser.cs
@@ -738,8 +738,7 @@ namespace CefSharp.OffScreen
                 }
                 else if (BackGroundImage == null)
                 {
-                    throw new InvalidOperationException(string.Format("Background image was empty. Should not happen in {0} mode.",
-                        nameof(PopupBlending.Blend)));
+                    throw new InvalidOperationException("Background image was empty. Should not happen in Blend mode.");
                 }
             }
             TriggerNewScreenshot();
--- a/CefSharp.OffScreen/ChromiumWebBrowser.cs
+++ b/CefSharp.OffScreen/ChromiumWebBrowser.cs
@@ -4,6 +4,7 @@
 
 using System;
 using System.Drawing;
+using System.Drawing.Imaging;
 using System.Threading.Tasks;
 using CefSharp.Internals;
 
@@ -97,6 +98,13 @@ namespace CefSharp.OffScreen
         /// </value>
         public bool PopupJustOpened { get; protected set; }
         /// <summary>
+        /// Gets or sets a value indicating whether the popup is open.
+        /// </summary>
+        /// <value>
+        /// <c>true</c> if popup is opened; otherwise, <c>false</c>.
+        /// </value>
+        public bool PopupOpened { get; protected set; }
+        /// <summary>
         /// A flag that indicates whether the state of the control currently supports the GoForward action (true) or not (false).
         /// </summary>
         /// <value><c>true</c> if this instance can go forward; otherwise, <c>false</c>.</value>
@@ -114,10 +122,6 @@ namespace CefSharp.OffScreen
         /// <value>The request context.</value>
         public RequestContext RequestContext { get; private set; }
 
-        /// <summary>
-        /// The blending congiguration for the popup and background.
-        /// </summary>
-        public PopupBlending BlendingConfiguration { get; set; }
         /// <summary>
         /// Implement <see cref="IJsDialogHandler" /> and assign to handle events related to JavaScript Dialogs.
         /// </summary>
@@ -297,12 +301,7 @@ namespace CefSharp.OffScreen
         /// <summary>
         /// The popup Bitmap.
         /// </summary>
-        public Bitmap PopupImage;
-
-        /// <summary>
-        /// The Background Bitmap.
-        /// </summary>
-        public Bitmap BackGroundImage;
+        public Bitmap Popup { get; set; }
 
         /// <summary>
         /// Create a new OffScreen Chromium Browser
@@ -314,8 +313,7 @@ namespace CefSharp.OffScreen
         /// <param name="blendPopup">Should the popup be blended in the background in the rendering</param>
         /// <exception cref="System.InvalidOperationException">Cef::Initialize() failed</exception>
         public ChromiumWebBrowser(string address = "", BrowserSettings browserSettings = null,
-            RequestContext requestContext = null, bool automaticallyCreateBrowser = true,
-            PopupBlending blendPopup = PopupBlending.Separate)
+            RequestContext requestContext = null, bool automaticallyCreateBrowser = true)
         {
             if (!Cef.IsInitialized && !Cef.Initialize())
             {
@@ -340,7 +338,6 @@ namespace CefSharp.OffScreen
 
             popupPosition = new Point();
             popupSize = new Size();
-            BlendingConfiguration = blendPopup;
         }
 
         /// <summary>
@@ -413,18 +410,18 @@ namespace CefSharp.OffScreen
         /// Gets the size of the popup.
         /// </summary>
         /// <param name="size">Size of the popup.</param>
-        public void GetPopupSize(ref Size size)
+        public Size GetPopupSize()
         {
-            size = popupSize;
+            return popupSize;
         }
 
         /// <summary>
         /// Gets the popup position.
         /// </summary>
         /// <param name="popupPos">The popup position.</param>
-        public void GetPopupPosition(ref Point popupPos)
+        public Point GetPopupPosition()
         {
-            popupPos = popupPosition;
+            return popupPosition;
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


amaitland:
@valentin-deschaintre Great :+1: This `PR` is third in the queue for review.

CefSharp.OffScreen/ChromiumWebBrowser.cs
```diff
@@ -273,15 +285,37 @@ public class ChromiumWebBrowser : IRenderWebBrowser
...
+
+        /// <summary>
+        ///  Size of the popup.
+        /// </summary>
+        private Size popupSize;
+
+        /// <summary>
+        /// The popup Bitmap.
+        /// </summary>
+        public Bitmap PopupImage;
```
amaitland: Missing `get; set;`

CefSharp.OffScreen/ChromiumWebBrowser.cs
```diff
@@ -373,11 +410,29 @@ protected virtual void Dispose(bool disposing)
         }
 
         /// <summary>
+        /// Gets the size of the popup.
+        /// </summary>
+        /// <param name="size">Size of the popup.</param>
+        public void GetPopupSize(ref Size size)
```
amaitland: `out` would be the correct modifier here. I assume your trying to avoid copying the `Struct`? Is this absolutely necessary? Are their other `.Net` frameworks that apply this concept?

CefSharp.OffScreen/ChromiumWebBrowser.cs
```diff
@@ -826,5 +961,24 @@ void IWebBrowserInternal.SetTooltipText(string tooltipText)
...
             TooltipText = tooltipText;
         }
+
+        /// <summary>
+        /// Merges the popup bitmap in the background bitmap.
+        /// </summary>
+        /// <returns>The merged bitmap, size of the background bitmap</returns>
+        private Bitmap MergeBackGroundPopupBitmaps()
+        {
+            System.Drawing.Bitmap myBitmap = new System.Drawing.Bitmap(BackGroundImage.Width,
```
amaitland: Remove redundant namespace qualifiers please.

CefSharp.OffScreen/ChromiumWebBrowser.cs
```diff
@@ -614,20 +678,77 @@ void IRenderWebBrowser.InvokeRenderAsync(BitmapInfo bitmapInfo)
...
+                    if (Bitmap != null)
+                    {
+                        Bitmap.Dispose();
+                        Bitmap = null;
+                    }
+
+                    Bitmap = gdiBitmapInfo.CreateBitmap();
                 }
+            }
+            else if (BlendingConfiguration == PopupBlending.Blend)
```
amaitland: What about moving this to `ScreenshotOrNull` and adding a `PopupBlending` as a param? Greatly simplify code, no need to store the `BackGroundImage`, just check if their is a popup open and merge on demand.

Performance would be a concern in your current implementation as every frame rendered would be `blended`, which is likely unnecessary, just the frames the user chooses to capture require blending.

Comments?

CefSharp.OffScreen/ChromiumWebBrowser.cs
```diff
@@ -663,6 +784,16 @@ bool IRenderWebBrowser.StartDragging(IDragData dragData, DragOperationsMask mask
         /// <param name="show">if set to <c>true</c> [show].</param>
         void IRenderWebBrowser.SetPopupIsOpen(bool show)
         {
+            if (!show && PopupImage != null)
```
amaitland: Preferable if the `true` condition comes first. In this case just set `PopupJustOpened = show` and perform your other checks as they are now, reduces code and branching.

CefSharp.OffScreen/ChromiumWebBrowser.cs
```diff
@@ -614,20 +678,77 @@ void IRenderWebBrowser.InvokeRenderAsync(BitmapInfo bitmapInfo)
...
+                else if (PopupImage == null)
+                {
+                    Bitmap = (Bitmap)BackGroundImage.Clone();
+                }
+                else if (BackGroundImage == null)
+                {
+                    throw new InvalidOperationException("Background image was empty. Should not happen in Blend mode.");
+                }
             }
+            TriggerNewScreenshot();
```
amaitland: Seems kinda pointless to have a method that simply fires an event handler?

CefSharp.OffScreen/ChromiumWebBrowser.cs
```diff
@@ -614,20 +678,77 @@ void IRenderWebBrowser.InvokeRenderAsync(BitmapInfo bitmapInfo)
...
+                {
+                    Bitmap = MergeBackGroundPopupBitmaps();
+                }
+                else if (PopupImage == null)
+                {
+                    Bitmap = (Bitmap)BackGroundImage.Clone();
+                }
+                else if (BackGroundImage == null)
+                {
+                    throw new InvalidOperationException("Background image was empty. Should not happen in Blend mode.");
```
amaitland: It's not safe to throw an exception here as it will bubble into the `unmanaged` code and crash hard.

amaitland:
@valentin-deschaintre In general, the concept of combining the `popup` makes sense :+1: As I've said, I think performance of your current implementation would be a big concern,. in general it can be greatly simplified I believe. Let me know your thoughts.

CefSharp.OffScreen/ChromiumWebBrowser.cs
valentin-deschaintre: Size and Point are struct, which make them complicated to use as auto properties, I'm not sure making them properties is useful ? On the bitmap they are missing though, I'll fix it right now.

CefSharp.OffScreen/ChromiumWebBrowser.cs
valentin-deschaintre: It is probably not necessary since we don't want to be able to modify them outside the class. I'll change it to classic getters and see how it behaves.

CefSharp.OffScreen/ChromiumWebBrowser.cs
valentin-deschaintre: Method that would override InvokeRenderAsync are not able to trigger the new screenshot event (since C# only allows to fire an event from its class). I thought it could be missing, so I created this method, but maybe I misunderstood something about this event ?

CefSharp.OffScreen/ChromiumWebBrowser.cs
valentin-deschaintre: This would be more effective, but I think the one bitmap we should save still is the background, since in case of popup, "Bitmap" is set to the popup image only. That would probably allow us to keep only one image, and as you said only merge when really needed. I'll try it right now.

CefSharp.OffScreen/ChromiumWebBrowser.cs
amaitland: Hmm, yeah I suppose auto property wouldn't work, `private` with a `getter` then.

CefSharp.OffScreen/ChromiumWebBrowser.cs
amaitland: `Background` is not really a great name here, just `bitmap` and `popup` would be fine. For simplicity, it's just a matter of storing `popup`, then writing `popup` onto `bitmap` when a screen shot is requested.

CefSharp.OffScreen/ChromiumWebBrowser.cs
amaitland: I guess, I'm still not convinced that it's required. You'd just implement your own event with custom properties, just like implementing any other `.Net` class.

valentin-deschaintre: OK, it was just to simplify the override, but i'll remove it :).

valentin-deschaintre:
I just pushed a commit taking your advices into account, in my test environment it works just like before, probably faster though ;).

CefSharp.OffScreen/ChromiumWebBrowser.cs
amaitland: Just move the call into `IRenderWebBrowser.InvokeRenderAsync`, so it's always fired :+1:

amaitland:
> I just pushed a commit taking your advices into account,

Great, thanks :+1: 

> probably faster though ;).

Just less memory copies when your not taking a screen shot :+1: 

Few subsequent changes I'd like to make, will do those after merge and post a link to the relevant commit.

valentin-deschaintre:
Great, thanks :).

amaitland:
@valentin-deschaintre Follow up changes are https://github.com/cefsharp/CefSharp/commit/a9661a73eb56e65c6a07aba3a200a9de93df71a8

I've removed `PopupJustOpened`, is it actually required? If you can test with your use case and let me know, thanks :+1:

valentin-deschaintre:
Yes, it was needed because if you open a popup, close it (through clicking elsewhere) and reopen the same popup, cef seems to believe there is nothing new to draw, and so no new popup is generated.

amaitland:
If that's the case that should be resolved at a lower level.

amaitland:
If you have an example you can contribute to facilitate testing that would be handy.

valentin-deschaintre:
The way I display and interact with the offline rendering is in a proprietary library, which makes it complicated to release. I can explain in details the behavior inside the InvokeRenderAsync in the case I'm describing though.

amaitland:
> I can explain in details the behavior inside the InvokeRenderAsync in the case I'm describing though.

I'm more interested in updating the `OffScreen.Example` project to demo/test the feature.

valentin-deschaintre:
Ok, i'll try to do that.

PR #1738, opened 2016-07-07
Add overloads for EvaluateScriptAsync to support methodName and args

PrinceOfEgypt:
Added two overloads for EvaluateScriptAsync to provide the option to pass in methodName and args instead of the string representation of the Javascript. Two overloads were needed because EvaluateScriptAsync already used an optional parameter, so it wouldn't be possible to mix both optional parameters and params in the same signature without forcing the callers to always use named parameters.


amaitland:
@PrinceOfEgypt Thanks, will try to look at this sometime next week. In the mean time if you could fix up the minor type `teh` = `the` that would be great :+1:

PrinceOfEgypt:
@amaitland done, typo has been corrected

Issue #1714, opened 2016-06-17
Feature Request - Add AnyCPU Support


amaitland:
Doing some research on another topic I believe I've come up with a method of making `AnyCPU` a reality without introducing maintainable code to the framework.

With https://github.com/cefsharp/CefSharp/commit/60a1e615fcd1b4211951201fb2a5f45ab2147eca it's now possible to target `AnyCPU`, it's important to note that **it won't just magically work out of the box** like `x86` and `x64` will. Attempting to build using `AnyCPU` will throw an error directing you to this issue. You must add `<CefSharpAnyCpuSupport>true</CefSharpAnyCpuSupport>` to the first `<PropertyGroup>` in your project (e.g. `.csproj` file) and pick one of the two options below.

### Option 1

This option effectively makes your `AnyCPU` program always run as `x86`, this is the **simplest** option.

- Add `<probing privatePath="x86"/>` to your `app.config` as outlined in https://msdn.microsoft.com/en-us/library/4191fzwb.aspx
- Set `Prefer 32bit` in your project properties see http://blogs.microsoft.co.il/sasha/2012/04/04/what-anycpu-really-means-as-of-net-45-and-visual-studio-11/ for some background
- Set `settings.BrowserSubprocessPath = @"x86\CefSharp.BrowserSubprocess.exe";` when calling `Cef.Initialize` example below.

```c#
[STAThread]
public static void Main()
{
	var settings = new CefSettings();
	settings.BrowserSubprocessPath = @"x86\CefSharp.BrowserSubprocess.exe";

	Cef.Initialize(settings, shutdownOnProcessExit: false, performDependencyCheck: false);

	var browser = new BrowserForm();
	Application.Run(browser);
}
```

Example `probingPath`
```
<runtime>
	<assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
		<probing privatePath="x86"/>
	</assemblyBinding>
</runtime>
```

### Option 2

Add a dependency resolver, this is more complicated and needs to be hooked up before any calls to classes in the `CefSharp.*` namespaces. Here is one method of doing this, I'll add some helper classes in the future. It's important that `LoadApp` is not in-lined, so the calls to `CefSharp` are delayed long enough to hookup the `Assembly Resolver`

```c#
[STAThread]
public static void Main()
{
	AppDomain.CurrentDomain.AssemblyResolve += Resolver;

	LoadApp();
}

[MethodImpl(MethodImplOptions.NoInlining)]
[...]

PR #1753, opened 2016-07-27
Configure directory where msbuild will copy CefSharp files

intrueder:
During build process msbuild copies CefSharp files to `$(TargetDir)`
Sometimes it is not suitable, and instead of override `$(TargetDir)` globally another property has been introduced. It's name is `$(CefSharpTargetDir)`
It is possible to override this property in `csproj` file
```xml
<PropertyGroup>
  <CefSharpTargetDir>C:\MyProjectOutDir\Cef</CefSharpTargetDir>
</PropertyGroup>
```

If this property should depend on `$(TargetDir)` don't forget to override property after importing `Microsoft.CSharp.targets` (`$(TargetDir)` is defined there)
```xml
<Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
<PropertyGroup>
  <CefSharpTargetDir>$(TargetDir)\Cef</CefSharpTargetDir>
</PropertyGroup>
```

Also, `$(TargetDir)` is used by default if `$(CefSharpTargetDir)` was not specified in `csproj` file

### DIFF
--- a/NuGet/CefSharp.Common.targets
+++ b/NuGet/CefSharp.Common.targets
@@ -4,20 +4,24 @@
     <Error Text="$(MSBuildThisFileName) will work out of the box if you specify platform (x86 / x64). For AnyCPU Support see https://github.com/cefsharp/CefSharp/issues/1714" />
   </Target>
 
+  <PropertyGroup>
+    <CefSharpTargetDir Condition=" '$(CefSharpTargetDir)' == '' ">$(TargetDir)</CefSharpTargetDir>
+  </PropertyGroup>
+
   <Target Name="CefSharpCopyLibs86" BeforeTargets="AfterBuild" Condition="'$(Platform)' == 'x86'">
     <ItemGroup>
       <CefSharpBinaries Include="$(MSBuildThisFileDirectory)..\CefSharp\x86\*.*" />
     </ItemGroup>
 
     <Message Importance="high" Text="Copying cef.redist x86 binaries" />
-    <Copy SourceFiles="@(CefBinaries32)" DestinationFolder="$(TargetDir)" SkipUnchangedFiles="true" />
+    <Copy SourceFiles="@(CefBinaries32)" DestinationFolder="$(CefSharpTargetDir)" SkipUnchangedFiles="true" />
     <Message Importance="high" Text="Copying cef.redist x86 pak files" />
-    <Copy SourceFiles="@(CefPakFiles32)" DestinationFolder="$(TargetDir)" SkipUnchangedFiles="true" />
+    <Copy SourceFiles="@(CefPakFiles32)" DestinationFolder="$(CefSharpTargetDir)" SkipUnchangedFiles="true" />
     <Message Importance="high" Text="Copying cef.redist x86 locales " />
-    <Copy SourceFiles="@(CefLocales32)" DestinationFolder="$(TargetDir)\locales" SkipUnchangedFiles="true" />
+    <Copy SourceFiles="@(CefLocales32)" DestinationFolder="$(CefSharpTargetDir)\locales" SkipUnchangedFiles="true" />
 
-    <Message Importance="high" Text="-- CefSharp.Common.targets: Copying files from $(MSBuildThisFileDirectory)..\CefSharp\x86 to $(TargetDir)" />
-    <Copy SourceFiles="@(CefSharpBinaries)" DestinationFolder="$(TargetDir)" SkipUnchangedFiles="true" />
+    <Message Importance="high" Text="-- CefSharp.Common.targets: Copying files from $(MSBuildThisFileDirectory)..\CefSharp\x86 to $(CefSharpTargetDir)" />
+    <Copy SourceFiles="@(CefSharpBinaries)" DestinationFolder="$(CefSharpTargetDir)" SkipUnchangedFiles="true" />
   </Target>
 
   <Target Name="CefSharpCopyLibsWin32" BeforeTargets="AfterBuild" Condition="'$(Platform)' == 'Win32'">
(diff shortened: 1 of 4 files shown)


amaitland:
@intrueder Thanks for the contribution! I'm currently working on other things at the moment, will probably take at least a few weeks before this is reviewed.

amaitland:
@intrueder I've had a quick look over your `PR` (still needs a full review). How do you plan on documenting this new feature? Simply copying to another folder isn't enough to resolve the `unmanaged` dependencies as I'm sure your aware.

intrueder:
This feature should be used in conjunction with AnyCPU support ( https://github.com/cefsharp/CefSharp/issues/1714 ) code sample.
Actually AnyCPU support can be documented and `CefSharpTargetDir` property can be as a remark.

amaitland:
> This feature should be used in conjunction with AnyCPU support ( #1714 ) code sample.

It's not limited to `AnyCPU` support though.......

> Actually AnyCPU support can be documented

I have no plans to document `AnyCPU` support past what is already included in #1714 

> CefSharpTargetDir property can be as a remark.

Please write some documentation and I'll consider it for inclusion in #1714

intrueder:
Complete code sample. (I've modified original sample a bit, becasue `AppDomain.CurrentDomain.SetupInformation.ApplicationBase` does not work for me)

**.csproj**
```xml
<PropertyGroup>
  <CefSharpTargetDir>C:\MyProjectOutDir\Cef</CefSharpTargetDir>
</PropertyGroup>
```

**.cs**
```cs
public partial class App : Application
{
    public App()
    {
        //Add Custom assembly resolver
        AppDomain.CurrentDomain.AssemblyResolve += Resolver;

        //Any CefSharp references have to be in another method with NonInlining
        // attribute so the assembly rolver has time to do it's thing.
        InitializeCefSharp();
	}

    [MethodImpl(MethodImplOptions.NoInlining)]
    private static void InitializeCefSharp()
    {
        var settings = new CefSettings();
        settings.EnableInternalPdfViewerOffScreen();

        // Set BrowserSubProcessPath based on app bitness at runtime
        settings.BrowserSubprocessPath = GetArchitectureSpecificPath("CefSharp.BrowserSubprocess.exe");

        // Make sure you set performDependencyCheck false
        Cef.Initialize(settings, shutdownOnProcessExit: false, performDependencyCheck: false);
    }

	// Will attempt to load missing assembly from either x86 or x64 subdir
	// Required by CefSharp to load the unmanaged dependencies when running using AnyCPU
	private static Assembly Resolver(object sender, ResolveEventArgs args)
	{
		if (args.Name.StartsWith("CefSharp"))
		{
			string assemblyName = args.Name.Split(new[] { ',' }, 2)[0] + ".dll";
			string archSpecificPath = GetArchitectureSpecificPath(assemblyName);

			if (File.Exists(archSpecificPath))
			{
				return Assembly.LoadFile(archSpecificPath);
			}
		}

		return null;
	}

	private static string GetAssemblyDirectory()
	{
		string codeBase = Assembly.GetExecutingAssembly().CodeBase;
		UriBuilder uri = new UriBuilder(codeBase);
		string path = Uri.UnescapeDataString(uri.Path);
		return Path.GetDirectoryName(path);
	}

	private static string GetArchitectureSpecificPath(string fileName)
	{
		// subfolder is the same as $(CefSharpTargetDir) points to
		string cefSubfolder = "Cef";
		return Path.Combine(GetAssemblyDirectory(), cefSubfolder, Environment.Is64BitProcess ? "x64" : "x86", fileName);
	}
}
```

amaitland:
> AppDomain.CurrentDomain.SetupInformation.ApplicationBase does not work for me)

What problems are you seeing using `ApplicationBase`?

amaitland:
This will likely be included as part of `53.0.0`, should be merged before the first `-pre` release is built.

intrueder:
> What problems are you seeing using `ApplicationBase`?

It is probably a special case..
That's what I have:

1. Unmanaged MyApplication.exe located in C:\ProgramFiles\MyApp
2. Additional libs/plugins in subfolder C:\ProgramFiles\MyApp\Libs. They are loaded via LoadLibrary
3. Managed library ChromiumHost.dll located in C:\ProgramFiles\MyApp\Libs. It is the only who uses CefSharp.

and now, to avoid putting CefSharp files directly into C:\ProgramFiles\MyApp\Libs, I want to put them into subfolder.

ApplicationBase points to MyApplication.exe path, i.e. C:\ProgramFiles\MyApp.
I need a folder, that will be relative to assembly where CefSharp is used (ChromiumHost.dll -> C:\ProgramFiles\MyApp\Libs), `Assembly.GetExecutingAssembly().CodeBase` helps with this.

amaitland:
> Unmanaged MyApplication.exe located in C:\ProgramFiles\MyApp

If your just using straight `C++`, why not use `CEF` directly?

intrueder:
It is not `C++` actually :) and we are about to migrate all the old code to `.Net`
so, we can't use `CEF` directly

zcanann:
Out of curiosity, does that mean you guys are shifting to a pure C#/.NET solution?

amaitland:
> Out of curiosity, does that mean you guys are shifting to a pure C#/.NET solution?

If your referring to `CefSharp`, then the answer is no. There are already pure `.Net` implementations out there.

syqjyh:
i use CefSharp to create a wpf dialog to open a url ,and this dialog is embedded in excel. our product need call excelAddIn.dll and work on anyCPU platform . so i download the cef demo project from https://github.com/cefsharp/CefSharp.MinimalExample/tree/demo/anycpu .Now i encountered a problem that i just replace the url in MainWindow.xaml with my URL, and it works well, but after i finished all coding and dependency related according to the demo project , our project still can't work and throw a exception (An unhandled exception of type 'System.IO.FileNotFoundException' occurred in Unknown Module.
Additional information: Could not load file or assembly 'CefSharp, Version=63.0.0.0, Culture=neutral, PublicKeyToken=40c4b6fc221f4138' or one of its dependencies. The system cannot find the file specified.) but i have added the CefSharp and related dependency to the related directory. i don't know why. can someone help me. thanks. and the MIMICAddIn.txt is our product entrance csproj file and i need create the wpf dialog in the csproj Morningstar.Commodity.xL.Market.Data. 
[MIMICAddIn.txt](https://github.com/cefsharp/CefSharp/files/1919057/MIMICAddIn.txt)
[Morningstar.Commodity.xL.Market.Data.txt](https://github.com/cefsharp/CefSharp/files/1919058/Morningstar.Commodity.xL.Market.Data.txt)

Issue #1574, opened 2016-02-02
WinForms browser has to be removed from Controls before dispose


MihaMarkic:
I've found out that when using ChromiumWebBrowser on a (Win)Form, one should remove it from Controls collection before calling browser.Dispose().
If that's not the case then the MDI container or the form (when not a MDI child) will just close. Looks like WM_CLOSE is passed to the main form in such case for some reason. Not sure if this is documented anywhere.

Repro code:

```
        private ChromiumWebBrowser browser;
        public Form1()
        {
            InitializeComponent();
            browser = new ChromiumWebBrowser("www.google.com")
            {
                Dock = DockStyle.Fill
            };
            Controls.Add(browser);
        }

        private void button1_Click(object sender, EventArgs e)
        {
            //Controls.Remove(browser);
            browser.Dispose();
        }
```

amaitland:
By default `CEF` will attempt to close your form, this is known, not sure if it's written down anywhere. You can implement `ILifespanHandler.DoClose` and ask it politely not to (`return true`).

https://github.com/cefsharp/CefSharp/blob/master/CefSharp/ILifeSpanHandler.cs#L109

MihaMarkic:
Hi Alex,

Do you see any differences in the two approaches? Better/safer to use ILifeSpanHandler?

amaitland:
> Do you see any differences in the two approaches? Better/safer to use ILifeSpanHandler?

Either should be fine. Typically I'd remove the control or it's parent, then `Dispose` the control that was removed. e.g. If the `ChromiumWebBrowser` is on a tab control, remove the tab control, then `Dispose` of the tab control, the `ChromiumWebBrowser` instance will be disposed. In general `WinForms` cleans up after it's self pretty well, it will call `Dispose` on everything eventually. Much nicer than `WPF`!

https://github.com/cefsharp/CefSharp/blob/master/CefSharp.WinForms.Example/BrowserForm.cs#L137

The documentation of `DoClose` is long, if you end up running into problems it's worth a read. https://github.com/cefsharp/CefSharp/blob/master/CefSharp/ILifeSpanHandler.cs#L97

MihaMarkic:
Thanks! I'll just clean up normally for now.

Issue #1683, opened 2016-05-10
SetFocus AccessViolation


amaitland:
```
Unhandled exception at 0x0643E3CB (libcef.dll) in CefSharp.WinForms.Example.exe: 0xC0000005: Access violation reading location 0x00000000.
```

In the `WinForms` control the `OnGotFocus` method calls `SetFocus` which, if the control is immediately `Disposed` causes an access violation. `CEF` performed some refactoring which appears to cause the problem. `CefBrowserHostImpl::SetFocus` calls `OnSetFocus` which immediately posts to the `CEF UI` thread, so it would appear the platform delegate has been released.

https://bitbucket.org/chromiumembedded/cef/src/b660893c6e800374d2b908a0b56ca47441ea2135/libcef/browser/browser_host_impl.cc?at=2526&fileviewer=file-view-default#browser_host_impl.cc-1937

https://bitbucket.org/chromiumembedded/cef/src/64e2fe1d2bd7317bd8305236c9bbe9a1f6313be8/libcef/browser/browser_host_impl.cc?at=2526&fileviewer=file-view-default#browser_host_impl.cc-1604

Change in `CEF` introduced in the `2623` branch (`49`).

Short term it would be ideal to find a way to guard against calling `SetFocus` if the control is `Disposing`.

amaitland:
Short term workaround, remove the control before `Disposing`

amaitland:
http://magpcss.org/ceforum/viewtopic.php?f=6&t=14429

amaitland:
Resolved with https://bitbucket.org/chromiumembedded/cef/commits/e0a22fe9fa257805a0190d9d91d846ea28fb95c5

amaitland:
`master` has been upgraded, this change will be part of `53.0.0`.

PR #1764, opened 2016-08-02
Fix LifeSpanHandler not Called During Dispose

dpbackes:
- Make it so that closing a tab in the WinForms example does not close the
  entire application. We need to make sure that the LifeSpanHandler is
  still around when `managedCefBrowserAdapter` is disposed since it is
  going to call `CloseBrowser`. That way, we can implement LifeSpanHandler
  and reject closing the entire application as described in #1574

### DIFF
--- a/CefSharp.WinForms.Example/Handlers/LifeSpanHandler.cs
+++ b/CefSharp.WinForms.Example/Handlers/LifeSpanHandler.cs
@@ -112,17 +112,19 @@ namespace CefSharp.WinForms.Example.Handlers
 
         void ILifeSpanHandler.OnAfterCreated(IWebBrowser browserControl, IBrowser browser)
         {
-            
+
         }
 
         bool ILifeSpanHandler.DoClose(IWebBrowser browserControl, IBrowser browser)
         {
-            return false;
+            // We do not want to close the entire application when a tab is closed. Thus we return true here to indicate
+            // that the parent window should not be closed when the browser is closed.
+            return true;
         }
 
         public void OnBeforeClose(IWebBrowser browserControl, IBrowser browser)
         {
-            
+
         }
     }
 }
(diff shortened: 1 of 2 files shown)


CefSharp.WinForms/ChromiumWebBrowser.cs
```diff
@@ -339,6 +336,10 @@ protected override void Dispose(bool disposing)
                 TitleChanged = null;
                 IsBrowserInitializedChanged = null;
             }
+
+            // Don't utilize any of the handlers anymore:
+            this.SetHandlersToNull();
```
amaitland: Needs additional comment regarding being called after `managedCefBrowserAdapater` being `Disposed`.

CefSharp.WinForms.Example/Handlers/LifeSpanHandler.cs
```diff
@@ -117,7 +117,7 @@ void ILifeSpanHandler.OnAfterCreated(IWebBrowser browserControl, IBrowser browse
 
         bool ILifeSpanHandler.DoClose(IWebBrowser browserControl, IBrowser browser)
         {
-            return false;
+            return true;
```
amaitland: As `false` is the default, please add additional comment

amaitland:
@dpbackes Please update the title of this `PR` as it doesn't modifying the example isn't the primary purpose of this change.

This `PR` is second in the queue for review, hopefully be reviewed next week (or at least the next couple of weeks).

> Make it so that closing a tab in the WinForms example does not close the entire application.

The `WinForms` example likely crashes when closing a tab as per #1683. I've never seen it close the whole application, how do you reproduce this using the example? (I'm aware of the problem in general).

dpbackes:
Yeah it definitely suffers from #1683. To make this PR I just added 

```
Controls.Remove((Control) Browser);
Browser.Dispose();
```

to the `BrowserTabControl.Dispose()` method while I was developing.

dpbackes:
Updated to address PR comments.

amaitland:
> Yeah it definitely suffers from #1683. To make this PR I just added

@dpbackes In general if you need to make changes to the example to reproduce an issue then those changes need to be included, make sense?

In general, looks ok, will merge when I have a minute.

amaitland:
Follow up commit https://github.com/cefsharp/CefSharp/commit/79f3844cab2bd892276faba2b703457b54564089

Issue #1790, opened 2016-09-01
NullReferenceException from SchemeHandlerFactoryWrapper.Create


justin-caldicott:
- **What version of the product are you using?**

49.0.1 via NuGet.

- **What architecture x86 or x64?**

x86

- **On what operating system?**

Win10

- **Are you using `WinForms`, `WPF` or `OffScreen`?**

WinForms

- **What steps will reproduce the problem?**

Return null from a custom implementation of ISchemeHandlerFactory.Create

https://gist.github.com/justin-caldicott/c20bbd011c250d9f6d226d598cb2b9ff

- **What is the expected output? What do you see instead?**

This previously allowed default CEF handling to occur.  We were using it for requests to third party services such as Mapbox.

It now throws a NullReferenceException.

- **Please provide any additional information below.**

Looks like it was broken in this commit:
https://github.com/cefsharp/CefSharp/commit/12f42331dcd25135ff7ddf041a571b63bd32bde2#diff-e8faa1453883ff4c7ab38a74eac0b01eL40

When an if block was removed.  It was OK in 47.0.2.

Adding back the if block, with just the return null, fixes it for me:
https://github.com/bluebirdtech/CefSharp/commit/7a2d7896035dbaadf4ad26b48df99c3264651b16

- **Does this problem also occur in the `Cef` TestApp from http://opensource.spotify.com/cefbuilds/index.html (http://cefbuilds.com for older version)?**

Not tried it.  I expect this fix to be straight forward to apply, but let me know if you need more detail.

amaitland:
@justin-caldicott By all means submit a `PR` targeted against `master`

> 49.0.1 via NuGet.

For reference this version is no longer supported.

justin-caldicott:
Thanks @amaitland I'll do that.  Needed to patch 49.0.1 as we're not able to jump to .NET 4.5 just yet.

amaitland:
No problem, you'll have to build your own packages.

justin-caldicott:
That's fine I've done that already.  Thanks for doing such a great job BTW building locally was very straight forward!

amaitland:
Resolved with #1792

PR #1792, opened 2016-09-01
Fix NRE when returning null from a scheme handler factory

justin-caldicott:
Regression in 49.0 release.  Prevented use of default CEF handler when
using a custom scheme handler.


amaitland:
Resolves #1790

amaitland:
@justin-caldicott Thanks, greatly appreciated :+1:

PR #1797, opened 2016-09-13
Fix spelling error (#1)

kdriedger-simutech:


amaitland:
@kdriedger-simutech Thanks.

PR #1801, opened 2016-09-19
Fix for browser subprocess crash with null reference exception.

maloshuk:
CefSharp.BrowserSubprocess.exe sometimes crashes.

Logs from system event viewer:

> 4/29/2016 10:31:47 AM
> Faulting application name: CefSharp.BrowserSubprocess.exe, version: 47.0.3.0, time stamp: 0x56ce5bf5
> Application: CefSharp.BrowserSubprocess.exe
> Framework Version: v4.0.30319
> Description: The process was terminated due to an unhandled exception.
> Exception Info: System.NullReferenceException
> Stack:
> at >.CefSharp.CefAppUnmanagedWrapper.OnRenderThreadCreated(CefSharp.CefAppUnmanagedWrapper, CefRefPtr)
> at .CefExecuteProcess(CefMainArgs, CefRefPtr, Void*)
> at CefSharp.CefAppWrapper.Run()
> at CefSharp.BrowserSubprocess.Program.Main(System.String[])

I fixed this by adding checks on NULL for values in CefAppUnmanagedWrapper.cpp file function CefAppUnmanagedWrapper::OnRenderThreadCreated(...). 
Without looking for reasons of NULL values.

### DIFF
--- a/CefSharp.BrowserSubprocess.Core/CefAppUnmanagedWrapper.cpp
+++ b/CefSharp.BrowserSubprocess.Core/CefAppUnmanagedWrapper.cpp
@@ -424,7 +424,17 @@ namespace CefSharp
 
     void CefAppUnmanagedWrapper::OnRenderThreadCreated(CefRefPtr<CefListValue> extraInfo)
     {
+		if (extraInfo == NULL) {
+			LOG(ERROR) << "extraInfo parameter in CefAppUnmanagedWrapper::OnRenderThreadCreated() is NULL";
+			return;
+		}
+
         auto extensionList = extraInfo->GetList(0);
+
+		if (extensionList == NULL) {
+			LOG(ERROR) << "extensionList value in CefAppUnmanagedWrapper::OnRenderThreadCreated() is NULL";
+			return;
+		}
 
         for (size_t i = 0; i < extensionList->GetSize(); i++)
         {
--- a/CefSharp.Wpf/ChromiumWebBrowser.cs
+++ b/CefSharp.Wpf/ChromiumWebBrowser.cs
@@ -470,6 +470,7 @@ namespace CefSharp.Wpf
                     if (tooltipTimer != null)
                     {
                         tooltipTimer.Tick -= OnTooltipTimerTick;
+                        tooltipTimer.Stop();
                     }
 
                     if (CleanupElement != null)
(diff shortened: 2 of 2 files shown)


CefSharp.BrowserSubprocess.Core/CefAppUnmanagedWrapper.cpp
```diff
@@ -424,8 +424,18 @@ namespace CefSharp
 
     void CefAppUnmanagedWrapper::OnRenderThreadCreated(CefRefPtr<CefListValue> extraInfo)
     {
+		if (extraInfo == NULL) {
```
amaitland: Spaces not tabs, braces on newline for consistency.

amaitland:
@maloshuk I've committed https://github.com/cefsharp/CefSharp/commit/bdb37881c550c064c2011c1754060db3549696bf which checks for a valid `list` in `CefAppUnmanagedWrapper::OnRenderThreadCreated`. 

As I've no way to reproduce either of your issues and your not particularly responsive at the moment this is likely the only change that will be included in the upcoming `53.0.0` release. The `DispatcherTimer`, should probably be revisited at some point. I'm surprised that `.Net` doesn't cleanup after it's self (nobody else has reported the problem before you). 

I didn't write the original `DispatcherTimer` related code, looking at it now I'm surprised a new `DispatcherTimer` is required, would make sense to reuse the existing one, so that's the solution I think should be investigated.

jltrem:
@amaitland I've hit the same crash.  Like @maloshuk, my situation requires creating and destroying ChromiumWebBrowser often.  I've not yet had a chance to evaluate the memory issue but will keep an eye on it.

amaitland:
> @amaitland I've hit the same crash.

Are your using `53.0.0`?

jltrem:
> Are your using 53.0.0?
No... 49.0.1.  Working on a project that must stay on .NET 4.0 for now.

amaitland:
That version is no longer supported. If you or your company wish to take ownership of supporting that old branch then your welcome to take ownership and apply any patches you like.

amaitland:
`ToolTipTimer` is now reused instead of a new one created every time `Load` is called. See https://github.com/cefsharp/CefSharp/commit/56f693fa69de2e5047043a59e243f01b1de5ecb1

amaitland:
Anyone interested in sponsoring a `49.0.2` bug fix release see #1910

PR #1847, opened 2016-10-27
Cefsharp 53

markkslee:


TheScripter317:
Executing Javascript with result should be easier.

I used Awesomium 1.7.5.1 before, and it was really easy:
WebControl.ExecuteJavascriptWithResult("document.getElementById('link').href;")

Now it is more complicated, could this be improved for the next version of CefSharp?

PR #1878, opened 2016-12-02
Add IRequestHandler.OnSelectClientCertificate

smaranh:
To show a custom certificate selection dialog by providing a Client Certificate selection hook.

Current Behavior:
(1) it automatically selects the first server certificate instead of allowing the user to cancel and not use any client certificate (e.g. if client cert auth is optional)
(2) it doesn't allow the user to choose a certificate if multiple are present.

Required Behavior:
Provide a onSelectClientCertificate callback that would allow display of certificate list to the user and return the selected certificate.

Combination of these two commits in branch 2883

https://bitbucket.org/chromiumembedded/cef/commits/97fe56d30cb7440813f58ed4bc0ca76e30ec75c4?at=master

https://bitbucket.org/chromiumembedded/cef/commits/97fe56d30cb7440813f58ed4bc0ca76e30ec75c4?at=master


amaitland:
- Add a description and link the relevant commit where this feature is added to `CEF`.
- Add all the relevant interfaces
- Add wrappers for the certificate and callback
- Plumb it all together
- Cleanup formatting (spaces not tabs)

For dealing with the vector, find an example that references `iterator`, probably post data would work.

smaranh:
So this how I am connecting the following,
1. `CefRequestHandler::X509CertificateList` to `[System.Security.Cryptography.X509Certificates.X509Certificate2Collection](https://msdn.microsoft.com/en-us/library/system.security.cryptography.x509certificates.x509certificate2collection(v=vs.110).aspx)`
2. CefX509Certificate to `[System.Security.Cryptography.X509Certificates.X509Certificate](https://msdn.microsoft.com/en-us/library/system.security.cryptography.x509certificates.x509certificate(v=vs.110).aspx)`

This way, I will be able to use `[X509Certificate2UI](https://msdn.microsoft.com/en-us/library/system.security.cryptography.x509certificates.x509certificate2ui(v=vs.110).aspx)` which has a method `SelectFromCollection()` which will allow the display of certificate collection.

Converting from CefType to the required .NET type is where I am stuck.

Am I going in the direction? If not, what would be the right way?

CefSharp/ISelectCertCallback.cs
```diff
@@ -0,0 +1,25 @@
...
+// Use of this source code is governed by a BSD-style license that can be found in the LICENSE file.
+using System;
+using System.Security.Cryptography.X509Certificates;
+
+namespace CefSharp
+{
+    /// <summary>
+    /// Callback interface used for fetching the selected certificate.
+    /// </summary>
+    public interface ISelectCertCallback : IDisposable
```
amaitland: This should be `ISelectClientCertificateCallback`. Keep the existing convention so it's easier to map the `CEF` methods to `CefSharp` ones.

smaranh: will do

amaitland:
> This way, I will be able to use X509Certificate2UI msdn link which has a method SelectFromCollection() which will allow the display of certificate collection.

Nice idea in theory. Can you load a `X509Certificate` from one of the binary formats exposed by the `CefX509Certificate`, work that out then you can determine which direction you'd need to take.

smaranh:
Just to add to that, I think I should follow the similar implementation as done in CEF's [client_handler](https://bitbucket.org/chromiumembedded/cef/src/d7f087e76634af8ceffdef28c794b7310bb507e5/tests/cefclient/browser/client_handler.cc?at=2883&fileviewer=file-view-default#client_handler.cc-712)

For this, I think I have to do the following,

1. Provide Certificates to my `CefCertCallbackWrapper`, this way I can make use of the `iterator` you provided and is also how the selection is being returned in the Cef's client_handler (the link is above)
2. Like the client_handler, I will need to return the `cert_name` which I need to figure out which .NET X509Certificate2 property/method to use to fetch that
3. If I get that right, then I can [iterate through the certificates](https://bitbucket.org/chromiumembedded/cef/src/d7f087e76634af8ceffdef28c794b7310bb507e5/tests/cefclient/browser/client_handler.cc?at=2883&fileviewer=file-view-default#client_handler.cc-726) and return the selected certificate in `SELECT` method

Now the only thing that is stopping me, is how to wrap the CEF certificateList to X509Certificate2Collection, so that there is no typecast error during compiling this.

amaitland:
> Provide Certificates to my CefCertCallbackWrapper, this way I can make use of the iterator you provided and is also how the selection is being returned in the Cef's client_handler (the link is above)

Work out if you can load an `X509Certificate2` from the `CefX509Certificate`, You'll waste a lot of time if you find out that it's not possible.

amaitland:
> Like the client_handler, I will need to return the cert_name which I need to figure out which .NET X509Certificate2 property/method to use to fetch that
> If I get that right, then I can iterate through the certificates and return the selected certificate in SELECT method

This doesn't directly map the `CEF API`, which in general I'm against. So you'd really need a good reason to do so. If you can reuse `X509Certificate2` then that maybe an acceptable reason, otherwise it will need to be a direct `1:1` mapping.

smaranh:
> Work out if you can load an X509Certificate2 from the CefX509Certificate, You'll waste a lot of time if you find out that it's not possible.

Is there a sample, as to how I can do this? Yes, if this is doable, things will be lot easier.

amaitland:
> Is there a sample, as to how I can do this?

Unfortunately no, the `CefX509Certificate` class was only added fairly recently and is not exposed.

https://bitbucket.org/chromiumembedded/cef/commits/c83b6a1

smaranh:
No, what I meant was a sample for `1:1` mapping following the convention in CefSharp. Which I can use to replicate `CefX509Certificate`.

amaitland:
> No, what I meant was a sample for 1:1 mapping following the convention in CefSharp. Which I can use to replicate CefX509Certificate.

Large parts of the `API` are `1:1` mapping, like `IRequest/IFrame/IBrowser/IPostData`, etc

Thinking about it more, I think you'd be better off starting small, expose the `CefX509Certificate` method that was added to ` CefSSLInfo` in https://bitbucket.org/chromiumembedded/cef/commits/c83b6a1

You'd only need to add a very simple wrapper for `CefSSLStatus` and then it would be much easier to test if you can map the binary data and create a new instance of `X509Certificate2`

> Like the client_handler, I will need to return the cert_name which I need to figure out which .NET X509Certificate2 property/method to use to fetch that

Thinking about this more, and the question I have is are certificate names unique? Might need to match a combination of things, expiration and name possibly.

smaranh:
> Thinking about this more, and the question I have is are certificate names unique? Might need to match a combination of things, expiration and name possibly.

Looks like the cert_name is X509Certificate->subject name, which seems to be a [unique identifier.](https://blogs.technet.microsoft.com/option_explicit/2012/04/09/pki-certificates-and-the-x-509-standard/)

smaranh:
> You'd only need to add a very simple wrapper for CefSSLStatus and then it would be much easier to test if you can map the binary data and create a new instance of X509Certificate2

So basically, after I get `CefX509Certificate`, I should extract the `DER Encoded binary` from it and create a `new X509Certificate`? I will basically have to do this for the entire `X509CertificateList`, right?

amaitland:
> So basically, after I get CefX509Certificate, I should extract the DER Encoded binary from it and create a new X509Certificate?

That or `PEM` encoded, which ever works. 

> I will basically have to do this for the entire X509CertificateList, right?

You'll have to iterate through the vector yes

smaranh:
> That or PEM encoded, which ever works.

The X509Certificate2 [doc](https://msdn.microsoft.com/en-us/library/ms148413(v=vs.110).aspx#Anchor_2) states,

> This constructor creates a new X509Certificate2 object using certificate information from a byte array. The byte array can be binary (DER) encoded or Base64-encoded X.509 data.

and when I have the selected certificate from the X509Certificate2UI, I will also need to recreate CefX509Certificate. What should I use in that case?

amaitland:
> This constructor creates a new X509Certificate2 object using certificate information from a byte array. The byte array can be binary (DER) encoded or Base64-encoded X.509 data.

`Base64 Encoded DER` = `PEM` from memory.

> and when I have the selected certificate from the X509Certificate2UI, I will also need to recreate CefX509Certificate. What should I use in that case?

Is there a `CefX509Certificate::Create` method? If not then you won't be able to create one yourself. You'll have to keep a reference to the vector within the callback.

smaranh:
> Thinking about it more, I think you'd be better off starting small, expose the CefX509Certificate method that was added to `CefSSLInfo`

I tried creating the Interface and wrapper for the `CefSSLInfo` but when it comes to exposing the `GetX509Certificate()`, how can I return the certificate, since I don't have a reference to it. I cannot pass in the certificate when I create `CefSSLInfo` as that is not how the constructor is designed. I looked at [ssl_status_impl.cc](https://bitbucket.org/chromiumembedded/cef/commits/c83b6a1#Llibcef/browser/ssl_status_impl.ccT37) where a certificate is being returned using this function but `ssl_status_impl` already has a reference to the certificate.

amaitland:
> I tried creating the Interface and wrapper for the CefSSLInfo but when it comes to exposing the GetX509Certificate(), how can I return the certificate, since I don't have a reference to it. I cannot pass in the certificate when I create CefSSLInfo as that is not how the constructor is designed. I looked at ssl_status_impl.cc where a certificate is being returned using this function but ssl_status_impl already has a reference to the certificate.

I'm not sure what you mean. You call `CefSSLInfo::GetX509Certificate()`, do an `if(cert.get())` to check you have a valid reference, get the `PEM or DER` encoded binary data, search for (`->GetData` as there are examples), return the `gcnew X509Certificate2(bytes)`  where bytes is the array you've read from the `CefBinaryData`. It sounds complicated, it's not, there are examples in the not currently used `CefSslInfoWrapper`

smaranh:
> You call `CefSSLInfo::GetX509Certificate()`

Let me explain how I am looking at the code flow. In the `OnSelectClientCertificate` method in ClientAdapter will receive the `X509CertificateList`, which is a `std::vector<CefRefPtr<CefX509Certificate>>`. In this method, I do not have any reference to `CefSSLInfo`. What I was trying to do, is fetch the certificateList in a CertificateListWrapper, iterate over the list and fetch the certificate and get the `GetDEREncoded()` which is in type `CefRefPtr<CefBinaryValue>`. 

Now I need help in converting the type from `CefRefPtr<CefBinaryValue>` to `System::Byte[]`, which is the input that `gcnew X509Certificate2(bytes)` will need.

Inside ClientAdapter.cpp -> onSelectClientCertificate
`auto certificateListWrapper = gcnew CefCertificateListWrapper(certificates);`

Inside CefCertificateListWrapper.h (wrapper for certificateList)
`CefCertificateListWrapper(CefRequestHandler::X509CertificateList& certificates)
				: _certificateList(certificates)
			{
				std::vector<CefRefPtr<CefX509Certificate> >::const_iterator it =
					_certificateList.begin();
				for (; it != _certificateList.end(); ++it) {
					System::Byte[] bytes((*it)->GetDEREncoded());
                                        //System::Security::Cryptography::X509Certificates::X509Certificate2 cert = gcnew System::Security::Cryptography::X509Certificates::X509Certificate2(bytes);
				}
			}`

smaranh:
@amaitland Ok so after some head scratching, I did the following,

1. Get certificates from CefCertificateList
2. Create X509Certificate2 from DER encoded data
3. Add newly created certificate to X509CertificateCollection
4. Send this to IRequestHandler

The build was successful but now when I open the browser and try to reach the PKI portal, it returns an error: `Load Error:ConnectionTimedOut;`

amaitland:
> Ok so after some head scratching, I did the following,

Lots of formatting to fix up and some code improvements, in general it looks ok 👍 

> The build was successful but now when I open the browser and try to reach the PKI portal, it returns an error: Load Error:ConnectionTimedOut;

Sorry, no idea. Play around with `cefclient`, ask for help on `ceforum` if you can reproduce the problem.

CefSharp.Core/CefSharp.Core.vcxproj.filters
```diff
@@ -259,6 +259,12 @@
     <ClInclude Include="CookieManager.h">
       <Filter>Header Files</Filter>
     </ClInclude>
+    <ClInclude Include="Internals\CefCertificateCallbackWrapper.h">
+      <Filter>Header Files</Filter>
+    </ClInclude>
+    <ClInclude Include="Internals\CefCertificateListWrapper.h">
```
amaitland: And this?

smaranh:
> Sorry, no idea. Play around with cefclient, ask for help on ceforum if you can reproduce the problem.

Does it matter if I build CefSharp with all these changes in VS2013 or VS2015?

smaranh:
Never mind. It works great. The PKI portal I was hitting was down :dancer:

smaranh:
Looks like to make use of `X509Certificate2UI`, I will need to add [System.Security.dll](http://stackoverflow.com/questions/15617409/name-x509certificate2ui-does-not-exist-in-the-current-context) to the project. 
Should I add it to CefSharp or CefSharp.WinForms?

amaitland:
> Looks like to make use of X509Certificate2UI, I will need to add System.Security.dll to the project.
> Should I add it to CefSharp or CefSharp.WinForms?

Neither, just add it to the example project.

CefSharp.Core/Internals/CefCertificateCallbackWrapper.h
```diff
@@ -0,0 +1,65 @@
...
+		public:
+			CefCertificateCallbackWrapper(CefRefPtr<CefSelectClientCertificateCallback>& callback, const CefRequestHandler::X509CertificateList& certificates)
+				: _callback(callback), _certificateList(certificates)
+			{
+
+			}
+
+			!CefCertificateCallbackWrapper()
+			{
+				_callback = NULL;
```
amaitland: For consistency release `_certificateList` here.

CefSharp.Core/Internals/CefCertificateCallbackWrapper.h
```diff
@@ -0,0 +1,65 @@
...
+#include "Stdafx.h"
+
+#include "include\cef_request_handler.h"
+#include "CefWrapper.h"
+
+namespace CefSharp
+{
+	namespace Internals
+	{
+		private ref class CefCertificateCallbackWrapper : public ISelectClientCertificateCallback, public CefWrapper
```
amaitland: File likely needs to be reformatted for spaces not tabs.

CefSharp.Core/Internals/CefCertificateCallbackWrapper.h
```diff
@@ -0,0 +1,65 @@
...
+			{
+				this->!CefCertificateCallbackWrapper();
+
+				_disposed = true;
+			}
+
+			virtual void Select(System::Security::Cryptography::X509Certificates::X509Certificate2^ cert)
+			{
+				ThrowIfDisposed();
+				if (!cert){
```
amaitland: change this to `if(cert == nullptr)` for consistency

CefSharp.Core/Internals/CefCertificateCallbackWrapper.h
```diff
@@ -0,0 +1,65 @@
...
+					_callback->Select(NULL);
+				}
```
amaitland: Add an `else` block here.

CefSharp.Core/Internals/CefCertificateCallbackWrapper.h
```diff
@@ -0,0 +1,65 @@
...
+
+				//Need cert_name from cert
+				auto cert_name = cert->SubjectName->Name;
```
amaitland: `cert_name` should be `certName` (maybe it even should be `subjectName`?)

CefSharp.Core/Internals/CefCertificateCallbackWrapper.h
```diff
@@ -0,0 +1,65 @@
...
+
+				std::vector<CefRefPtr<CefX509Certificate> >::const_iterator it =
+					_certificateList.begin();
+				for (; it != _certificateList.end(); ++it) {
```
amaitland: Braces on new line.

CefSharp.Core/Internals/CefCertificateCallbackWrapper.h
```diff
@@ -0,0 +1,65 @@
...
+					CefString subject((*it)->GetSubject()->GetDisplayName());
+					if (subject == StringUtils::ToNative(cert_name)) {
```
amaitland: Remove the `ToNative` call outside of the loop, no need to convert it every iteration.

CefSharp.Core/Internals/ClientAdapter.cpp
```diff
@@ -737,6 +738,40 @@ namespace CefSharp
                 StringUtils::ToClr(scheme), callbackWrapper);
         }
 
+		bool ClientAdapter::OnSelectClientCertificate(CefRefPtr<CefBrowser> browser, bool isProxy, const CefString& host,
```
amaitland: Formatting, `spaces` not `tabs`.

CefSharp.Core/Internals/ClientAdapter.cpp
```diff
@@ -737,6 +738,40 @@ namespace CefSharp
...
+
+			if (handler == nullptr)
+			{
+				return false;
+			}
+
+			auto browserWrapper = GetBrowserWrapper(browser->GetIdentifier(), browser->IsPopup());
+			auto callbackWrapper = gcnew CefCertificateCallbackWrapper(callback, certificates);
+
+			auto list = gcnew System::Security::Cryptography::X509Certificates::X509Certificate2Collection();
```
amaitland: Add a `using statement` for `System::Security::Cryptography::X509Certificates`, just to tidy these up a bit.

CefSharp.Core/Internals/ClientAdapter.cpp
```diff
@@ -737,6 +738,40 @@ namespace CefSharp
...
+
+			std::vector<CefRefPtr<CefX509Certificate> >::const_iterator it =
+				certificates.begin();
+			for (; it != certificates.end(); ++it) {
```
amaitland: Braces on new lines.

CefSharp.Core/Internals/ClientAdapter.cpp
```diff
@@ -737,6 +738,40 @@ namespace CefSharp
...
+				CefRefPtr<CefBinaryValue> bytes((*it)->GetDEREncoded());
```
amaitland: You could have just used `auto` here for simplicity.

CefSharp.Core/Internals/ClientAdapter.h
```diff
@@ -135,6 +135,8 @@ namespace CefSharp
             virtual DECL void OnPluginCrashed(CefRefPtr<CefBrowser> browser, const CefString& plugin_path) OVERRIDE;
             virtual DECL void OnRenderViewReady(CefRefPtr<CefBrowser> browser) OVERRIDE;
             virtual DECL void OnRenderProcessTerminated(CefRefPtr<CefBrowser> browser, TerminationStatus status) OVERRIDE;
+			virtual DECL bool OnSelectClientCertificate(CefRefPtr<CefBrowser> browser, bool isProxy, const CefString& host, int port,
```
amaitland: Formatting.

CefSharp/IRequestHandler.cs
```diff
@@ -104,6 +106,8 @@ public interface IRequestHandler
         /// <returns>Return true to continue the request and call CefAuthCallback::Continue() when the authentication information is available. Return false to cancel the request. </returns>
         bool GetAuthCredentials(IWebBrowser browserControl, IBrowser browser, IFrame frame, bool isProxy, string host, int port, string realm, string scheme, IAuthCallback callback);
 
+        bool OnSelectClientCertificate(IWebBrowser browserControl, IBrowser browser, bool isProxy, string host, int port, X509Certificate2Collection certificates, ISelectClientCertificateCallback callback);
```
amaitland: Needs `xml` comment.

CefSharp/ISelectClientCertificateCallback.cs
```diff
@@ -0,0 +1,25 @@
...
+
+namespace CefSharp
+{
+    /// <summary>
+    /// Callback interface used for fetching the selected certificate.
+    /// </summary>
+    public interface ISelectClientCertificateCallback : IDisposable
+    {
+        /// <summary>
+        /// Return the selected certificate
```
amaitland: `Return` seems out of place here (if it's copied from the `CEF` doco, then it should be improved).

CefSharp.Core/Internals/CefCertificateCallbackWrapper.h
smaranh: will do

CefSharp.Core/Internals/CefCertificateCallbackWrapper.h
smaranh: will do

CefSharp.Core/Internals/CefCertificateCallbackWrapper.h
smaranh: will do

CefSharp.Core/Internals/CefCertificateCallbackWrapper.h
smaranh: I think I need to look at this logic again. Seems u were right about this not being a unique Identifier. I am trying to make use of `SerialNumber`. I have that as a string with `cert->GetSerialNumberString` but from CEF, I have the serial number as a `CefRefPtr<CefBinaryValue>`.

How do I convert `CefRefPtr<CefBinaryValue>` into a string?

CefSharp/IRequestHandler.cs
smaranh: will do

CefSharp.Core/Internals/ClientAdapter.cpp
smaranh: will do

CefSharp.Core/Internals/ClientAdapter.cpp
smaranh: will do

CefSharp.Core/Internals/ClientAdapter.cpp
smaranh: will do

CefSharp.Core/Internals/CefCertificateCallbackWrapper.h
smaranh: will do

CefSharp.Core/Internals/CefCertificateCallbackWrapper.h
smaranh: how do I release `_certificateList`? It shows error if I `delete` it or if I set it to `nullptr`

amaitland: Try removing the `const` keyword from it's deceleration then set it to `NULL`.

CefSharp.Core/Internals/CefCertificateCallbackWrapper.h
amaitland: Read the binary data and use one of the `Encoding` classes to convert to a `string`. If you have to compare the serial, then you'd probably be better to convert the string you have and compare the two arrays.

CefSharp/ISelectClientCertificateCallback.cs
amaitland: Start with the comment from the `CEF` source and just change the `null` to lowercase

`Chooses the specified certificate for client certificate authentication. null value means that no client certificate should be used.`

CefSharp.Core/Internals/CefCertificateCallbackWrapper.h
smaranh: Ok, so I tried to read the binary data and use `Encodeing` class like this,
```
auto serialNum((*it)->GetSerialNumber());

auto byteSize = serialNum->GetSize();

auto bufferByte = gcnew cli::array<Byte>(byteSize);
pin_ptr<Byte> src = &bufferByte[0]; // pin pointer to first element in arr

serialNum->GetData(static_cast<void*>(src), byteSize, 0);

UTF8Encoding^ encoding = gcnew UTF8Encoding;
auto serialStr(encoding->GetString(bufferByte));
```
but `serialStr` comes out as a messed up string (more like symbols that strings). I tried it with all the `Encoding` types [msdn link](https://msdn.microsoft.com/en-us/library/system.text.encoding(v=vs.110).aspx?cs-save-lang=1&cs-lang=cpp#code-snippet-1).

I ultimately, recreated the certificate from the certificateList in the iteration for each CefCertificate and then compared their serial number and that worked. I guess this is not a workaround.

CefSharp.Example/RequestHandler.cs
```diff
@@ -137,8 +137,15 @@ bool IRequestHandler.OnSelectClientCertificate(IWebBrowser browserControl, IBrow
             // We also suggest you explicitly Dispose of the callback as it wraps an unmanaged resource.
 
             callback.Dispose();
-            //X509Certificate2Collection selectedCertificateCollection = System.Security.Cryptography.X509Certificates.X509Certificate2UI
             return false;
+
+            // This is an alternate for displaying a Client Certificate Dialog for user to select a certificate
+            //X509Certificate2Collection selectedCertificateCollection = X509Certificate2UI.SelectFromCollection(certificates, "Certificates Dialog", "Select Certificate for authentication", X509SelectionFlag.SingleSelection);
```
amaitland: Does this work? May as well uncomment it if the code works.

CefSharp.Core/Internals/CefCertificateCallbackWrapper.h
amaitland: Doesn't sound like a great solution. I guess the collection size will be fairly small, so the performance cost won't be too high. Might be workable for now, make sure to add a `TODO` with relevant comments.

amaitland:
Formatting in general still needs some work, I use `Productivity Power Tools from Microsoft` to `untabify` the code. https://marketplace.visualstudio.com/items?itemName=VisualStudioProductTeam.ProductivityPowerTools

CefSharp.Example/RequestHandler.cs
smaranh: Yes, this works. This should provide a default dialog but not sure if that will work for WPF?

smaranh: Never mind I will move it to `WinFormsRequestHandler.cs`

smaranh:
Remove all the tabs with VS untabify. If you see a preexisting code as changed, that is because they have been untabified.
Also, added the Certificate Dialog to WinForms example.

CefSharp/IRequestHandler.cs
```diff
@@ -3,6 +3,8 @@
 // Use of this source code is governed by a BSD-style license that can be found in the LICENSE file.
 
 using System;
+using System.Collections.Generic;
```
amaitland: Is this using required still?

CefSharp/IRequestHandler.cs
```diff
@@ -105,6 +107,20 @@ public interface IRequestHandler
...
+        /// Called when the browser needs user to select Client Certificate for authentication requests (eg. PKI authentication).
+        /// </summary>
+        /// <param name="browserControl">The ChromiumWebBrowser control</param>
+        /// <param name="browser">the browser object</param>
+        /// <param name="isProxy">indicates whether the host is a proxy server</param>
+        /// <param name="host">hostname</param>
+        /// <param name="port">port number</param>
+        /// <param name="certificates">List of Client certificates for selection</param>
+        /// <param name="callback">Callback interface used for asynchronous continuation of client certificate selection for authentication requests.</param>
+        /// <returns>Return true to continue the request and call CefSelectClientCertificateCallback::Select() with the selected certificate for authentication.
```
amaitland: Need to replace `CefSelectClientCertificateCallback::Select()` with `ISelectClientCertificateCallback.Select()`

CefSharp.Example/RequestHandler.cs
```diff
@@ -4,6 +4,8 @@
 
 using System;
 using CefSharp.Example.Filters;
+using System.Collections.Generic;
```
amaitland: Is this required?

CefSharp.Core/Internals/CefCertificateCallbackWrapper.h
```diff
@@ -0,0 +1,79 @@
...
+                    for (; it != _certificateList.end(); ++it) 
+                    {
+
+                        auto bytes((*it)->GetDEREncoded());
+                        auto byteSize = bytes->GetSize();
+
+                        auto bufferByte = gcnew cli::array<Byte>(byteSize);
+                        pin_ptr<Byte> src = &bufferByte[0]; // pin pointer to first element in arr
+
+                        bytes->GetData(static_cast<void*>(src), byteSize, 0);
```
amaitland: Need a `TODO` here as this should be looked at again at some point, include what you've tried already and the reason this was selected as an implementation.

CefSharp.Core/Internals/CefCertificateCallbackWrapper.h
```diff
@@ -0,0 +1,79 @@
...
+            {
+                this->!CefCertificateCallbackWrapper();
+
+                _disposed = true;
+            }
+
+            virtual void Select(System::Security::Cryptography::X509Certificates::X509Certificate2^ cert)
+            {
+                ThrowIfDisposed();
+                if (cert == nullptr){
```
amaitland: Brace on new line.

CefSharp/IRequestHandler.cs
smaranh: removed

CefSharp/IRequestHandler.cs
smaranh: done

CefSharp.Example/RequestHandler.cs
smaranh: removed

CefSharp.Core/Internals/CefCertificateCallbackWrapper.h
smaranh: done

amaitland:
Merged 👍 Follow up changes are https://github.com/cefsharp/CefSharp/commit/b7f7cc4dc6c5da4facc11c7eb972723df65284f0

The certificate selection should be made on the `UI` thread so as not to block the browser, example was updated.

smaranh:
Awesome, my first open source merge :dancer:

Issue #1904, opened 2016-12-30
Fix: Network Graph - "Too many forks to display"


jdc20181:
Howdy!
I was looking through the Readme, seen where you link to your network graph here is a quote for reference:
"This is the official CefSharp fork, as maintained by the CefSharp community. You can also view the **entire network of public forks/branches**." - Readme.md

The bolded section is where the link is. 

However, I get the following error:

"too many forks to load" 

So.... 

How can you view them when github doesn't allow it?

amaitland:
You would have to raise the issue with Github, outside of our control. There is no way to view the network.

cztomczak:
You can find all forks using GitHub search: "[cefsharp fork:true](https://github.com/search?utf8=%E2%9C%93&q=cefsharp+fork%3Atrue&type=Repositories&ref=searchresults)"

jdc20181:
The readme should be updated then...

amaitland:
@cztomczak Thanks for the handy tip 👍 

@jdc20181 Pull requests welcome.

PR #1905, opened 2017-01-01
See Full Desc. Fixes Forks Display.

jdc20181:
Updated: Line 39 fixed the link to go to the alternative method. 
Additonally, I added around line 40, a explaination on how to view all forks of Cefsharp. This will solve the Issue  #1904


amaitland:
Thx 👍

PR #1918, opened 2017-01-18
Request event handler

FlorianObermayer:
I implemented an example for an  `IRequestHandler` implementation capable of using events. It's a first draft, maybe @amaitland wants to include it in the examples.


amaitland:
@FlorianObermayer Let me know when it's ready for review and I'll have a look.

Please match the existing coding style

- No regions
- braces on new lines
- One class per file
- Spaces instead of tabs

See https://github.com/cefsharp/CefSharp/blob/master/CONTRIBUTING.md#coding-style for more details

FlorianObermayer:
@amaitland I just applied your preferred coding style and proofread the whole thing.

Note that there are 2 `TODOs` in the code

CefSharp.Example/RequestEventHandler/RequestEventHandler.cs
```diff
@@ -0,0 +1,157 @@
...
+                args.Callback.Dispose();
+            }
+            return args.ContinueAsync;
+        }
+
+        public void OnResourceRedirect(IWebBrowser browserControl, IBrowser browser, IFrame frame, IRequest request, ref string newUrl)
+        {
+            var args = new OnResourceRedirectEventArgs(browserControl, browser, frame, request, new StringBuilder(newUrl));
+            OnResourceRedirectEvent?.Invoke(this, args);
+            //TODO: test if the string is really changed by using a StringBuilder (should work afaik ;) )
```
FlorianObermayer: `//TODO: test if the string is really changed by using a StringBuilder (should work afaik ;) )`

CefSharp.Example/RequestEventHandler/RequestEventHandler.cs
```diff
@@ -0,0 +1,155 @@
...
+            OnOpenUrlFromTabEvent?.Invoke(this, args);
+            return args.CancelNavigation;
+        }
+
+        public bool OnCertificateError(IWebBrowser browserControl, IBrowser browser, CefErrorCode errorCode, string requestUrl, ISslInfo sslInfo, IRequestCallback callback)
+        {
+            var args = new OnCertificateErrorEventArgs(browserControl, browser, errorCode, requestUrl, sslInfo, callback);
+            OnCertificateErrorEvent?.Invoke(this, args);
+
+            if (!args.Callback.IsDisposed)
```
FlorianObermayer: TODO: check if disposing the callback is intended like that

CefSharp.Example/RequestEventHandler/RequestEventHandler.cs
```diff
@@ -0,0 +1,155 @@
...
+                args.Callback.Dispose();
+            }
+            return args.ContinueAsync;
+        }
+
+        public void OnResourceRedirect(IWebBrowser browserControl, IBrowser browser, IFrame frame, IRequest request, ref string newUrl)
+        {
+            var args = new OnResourceRedirectEventArgs(browserControl, browser, frame, request, new StringBuilder(newUrl));
+            OnResourceRedirectEvent?.Invoke(this, args);
+            if (!Equals(newUrl, args.NewUrl.ToString()))
```
FlorianObermayer: TODO: test if the string is really changed by using a StringBuilder

CefSharp.Example/RequestEventHandler/EventArgs/OnProtocolExecutionEventArgs.cs
```diff
@@ -0,0 +1,17 @@
...
+        {
+            Url = url;
+        }
+
+        public string Url { get; }
+
+        /// <summary>
+        ///     Set to true to attempt execution via the registered OS protocol handler, if any. Otherwise set to false.
+        /// </summary>
+        public bool AttemptExecution { get; set; } = false;
```
FlorianObermayer: TODO: discuss if the default should be true

amaitland: The `CEF` default is `false`

CefSharp.Example/RequestEventHandler/RequestEventHandler.cs
amaitland: Seems a little odd your accessing the callback through the args, when it's a param of the method. Apart from that the concept is fine. You could have a static helper function, save some minor code repetition.

CefSharp.Example/RequestEventHandler/RequestEventHandler.cs
amaitland: And the reason your using a `StringBuilder` is? They're expensive, specially for short strings. Personally I'd find it very strange to be accessing the `newUrl` via one. Just use a `string` or maybe a `Uri`.

amaitland:
Looking greatly improved 👍 

> When adding new files, please prepend the standard license disclaimer (just copy and paste from another source file)

Please add the license disclaimer to each file.

CefSharp.Example/RequestEventHandler/EventArgs/GetResourceResponseFilterEventArgs.cs
```diff
@@ -0,0 +1,21 @@
...
+            Request = request;
+            Response = response;
+        }
+
+        public IFrame Frame { get; }
+        public IRequest Request { get; }
+        public IResponse Response { get; }
+
+        /// <summary>
+        ///     Set IResponseFilter to intercept this response, otherwise return null
```
amaitland: `return null` doesn't really seem appropriate as a comment here.

CefSharp.Example/RequestEventHandler/EventArgs/OnBeforeResourceLoadEventArgs.cs
```diff
@@ -0,0 +1,28 @@
...
+        ///     Callback interface used for asynchronous continuation of url requests.
+        /// </summary>
+        public IRequestCallback Callback { get; }
+
+        /// <summary>
+        ///     To cancel loading of the resource return <see cref="F:CefSharp.CefReturnValue.Cancel" />
+        ///     or <see cref="F:CefSharp.CefReturnValue.Continue" /> to allow the resource to load normally. For async
+        ///     return <see cref="F:CefSharp.CefReturnValue.ContinueAsync" />
+        /// </summary>
+        public CefReturnValue ResourceLoadHandling { get; set; } = CefReturnValue.Continue;
```
amaitland: Better name for this property perhaps?

CefSharp.Example/RequestEventHandler/EventArgs/OnQuotaRequestEventArgs.cs
```diff
@@ -0,0 +1,29 @@
...
+
+        /// <summary>
+        ///     Callback interface used for asynchronous continuation of url requests.
+        /// </summary>
+        public IRequestCallback Callback { get; }
+
+        /// <summary>
+        ///     Set to false to cancel the request immediately. Set to true to continue the request
+        ///     and call <see cref="M:CefSharp.IRequestCallback.Continue(System.Boolean)" /> either in this method or at a later
+        ///     time to
```
amaitland: Fix up this comment.

amaitland:
Just as a general note, up until now I've avoided using the newer compiler syntactic sugar like `.?` as the one time I accepted a `PR` that used it, people started complaining as they didn't have the relevant components installed. If people have trouble compiling we'll need to revert to a plainer more verbose syntax.

CefSharp.Example/RequestEventHandler/RequestEventHandler.cs
FlorianObermayer: I actually don't need that here at all.. Simply sticking to `string` now and check if that changed within event handler

CefSharp.Example/RequestEventHandler/RequestEventHandler.cs
FlorianObermayer: I just added 2 private methods `EnsureCallbackDisposal(...)` . I'd have preferred to use a kind of `IDisposalAware` interface as parameter instead of overloading it, but that requires core changes which is just too much here.

FlorianObermayer:
Regarding syntactical sugar, as this is intended to be part of the Example project, (which is not shipped via h NuGet - is it?), you can make an exception here. The crux with the old syntax is that we need an event invocator for every event so that would bloat the whole thing for no real benefit.

amaitland:
The problem arises when people attempt to build the project from source in an atempt to execute the example, so the problem shouldn't be so easily dismissed. We'll see how it goes for now.

amaitland:
Will look at merging this when `AppVeyor` builds are working again.

amaitland:
Attempting to build locally on a machine with only `VS2013` and it's failing with the `C# 6.0` features.  

https://social.msdn.microsoft.com/Forums/en-US/49ba9a67-d26a-4b21-80ef-caeb081b878e/will-c-60-ever-be-supported-by-vs-2013?forum=roslyn

I guess there are a few options

- Rewrite the code
- Exclude the files from the project

Until `VS2015` is the default we'll need to stay away from `C# 6.0` features.

amaitland:
`AppVeyor` builds are working again, as expected this fails. 

Let me know how you wish to proceed.

FlorianObermayer:
@amaitland I just made the code compatible to C# < 6. Feel free to merge it :)
Note that the target merge branch is `cefsharp:cefsharp/53` and not the latest master, not sure how to proceed here.

amaitland:
> @amaitland I just made the code compatible to C# < 6.

Great 👍 

> Note that the target merge branch is cefsharp:cefsharp/53 and not the latest master, not sure how to proceed here.

I will take care of it, in future if you can target against `master` that would be appreciated.

amaitland:
I've merged into `master` and made a few changes to remove some duplication and explicitly implement the interface

https://github.com/cefsharp/CefSharp/commit/64b63b3dce4226cc7824b9aa0319c158590dd384

The `IRequestHandler.OnSelectClientCertificate` method has been added, not implemented.

Issue #1928, opened 2017-02-08
V55: VisitUrlCookiesAsync NullReferenceException


bobahml:
- **After update to V55 (from Nuget). I see an System.NullReferenceException in my application log**
    - CefSharp.Core/CookieManager.cpp#L107
CefSharp.CookieManager.VisitUrlCookiesAsync(String url, Boolean includeHttpOnly) in c:\projects\cefsharp\cefsharp.core\cookiemanager.cpp:line 107

- **What version of the product are you using?**
    - Nuget v 55.0.0
- **What architecture x86 or x64?**
    - x64
- **On what operating system?**
    - windows 8.1 Pro 
- **Are you using `WinForms`, `WPF` or `OffScreen`?**
    - OffScreen
- **Gist for CefSharp.MinimalExample**
    - https://gist.github.com/bobahml/8f963f67e23938d3592da300e7536b31
- **What is the expected output? What do you see instead?**
    - It is expected the correct list of cookies
- **stack trace.**
    - CefSharp.CookieManager.VisitUrlCookiesAsync(String url, Boolean includeHttpOnly) in c:\projects\cefsharp\cefsharp.core\cookiemanager.cpp:line 107

amaitland:
You can easily debug the source now we have gitlink see #1680 

Please submit a PR when you have a proposed solution

amaitland:
Which version does your code work against?

amaitland:
Resolved in `master` with https://github.com/cefsharp/CefSharp/commit/e6bcbe411194466d4c902c32865cb1ed85797fc6

If unable to access the cookies `null` will be returned instead of a `Task`.

PR #1930, opened 2017-02-08
fix: VisitUrlCookiesAsync NullReferenceException #1928

bobahml:
The error contained in TaskCookieVisitor class. 
Method Dispose set taskCompletionSource=null, and then property Task called by an external code.

### DIFF
--- a/CefSharp.OffScreen.Example/Program.cs
+++ b/CefSharp.OffScreen.Example/Program.cs
@@ -82,6 +82,12 @@ namespace CefSharp.OffScreen.Example
 
                 var onUi = Cef.CurrentlyOnThread(CefThreadIds.TID_UI);
 
+                var completionHandler = new TaskCompletionHandler();
+                var cookieManager = browser.RequestContext.GetDefaultCookieManager(completionHandler);
+                await completionHandler.Task;
+                //Get cookies by Url
+                var cookes = await cookieManager.VisitUrlCookiesAsync(TestUrl, false);
+
                 // For Google.com pre-pupulate the search text box
                 await browser.EvaluateScriptAsync("document.getElementById('lst-ib').value = 'CefSharp Was Here!'");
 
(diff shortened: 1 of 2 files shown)


amaitland:
Great you've attempted a fix 👍 Please update the description with details of what your proposed changes achieve. 

Have you attempted to implement it in your code and verified it works? Obviously you can implement your own handler that has your proposed changes

I'm concerned you'll end up with a Task that's never executed.

amaitland:
> I'm concerned you'll end up with a Task that's never executed.

Ignore that, it should be ok. 

Review your changes above, formatting is out. Spaces not tabs thank you!

amaitland:
Thinking about it a little more, I think we should `return null` and update the documentation.

> Returns false if cookies cannot be accessed. 

`VisitUrlCookies` returning `false` means something different to there not being any cookies.

https://github.com/cefsharp/CefSharp/blob/cefsharp/55/CefSharp.Core/CookieManager.cpp#L104

bobahml:
I think it would be correct to throw an exception immediately if the method is called from no suitable thread.

amaitland:
> I think it would be correct to throw an exception immediately if the method is called from no suitable thread.

I'm fine with that, it's not clear if they explicitly need to be called on the `CEF IO` thread or not.

See http://magpcss.org/ceforum/apidocs3/projects/(default)/CefCookieManager.html for the `CEF API`.

amaitland:
> Returns false if cookies cannot be accessed. 

It's not clear the changes your proposing adequately express that cookies cannot be accessed.

> The methods of this class may be called on any thread unless otherwise indicated.

http://magpcss.org/ceforum/apidocs3/projects/(default)/CefCookieManager.html
Reading the `CEF API` again, I'm not sure it's threading related.

amaitland:
Closing this as  #1928 should now be resolved. Thanks for making an attempt, future contributions welcome 👍

Issue #1723, opened 2016-06-27
WPF crashing with open combobox


goeranbarz:
- **What version of the product are you using?**
  49.0.1 from NuGet
- **What architecture x86 or x64?**
   x64
- **On what operating system?**
  Windows 8.1
- **Are you using `WinForms`, `WPF` or `OffScreen`?**
  WPF
- **What steps will reproduce the problem?**
  Here is a simple html code which will cause the problem in the MinimalExample.WPF. The MinimalExample.Winforms seems to work correctly:

`<html>
    <body>
        <p>
            <select>
                <option>1</option>
                <option>2</option>
                <option>3</option>
                </select>
        </p>
    </body>
</html>`
- Open the combobox on the html page
- Move the mouse out of the browser to the windows task bar while the combobox is still open (Moving it to another window also often causes this error)
- Move the mouse back to the open combobox
- The program crashes with an exception: 
  "An unhandled exception of type 'System.ComponentModel.Win32Exception' occurred in WindowsBase.dll"
- **What is the expected output? What do you see instead?**
  Leaving the current window while a combobox is still open shouldn't cause any problems
- **Please provide any additional information below.**
  - debug.log is empty
- **Does this problem also occur in the `Cef` TestApp from http://cefbuilds.com?**
  No

amaitland:
> "An unhandled exception of type 'System.ComponentModel.Win32Exception' occurred in WindowsBase.dll"

Can you provide a symbolized stack trace please. Instructions for downloading `libcef.dll.pdb` are in the troubleshooting wiki article.

> Does this problem also occur in the Cef TestApp from http://cefbuilds.com?
> No

Have you tested this?

amaitland:
Looks like it's purely a `WPF` exception.

```c#
System.ComponentModel.Win32Exception was unhandled
  _HResult=-2147467259
  _message=The operation completed successfully
  HResult=-2147467259
  IsTransient=false
  Message=The operation completed successfully
  Source=WindowsBase
  ErrorCode=-2147467259
  NativeErrorCode=0
  StackTrace:
       at MS.Win32.UnsafeNativeMethods.CreateWindowEx(Int32 dwExStyle, String lpszClassName, String lpszWindowName, Int32 style, Int32 x, Int32 y, Int32 width, Int32 height, HandleRef hWndParent, HandleRef hMenu, HandleRef hInst, Object pvParam)
[...]

Issue #1893, opened 2016-12-19
Feature Request - Release 55.0.0


amaitland:
For those interested in sponsoring a `55.0.0` release then please check [Bounty Source](https://www.bountysource.com/issues/40201096-feature-request-release-55-0-0)

There are a few conditions that need to be met before an official `55.0.0` release can happen.

- The following `CEF` issue needs to be resolved https://bitbucket.org/chromiumembedded/cef/issues/1995/migrate-from-breakpad-to-crashpad
- A stable `CEF` build needs to appear on http://opensource.spotify.com/cefbuilds/index.html (currently the only `2883` builds are based on `Beta` releases of `Chromium`.

If the set `BountySource` target is not met then all is not lost, you can of course build your own `Nuget` packages and host then on your local feed. If another contributor to the project wishes to take charge of the release then by all means say so and I'll happily close this issue.

Constructive comments welcome.

smaranh:
you might also want to add it to the Readme.md.

GrabzIt:
@smaranh great idea!! I have added a PR to try to encourage people to contribute.

amaitland:
Firstly thank you for the amazing show of support! As a gesture here is a link to the `General Usage Guide` I had been working on, it started as a port of https://bitbucket.org/chromiumembedded/cef/wiki/GeneralUsage and now has many `CefSharp` specific details.

https://gist.github.com/amaitland/aa976285407ba35cf895c67c614ba82c

Please review at your leisure, if you have any requests for topics you'd like to see covered then let me know. This `Guide` will be included as part of the release and will replace large parts of the existing `FAQ` (The `Javascript` specific parts at a minimum).

If/when `50%` of the target is reached I will release a new `-pre` set of packages to `Nuget`. The `crashpad` issue listed above would require you to add a small block of code to your application to deal with launching of the new `crashpad` process type. It's relatively trivial, and details are in https://github.com/cefsharp/CefSharp/issues/1876#issuecomment-268422985 Is everyone happy to see a `-pre` release before this issue is resolved? There will be a application crash if the additional code isn't added. Feedback welcome 👍

amaitland:
If/when `75%` of the target is met I will attempt to add `GitLink` support as outlined in #1680 

In theory `GitLink` will allow you to step through the `CefSharp` code hosted on `GitHub` by rewriting the `pdb` source file locations.

jankurianski:
My company appreciates all the effort 👍

[...]

PR #1938, opened 2017-02-14
WPF Moved capture into Opened/Closed events for Popup

taylorjonl:
This commit is an attempt to fix the bug mentioned in this issue:

https://github.com/cefsharp/CefSharp/issues/1723

I modeled this change after the way the WPF ComboBox captures the mouse by looking at the code on referencesource.microsoft.com.  It needs further testing to verify it doesn't break other stuff, hoping to get some more testing from one of my users tomorrow, wanted to start the conversation.

### DIFF
--- a/CefSharp.Wpf/ChromiumWebBrowser.cs
+++ b/CefSharp.Wpf/ChromiumWebBrowser.cs
@@ -469,8 +469,8 @@ namespace CefSharp.Wpf
 
                     if(popup != null)
                     { 
-                        popup.MouseEnter -= PopupMouseEnter;
-                        popup.MouseLeave -= PopupMouseLeave;
+                        popup.Opened -= PopupOpened;
+                        popup.Closed -= PopupClosed;
                         popup = null;
                     }
 
@@ -1633,8 +1633,8 @@ namespace CefSharp.Wpf
                 Placement = PlacementMode.Absolute,
             };
 
-            newPopup.MouseEnter += PopupMouseEnter;
-            newPopup.MouseLeave += PopupMouseLeave;
+            newPopup.Opened += PopupOpened;
+            newPopup.Closed += PopupClosed;
 
             return newPopup;
         }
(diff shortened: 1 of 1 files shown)


amaitland:
> I modeled this change after the way the WPF ComboBox captures the mouse by looking at the code on referencesource.microsoft.com.

Do you have a link to the relevant file on the reference source?

taylorjonl:
Here is the link:  https://referencesource.microsoft.com/#PresentationFramework/src/Framework/System/Windows/Controls/ComboBox.cs,26d4d26579e03147

The basics I learned from there are to capture when the ComboBox is activated and release capture when it is closed and the ComboBox is still capturing.

I think I need to work on this a bit more, I believe I broke use cases where a drag starts in the web browser and ends outside, I will update the pull request shortly.

taylorjonl:
I had a user testing this change for a few days and they didn't notice any issues.  Not sure if there is any further testing you would like me to do or where to go from here?

amaitland:
> I believe I broke use cases where a drag starts in the web browser and ends outside, I will update the pull request shortly.

@taylorjonl What was the end result with this?

amaitland:
`Drop down` lists appear to be broken, unable to select a different element with the mouse.

For testing I used `CefSharp.Wpf.Example`, open `custom://cefsharp/TooltipTest.html` and use one of the two drop down lists.

taylorjonl:
The changes I made to support what I thought was drag'n drop broke it.  I pushed a new commit removing the code in the OnMouse[Up|Down] so it is back to the code I had my user testing.  Drag'n drop still works with the code deleted so I am not really sure the purpose of the code.  Before these changes combo boxes works because whenever the popup was entered it was capturing the mouse but the mouse and popup events were fighting a lot over capture which may be why the crash was occurring.

amaitland:
Code looks a lot cleaner now 👍 Overall I'm thinking it might be just easier to forward the mouse events directly, I'm not sure why `Mouse.Capture` was chosen as a solution. See https://github.com/cefsharp/CefSharp/commit/c0e8e8ef8de7f5bc9ff47093fd4eeba238667b49 (I've merged it already as I'm just about to have lunch and this was a quick and dirty fix).

Feedback welcome.

taylorjonl:
I actually like that solution better, what are the chances you would do a new package against 55 with that change?  The crash related to this capture is #2 on my list of crashes in my application and my managers are trying to chase %99.999 stability rate.

amaitland:
> what are the chances you would do a new package against 55 with that change?

Is your company interested in sponsoring a `55.0.1` release? (`55.0.0` was funded by user contributions see #1893 for more details).

taylorjonl:
This is my personal GitHub account, contact me at jltaylor@ebay.com and we can discuss this further.

PR #1943, opened 2017-02-14
Fixed LoadHtml loading the Url even if RegisterHandler returns false

merceyz:


amaitland:
`merged` 👍 Please add a description next time, thanks.

merceyz:
I didn't do it because the explanation fit into the summary, sorry.

amaitland:
No big deal, just a very long description in the changelog (sometimes they have to be long)

merceyz:
Just for a future reference, how detailed do you want it to be?

amaitland:
> Just for a future reference, how detailed do you want it to be?

Be pragmatic about it, just enough detail so that an outside could read the description and understand what the intent of the `PR` is/was.

PR #1946, opened 2017-02-16
Add Minimal WinForms Designer Support

### LINKED ISSUE #1702: Investigate Delayed Load Dll Linker Option
https://msdn.microsoft.com/en-us/library/151kt790.aspx

Would look something like

```
<DelayLoadDLLs>libcef.dll;%(DelayLoadDLLs)</DelayLoadDLLs>
```

Other project example https://github.com/blackberry/WebKit-BB10/blob/ef835e59b95c47914e1428b7f273b53c1b46184c/Source/WebKit/WebKit.vcxproj/WebKit/WebKitApple.props#L9

Maybe able to Delay load `libcef.dll` long enough to get designer support working, even maybe hosting on `Azure`.

merceyz:
Fixed designer support, resolves #1702

- Implemented the IDisposable interface as per https://msdn.microsoft.com/en-us/library/fs2xkftw(v=vs.110).aspx
  To prevent the default Control.Dispose() from running
  
- Added a parameterless constructor

- Moved the code from the constructor to a Initialize method that gets called in either OnHandleCreated, RegisterJsObject, or RegisterAsyncJsObject as long as the designer isn't active.
  This is because the DesignMode value isn't set until after the Constructor is done so it can't check for the designer in the constructor.

- Added libcef.dll as a "Delay Loaded DLL" in CefSharp.Core

- Made properties with "private set" invisible in the designer

- Removed unused using and sorted them

### DIFF
--- a/CefSharp.WinForms/ChromiumWebBrowser.cs
+++ b/CefSharp.WinForms/ChromiumWebBrowser.cs
@@ -2,11 +2,11 @@
 //
 // Use of this source code is governed by a BSD-style license that can be found in the LICENSE file.
 
-using System;
-using System.Windows.Forms;
 using CefSharp.Internals;
-using CefSharp.ModelBinding;
 using CefSharp.WinForms.Internals;
+using System;
+using System.ComponentModel;
+using System.Windows.Forms;
 
 namespace CefSharp.WinForms
 {
@@ -16,7 +16,7 @@ namespace CefSharp.WinForms
     /// <seealso cref="System.Windows.Forms.Control" />
     /// <seealso cref="CefSharp.Internals.IWebBrowserInternal" />
     /// <seealso cref="CefSharp.WinForms.IWinFormsWebBrowser" />
-    public class ChromiumWebBrowser : Control, IWebBrowserInternal, IWinFormsWebBrowser
+    public class ChromiumWebBrowser : Control, IWebBrowserInternal, IWinFormsWebBrowser, IDisposable
     {
         /// <summary>
         /// The managed cef browser adapter
@@ -30,12 +30,21 @@ namespace CefSharp.WinForms
         /// The browser
         /// </summary>
         private IBrowser browser;
+        /// <summary>
+        /// A flag that indicates whether or not the designer is active
+        /// </summary>
+        private bool designMode;
+        /// <summary>
+        /// A flag that indicates whether or not Initialize() has been called
+        /// </summary>
+        private bool initialized;
 
         /// <summary>
         /// Set to true while handing an activating WM_ACTIVATE message.
         /// MUST ONLY be cleared by DefaultFocusHandler.
         /// </summary>
         /// <value><c>true</c> if this instance is activating; otherwise, <c>false</c>.</value>
+        [Browsable(false), EditorBrowsable(EditorBrowsableState.Never)]
         public bool IsActivating { get; set; }
 
         /// <summary>
@@ -54,11 +63,13 @@ namespace CefSharp.WinForms
         /// <value><c>true</c> if this instance is loading; otherwise, <c>false</c>.</value>
         /// <remarks>In the WPF control, this property is implemented as a Dependency Property and fully supports data
         /// binding.</remarks>
+        [Browsable(false), EditorBrowsable(EditorBrowsableState.Never)]
         public bool IsLoading { get; private set; }
         /// <summary>
         /// The text that will be displayed as a ToolTip
         /// </summary>
         /// <value>The tooltip text.</value>
+        [Browsable(false), EditorBrowsable(EditorBrowsableState.Never)]
         public string TooltipText { get; private set; }
         /// <summary>
         /// The address (URL) which the browser control is currently displaying.
@@ -67,6 +78,7 @@ namespace CefSharp.WinForms
         /// <value>The address.</value>
         /// <remarks>In the WPF control, this property is implemented as a Dependency Property and fully supports data
         /// binding.</remarks>
+        [Browsable(false), EditorBrowsable(EditorBrowsableState.Never)]
         public string Address { get; private set; }
 
         /// <summary>
@@ -233,6 +245,7 @@ namespace CefSharp.WinForms
         /// <value><c>true</c> if this instance can go forward; otherwise, <c>false</c>.</value>
         /// <remarks>In the WPF control, this property is implemented as a Dependency Property and fully supports data
         /// binding.</remarks>
+        [Browsable(false), EditorBrowsable(EditorBrowsableState.Never)]
         public bool CanGoForward { get; private set; }
         /// <summary>
         /// A flag that indicates whether the state of the control current supports the GoBack action (true) or not (false).
@@ -240,6 +253,7 @@ namespace CefSharp.WinForms
         /// <value><c>true</c> if this instance can go back; otherwise, <c>false</c>.</value>
         /// <remarks>In the WPF control, this property is implemented as a Dependency Property and fully supports data
         /// binding.</remarks>
+        [Browsable(false), EditorBrowsable(EditorBrowsableState.Never)]
         public bool CanGoBack { get; private set; }
         /// <summary>
         /// A flag that indicates whether the WebBrowser is initialized (true) or not (false).
@@ -247,6 +261,7 @@ namespace CefSharp.WinForms
         /// <value><c>true</c> if this instance is browser initialized; otherwise, <c>false</c>.</value>
         /// <remarks>In the WPF control, this property is implemented as a Dependency Property and fully supports data
         /// binding.</remarks>
+        [Browsable(false), EditorBrowsable(EditorBrowsableState.Never)]
         public bool IsBrowserInitialized { get; private set; }
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


CefSharp.WinForms/ChromiumWebBrowser.cs
```diff
@@ -273,25 +288,76 @@ private static void OnApplicationExit(object sender, EventArgs e)
...
+        /// </summary>
         /// <param name="address">The address.</param>
-        /// <exception cref="System.InvalidOperationException">Cef::Initialize() failed</exception>
         public ChromiumWebBrowser(string address)
         {
-            if (!Cef.IsInitialized && !Cef.Initialize())
+            Address = address;
+        }
+
+        private void Initialize()
```
amaitland: More descriptive naming please. I'm wondering if this method should only contain the `unmanaged` calls.

Setting `Dock` seems to have vanished, need to double check nothing else has been missed.

CefSharp.WinForms/ChromiumWebBrowser.cs
```diff
@@ -273,25 +288,76 @@ private static void OnApplicationExit(object sender, EventArgs e)
...
+        {
+            if (!initialized)
             {
-                throw new InvalidOperationException("Cef::Initialize() failed");
-            }
+                if (!Cef.IsInitialized && !Cef.Initialize())
```
amaitland: Potentially calling `Cef.Initialize` during `Handle` creation seems questionable to me.

CefSharp.WinForms/ChromiumWebBrowser.cs
```diff
@@ -273,25 +288,76 @@ private static void OnApplicationExit(object sender, EventArgs e)
...
+                    FocusHandler = new DefaultFocusHandler(this);
+                }
+
+                if (ResourceHandlerFactory == null)
+                {
+                    ResourceHandlerFactory = new DefaultResourceHandlerFactory();
+                }
 
-            Dock = DockStyle.Fill;
+                if (BrowserSettings == null)
```
amaitland: Moving this opens up the potential for null references exceptions.

Previously you could write `browser.BrowserSettings.WhateverProperty = value`

I understand the reason for moving this, just needs to be documented. Probably no point setting it in a delayed fashion.

CefSharp.WinForms/ChromiumWebBrowser.cs
```diff
@@ -273,25 +288,76 @@ private static void OnApplicationExit(object sender, EventArgs e)
...
+            else
+            {
+                base.Dispose(false);
+            }
+        }
 
-            FocusHandler = new DefaultFocusHandler(this);
-            ResourceHandlerFactory = new DefaultResourceHandlerFactory();
-            BrowserSettings = new BrowserSettings();
+        public new void Dispose()
```
amaitland: What's the purpose exactly of implementing `IDisposable`? Was there a problem adding the design time checking logic to the existing `protected override void Dispose(bool disposing)`?

CefSharp.WinForms/ChromiumWebBrowser.cs
```diff
@@ -273,25 +288,76 @@ private static void OnApplicationExit(object sender, EventArgs e)
...
+        {
+            if (!designMode)
```
amaitland: If this was to stay then the condition should be switched around, `if(designMode)` flows cleaner rather than `if(!designMode)`

CefSharp.WinForms/ChromiumWebBrowser.cs
```diff
@@ -376,6 +442,11 @@ public void RegisterJsObject(string name, object objectToBind, BindingOptions op
                                     "called before the underlying CEF browser is created.");
             }
 
+            if (!initialized)
```
amaitland: Littering the code with this check and method call is far from ideal. I do understand the reason for delayed creation of unmanaged objects. If it has to be, then let's more the check into the method it's self.

CefSharp/WebBrowserExtensions.cs
```diff
@@ -245,9 +245,10 @@ public static void LoadString(this IWebBrowser browser, string html, string url)
         /// <param name="browser">The ChromiumWebBrowser instance this method extends</param>
         /// <param name="html">The HTML content.</param>
         /// <param name="url">The URL that will be treated as the address of the content.</param>
-        public static void LoadHtml(this IWebBrowser browser, string html, string url)
+        /// <returns>returns false if the Url was not successfully parsed into a Uri</returns>
```
merceyz: Yeah, I noticed that.

Followed https://help.github.com/articles/syncing-a-fork/

CefSharp.WinForms/ChromiumWebBrowser.cs
```diff
@@ -30,12 +30,21 @@ public class ChromiumWebBrowser : Control, IWebBrowserInternal, IWinFormsWebBrow
         /// The browser
         /// </summary>
         private IBrowser browser;
+        /// <summary>
+        /// A flag that indicates whether or not the designer is active
+        /// </summary>
+        private bool designMode;
```
merceyz: We do, sadly DesignMode becomes false by the time we get to the destructor/dispose

CefSharp.WinForms/ChromiumWebBrowser.cs
merceyz: Keeping the `managed` calls in the constructor results in the designer crashing.

Added it back, I figured the user would set it in the designer. But seems you want it to be `DockStyle.Fill` by default

amaitland: > It initializes stuff, i'm not a creative person. Name suggestion?

It doesn't need to be creative, just descriptive. 

> Keeping the managed calls in the constructor results in the designer crashing.

The purely managed classes in the `CefSharp` assembly?

> But seems you want it to be DockStyle.Fill by default

What about everyone creating the control in code as they do now? Needs to be the same as it is now.

CefSharp.WinForms/ChromiumWebBrowser.cs
amaitland: Ok no worries. That sort of detail needs to be in the comment.

CefSharp.WinForms/ChromiumWebBrowser.cs
merceyz: > The purely managed classes in the CefSharp assembly?
```
new DefaultFocusHandler(this);
new DefaultResourceHandlerFactory();
new BrowserSettings();
new ManagedCefBrowserAdapter(this, false);
```
Keeping these in the constructor causes it to crash

> What about everyone creating the control in code as they do now? Needs to be the same as it is now.
Since the constructor that is currently in use doesn't get called by the designer I made it effectively the same

CefSharp.WinForms/ChromiumWebBrowser.cs
merceyz: It now only gets delayed if created by the parameterless constructor

CefSharp.WinForms/ChromiumWebBrowser.cs
merceyz: I only had to prevent the default Control.Dispose() from getting called.
I removed the `IDisposable` as it can be done without it.

> Was there a problem adding the design time checking logic to the existing protected override void Dispose(bool disposing)?

Yes, it crashed when I checked it in that function.
So I could either move the code out of that function or replace the default Dispose()

CefSharp.WinForms/ChromiumWebBrowser.cs
```diff
@@ -409,14 +486,29 @@ public void RegisterAsyncJsObject(string name, object objectToBind, BindingOptio
...
+                // NOTE: Had to move the code out of this function otherwise the designer would crash
+                parentCheck();
+
+                ResizeBrowser();
             }
 
             base.OnHandleCreated(e);
         }
 
+        private void parentCheck()
```
merceyz: > Did you check this in Release mode? I'd be adding NoInlining to this, make sure the compiler doesn't optimise.

I did after adding `NoInlining`

CefSharp.WinForms/ChromiumWebBrowser.cs
```diff
@@ -666,16 +761,16 @@ protected override bool IsInputKey(Keys keyData)
                 case Keys.Up:
                 case Keys.Down:
                 case Keys.Tab:
-                {
```
merceyz: Done

CefSharp.WinForms/ChromiumWebBrowser.cs
amaitland: > Keeping these in the constructor causes it to crash

The last two I expect to cause problems, the first two are purely managed, so I'd surprised if they cause problems.

CefSharp.WinForms/ChromiumWebBrowser.cs
amaitland: > So I could either move the code out of that function or replace the default Dispose()

Extracting into a function sounds better than forcibly overriding the default dispose

CefSharp.WinForms/ChromiumWebBrowser.cs
amaitland: Thx

amaitland: Spaces not tabs?

merceyz: > Spaces not tabs?

Yeah, visual studio uses spaces by default

CefSharp.WinForms/ChromiumWebBrowser.cs
merceyz: > The last two I expect to cause problems, the first two are purely managed, so I'd surprised if they cause problems.

Consider yourself surprised

CefSharp.WinForms/ChromiumWebBrowser.cs
amaitland: `VS` uses `tabs` by default for `C#` files  unless you've changed the default.

`GitHub` still thinks the formatting has changed, usually it's when you've used `Tabs` instead of spaces.

amaitland:
As `System.ComponentModelComponent` (which `Control` inherits) already implements the Dispose pattern my preference is to have a non inlined method in the `Dispose override` we already have. 

This is assuming that keeps the designer happy.

amaitland:
Also, how are you testing these changes? Do you have an example?

CefSharp.WinForms/ChromiumWebBrowser.cs
merceyz: > Extracting into a function sounds better than forcibly overriding the default dispose

Tried now by moving it out of the function and removing the Destructor and Dispose(), this caused a crash

CefSharp.WinForms/ChromiumWebBrowser.cs
merceyz: > VS uses tabs by default for C# files unless you've changed the default.

I can't remember changing that and mine uses spaces

> GitHub still thinks the formatting has changed, usually it's when you've used Tabs instead of spaces.

I swear it's spaces

merceyz:
> This is assuming that keeps the designer happy.

It doesn't

> Also, how are you testing these changes? Do you have an example?

I just drag and drop it to the form to test the creation (or double click) then delete it to check the dispose pattern

CefSharp.WinForms/ChromiumWebBrowser.cs
amaitland: Leave it for now, I'll fix it later.

amaitland:
> It doesn't

Technically I cannot think of a reason it's not feasible. 

I've merged this and added https://github.com/cefsharp/CefSharp/commit/a9c29ae279cce60d14fd80a53778c64516dc2313

Give it a try when you have a chance, it implements my `Dispose` idea changes.

amaitland:
Format a method call see also https://github.com/cefsharp/CefSharp/commit/c827af48b344be81f48f2f6720815a76e62009be

merceyz:
> Technically I cannot think of a reason it's not feasible.
> Give it a try when you have a chance, it implements my Dispose idea changes.

I can't get it to work with your changes,

Also to get it to work in x64 `CefSharp.Winforms` has to be built as `Any CPU`

amaitland:
It needs to be built in both x86 and x86. 

I'll look at it again later, tested OK in vs2013

merceyz:
> I can't get it to work with your changes.

Re-cloned the project, now it works.
`Github Desktop` said there was no local changes so seems like something was corrupted

Doesn't work when built against x64 though, since the designer (x86) can't load `CefSharp.Winforms` (x64)

amaitland:
> Doesn't work when built against x64 though, since the designer (x86) can't load CefSharp.Winforms (x64)

That makes sense. Anyone using the official `Nuget` packages should be able to target `x86` or `AnyCPU` no problem. The examples within the main project are hard coded currently. Wasn't planning on changing it.

PR #1953, opened 2017-02-20
Updated OnTooltipChanged to allow for modifying text

### LINKED ISSUE #1648: Feature Request - Update OnTooltipChanged to allow for modifying text
Update `IDisplayHandler.OnTooltipChanged` so that `text` can be modified (make `ref` param)

merceyz:
Resolves #1648

- Made it possible to modify the `text` parameter

- Updated the examples accordingly

- Added a DisplayHandler to the WPF example

### DIFF
--- a/CefSharp.Core/Internals/ClientAdapter.cpp
+++ b/CefSharp.Core/Internals/ClientAdapter.cpp
@@ -321,25 +321,27 @@ namespace CefSharp
         {
             auto tooltip = StringUtils::ToClr(text);
             bool hasChanged = tooltip != _tooltip;
+            bool returnFlag = true;
 
             //NOTE: Only called if tooltip changed otherwise called many times
-            // also only appears to be called with OSR rendering at the moment
-            if(hasChanged)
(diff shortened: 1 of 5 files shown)


amaitland:
> When window rendering is disabled the application is responsible for drawing tooltips and the return value is ignored. 

http://magpcss.org/ceforum/apidocs3/projects/(default)/CefDisplayHandler.html#OnTooltip(CefRefPtr%3CCefBrowser%3E,CefString&)

The documentation implies it should be called for normal rendering, will need to be looked into further.

merceyz:
Not according to the issue I linked to in the source, one could of course ask Marshall

amaitland:
That issue hasn't been updated in years. I'd first look through the CEF source before asking.

merceyz:
Issue recently updated, see https://bitbucket.org/chromiumembedded/cef/issues/783/add-cefdisplayhandlerontooltip-call-for#comment-36392203

amaitland:
Guess that clarifies things. If you see value in this for the `WPF` version then by all means merge away. Should probably update #1648 as well.

Issue #1889, opened 2016-12-15
Black Lines when using CefSharp.WinForms in a WPF WindowsFormHost


RichiCoder1:
- **What version of the product are you using?**
  - CefSharp.Common, CefSharp.Wpf, CefSharp.WinForms all version 53.

- **What architecture x86 or x64?**
  - x86

- **On what operating system?**
    - Win10

- **Are you using `WinForms`, `WPF` or `OffScreen`?**
  - WinForms & WPF

- **What steps will reproduce the problem?**
  - Forked project here: https://github.com/RichiCoder1/CefSharp.MinimalExample
  - Run CefSharp.WpfAndWinForms
  - Slowly resize the window. Notice the black line at the bottom of the browser window appear and disappear.

- **What is the expected output? What do you see instead?**
  - Expected: ![Expected](http://i.imgur.com/BU01wrm.png)
  - Actual: ![Actual](http://i.imgur.com/Mo8kOs9.png)
Note the black line between the bottom bar and the browser (and ignore the black outlines, that just my terrible screenshotting)

- **Does this problem also occur in the `CEF` Sample Application from http://opensource.spotify.com/cefbuilds/index.html (http://cefbuilds.com for older version)?**
  - Honestly not sure how to test this.

I realize this isn't a hugely well supported scenario, but it's the only was to have the browser content be accessibility friendly in WPF. I notice that when I resize the Window, or unload the control containing the browser, there will be a flash of black, so this may just be a flaw or shortcoming of WindowsFormHost. I can't seem to find anyone else using WindowsFormHost having the black outline issue.

amaitland:
How does Chrome behave on your system? What about the WinForms example? You can defer updates until resize has completed if that's preferable. 

Have you tried on different hardware? 

Is this a bug in CefSharp? Probably not, you can always use CefGlue or ChromiumFx for comparison. They're all just wrappers around the Chromium embedded framework.

RichiCoder1:
>How does Chrome behave on your system? 

Chrome behaves perfectly, which is part of the reason I'm baffled.

>What about the WinForms example? You can defer updates until resize has completed if that's preferable.

I thought the WinForms example was fine, but it has the faint black lines when resized to certain sizes too. 

> Is this a bug in CefSharp? Probably not, you can always use CefGlue or ChromiumFx for comparison. They're all just wrappers around the Chromium embedded framework.

I'll pull or throw together a basic test app using one of those and see what I turn up.

[...]

PR #1962, opened 2017-02-23
WPF - Made GetViewRect always round up

merceyz:
See https://github.com/cefsharp/CefSharp/issues/1889#issuecomment-281741492

- Made GetViewRect always round up to prevent the image from being smaller than the control


amaitland:
I'm happy to `merge` this and see how it goes 👍

PR #1966, opened 2017-02-26
Updated the copyright year and made sure it uses the same format everywhere

merceyz:
- Updated the copyright year from 2016 to 2017 and made sure the same format is used everywhere


amaitland:
Thanks 👍

merceyz:
I didn't think this would be a problem but I ran a regex on the `CefSharp3.sln` project to update the copyright on the files in said project. 
Since `CefSharp3.sln` does not include `CefSharp.WinForms.Test` and `CefSharp.Test` they were not affected

amaitland:
> I didn't think this would be a problem but I ran a regex on the CefSharp3.sln project to update the copyright on the files in said project.

Absolutely fine 😄

PR #1979, opened 2017-03-08
Allow calling managed methods from objects bound with RegisterAsyncJsObject that return arrays

joaompneves:
**Problem:**
When using RegisterAsyncJsObject to bind a C# object it was not possible to call managed methods that return arrays, due to a crash in the browser sub process.

**Details:**
CefV8Value::CreateArray in DeserializeV8Object expects prior call to Enter of the V8 context, and without that call, CreateArray returned null.

**Fix:**
Call V8 context Enter in OnProcessMessageReceived from CefAppUnmanagedWrapper.

### DIFF
--- a/CefSharp.BrowserSubprocess.Core/CefAppUnmanagedWrapper.cpp
+++ b/CefSharp.BrowserSubprocess.Core/CefAppUnmanagedWrapper.cpp
@@ -396,7 +396,7 @@ namespace CefSharp
         {
             auto frameId = GetInt64(argList, 0);
             auto callbackId = GetInt64(argList, 1);
-            
+
             JavascriptRootObjectWrapper^ rootObjectWrapper;
             browserWrapper->JavascriptRootObjectWrappers->TryGetValue(frameId, rootObjectWrapper);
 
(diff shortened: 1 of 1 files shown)


CefSharp.BrowserSubprocess.Core/CefAppUnmanagedWrapper.cpp
```diff
@@ -405,17 +405,40 @@ namespace CefSharp
...
+                                if (success)
+                                {
+                                    callback->Success(DeserializeV8Object(argList, 3));
+                                }
+                                else
+                                {
+                                    callback->Fail(argList->GetString(3));
+                                }
+                                //dispose
+                                delete callback;
```
merceyz: If it somehow fails this wont be called, you should place it inside the finally block

joaompneves: Didn't change that bit, but will do it.

amaitland:
Need a better description for this `PR`. If I was to look at the change log and see `Fixed problem when calling CefV8Value::CreateArray` that doesn't really mean a whole heap. At a minimum need to make reference to it being related to `RegisterAsyncJsObject `

amaitland:
This PR is redundant

PR #1989, opened 2017-03-17
Added minimal WPF designer support

merceyz:
- Fixed WPF designer support

### DIFF
--- a/CefSharp.Wpf/ChromiumWebBrowser.cs
+++ b/CefSharp.Wpf/ChromiumWebBrowser.cs
@@ -45,7 +45,7 @@ namespace CefSharp.Wpf
         /// <summary>
         /// The tool tip
         /// </summary>
-        private readonly ToolTip toolTip;
+        private ToolTip toolTip;
         /// <summary>
         /// The managed cef browser adapter
         /// </summary>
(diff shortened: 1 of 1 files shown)


CefSharp.Wpf/ChromiumWebBrowser.cs
```diff
@@ -1473,7 +1502,7 @@ private void RemoveSourceHook()
         /// <returns>bool to indicate if browser was created. If the browser has already been created then this will return false.</returns>
         protected virtual bool CreateOffscreenBrowser(Size size)
         {
-            if (browserCreated || System.ComponentModel.DesignerProperties.GetIsInDesignMode(this) || size.IsEmpty || size.Equals(new Size(0, 0)))
+            if (browserCreated || size.IsEmpty || size.Equals(new Size(0, 0)))
```
amaitland: Is there a reason we're no longer checking design mode here?

merceyz: `CreateOffscreenBrowser` gets called from `OnActualSizeChanged` which isn't registered when in design mode.

amaitland: Makes sense 👍

amaitland:
All looks pretty straight forward 👍

Issue #2001, opened 2017-03-30
Encoding not working on larger files with custom scheme handler


FrankLeinbach:
I am attaching a sample page that causes this problem.  If you change the CefSharp.WinForms.Example.BrowserTabControl, adding 

BrowserSettings = new BrowserSetting() {
  DefaultEncoding="UTF-8"
}

And then add this page to the CefSharp.Example Resources (such that you can navigate to custom://cefsharp/test.html) and ensure not to cut and paste this into the Visual Studio enviroment, the Japanese characters will be mangled to look like ASCII.  This works in CefSharp 53 and below and stopped working in 55.

It appears to be related to the size of the file, which is the reason for the overly spacious and large stylesheet within this html (our html files are much larger in content).  There may be some sort of buffer overrun issue going on or some variable as it appears to occur when the file exceeds 32,955 bytes.  If you fix the file to be equal or under 32,955 bytes, the characters appear as they should.

This appears to only be related to the custom scheme handler.

amaitland:
> I am attaching a sample page that causes this problem. If you change the

No binaries as per https://github.com/cefsharp/CefSharp/blob/master/CONTRIBUTING.md#help-us-help-you

Somehow you seem to have skipped over the issue template, please read then fill out the bug report.
https://github.com/cefsharp/CefSharp/blob/master/ISSUE_TEMPLATE.md#bug-report

To reproduce your problem, best to fork the project and push your changes to `GitHub`.

FrankLeinbach:
What version of the product are you using?

57.0.0 (also applies to 55.0.0)

What version are you using? Nuget? CI Nuget? build from a branch? If so which branch? e.g. 53.0.1
What architecture x86 or x64?

Using Nuget cef redist version 3.2987.1597 on x86

On what operating system?

Windows 10

Are you using WinForms, WPF or OffScreen?

WinForms

What steps will reproduce the problem?

Get the project from https://github.com/FrankLeinbach/CefSharp/tree/cefsharp/57

(this is my first git upload, so hopefully this worked correctly).

I changed BrowserTabUserControl to setup the BrowserSettings with a DefaultEncoding of UTF-8.

[...]

PR #2002, opened 2017-03-31
Custom Scheme Problem

FrankLeinbach:
Problem with custom scheme handler and double byte languages when the
page is over a certain size


merceyz:
This belongs to issue #2001

PR #2007, opened 2017-04-05
Expanded proxy support, made it easier to set proxy settings

merceyz:
- Added a ProxyOptions class
- Added a CefSettings.Proxy property
- Made ClientAdapter::GetAuthCredentials take care of the authentication for the proxy


amaitland:
Will look at this sometime next week 👍

CefSharp/ProxyOptions.cs
```diff
@@ -0,0 +1,63 @@
...
+        /// The list of domains that shouldn't be affected by the proxy, Format: example.com;example2.com
+        /// </summary>
+        public string BypassList { get; private set; }
+
+        /// <param name="ip">The IP address for the proxy</param>
+        /// <param name="port">The port for the proxy</param>
+        /// <param name="username">The username required for authentication</param>
+        /// <param name="password">The password required for authentication</param>
+        /// <param name="bypassList">The list of domains that shouldn't be affected by the proxy, Format: example.com;example2.com</param>
+        public ProxyOptions(string ip, string port, string username, string password, string bypassList)
```
amaitland: Can we just use default values and get rid of the overloads?

CefSharp.Core/Internals/ClientAdapter.cpp
```diff
@@ -732,6 +733,16 @@ namespace CefSharp
         bool ClientAdapter::GetAuthCredentials(CefRefPtr<CefBrowser> browser, CefRefPtr<CefFrame> frame, bool isProxy,
             const CefString& host, int port, const CefString& realm, const CefString& scheme, CefRefPtr<CefAuthCallback> callback)
         {
+            if (isProxy && Cef::proxyOptions != nullptr && Cef::proxyOptions->IP == StringUtils::ToClr(host) && !String::IsNullOrEmpty(Cef::proxyOptions->Username)
```
amaitland: What if you wanted to provide a different `username` and `password` after you've initialized `CEF`? 

I'd rather not having additional code here, specially that adds a new behavior. 

It probably makes more sense to provide a default implementation of `IRequestHandler` that contains this logic. (If we do go down that path, probably should add default implementations for all the handlers.)

CefSharp.Core/CefSettings.h
```diff
@@ -368,6 +369,15 @@ namespace CefSharp
         }
 
         /// <summary>
+        /// The proxy options that will be used for all connections
+        /// </summary>
+        property ProxyOptions^ Proxy
```
amaitland: I'd rather see this as part of `CefSharpSettings`.

CefSharp/ProxyOptions.cs
merceyz: Done

CefSharp.Core/CefSettings.h
merceyz: Done

CefSharp.Core/Internals/ClientAdapter.cpp
merceyz: > What if you wanted to provide a different username and password after you've initialized CEF?

That's now possible by doing `CefSharpSettings.Proxy.Username = "x"` and `CefSharpSettings.Proxy.Username = "y"`

I don't quite agree with you, I'm hoping this can stay `as is` so that anyone using the `IRequestHandler` doesn't have to implement this logic themself

CefSharp/CefSharpSettings.cs
```diff
@@ -41,5 +41,10 @@ static CefSharpSettings()
         /// the event handlers are hooked in the static constructor for the ChromiumWebBrowser class
         /// </summary>
         public static bool ShutdownOnExit { get; set; }
+
+        /// <summary>
+        /// The proxy options that will be used for all connections
```
amaitland: A think we should expand on this, explain what's going on. Command line arguments are set, if you set username/password then `GetAuthCredentials` will be called automatically. Also note that if you use this option it's impossible to change the proxy at dynamically at runtime.

CefSharp.Core/Internals/ClientAdapter.cpp
```diff
@@ -732,6 +732,16 @@ namespace CefSharp
         bool ClientAdapter::GetAuthCredentials(CefRefPtr<CefBrowser> browser, CefRefPtr<CefFrame> frame, bool isProxy,
             const CefString& host, int port, const CefString& realm, const CefString& scheme, CefRefPtr<CefAuthCallback> callback)
         {
+            if (isProxy && CefSharpSettings::Proxy != nullptr && CefSharpSettings::Proxy->IP == StringUtils::ToClr(host) && !String::IsNullOrEmpty(CefSharpSettings::Proxy->Username)
+                && !String::IsNullOrEmpty(CefSharpSettings::Proxy->Password))
+            {
+                auto frameWrapper = gcnew CefFrameWrapper(frame);
```
amaitland: Seems unnecessary to create these wrappers, can just execute the callback directly.

CefSharp.Core/Internals/ClientAdapter.cpp
```diff
@@ -732,6 +732,16 @@ namespace CefSharp
         bool ClientAdapter::GetAuthCredentials(CefRefPtr<CefBrowser> browser, CefRefPtr<CefFrame> frame, bool isProxy,
             const CefString& host, int port, const CefString& realm, const CefString& scheme, CefRefPtr<CefAuthCallback> callback)
         {
+            if (isProxy && CefSharpSettings::Proxy != nullptr && CefSharpSettings::Proxy->IP == StringUtils::ToClr(host) && !String::IsNullOrEmpty(CefSharpSettings::Proxy->Username)
```
amaitland: What about creating a `HasUsernameAndPassword()` method (or property) to `ProxySettings`, make this a little cleaner, little less `C++`

CefSharp/CefSharpSettings.cs
amaitland: Also say that `GetAuthCredentials` won't be called, for a proxy server that matches the `IP`

amaitland:
@merceyz I've granted you access to all of the `CefSharp` repositories (assuming you accept the invitation of course).

Once your happy this is ready to release into the wild then feel free to merge yourself 👍

CefSharp/CefSharpSettings.cs
merceyz: Done

CefSharp.Core/Internals/ClientAdapter.cpp
merceyz: Done

CefSharp.Core/Internals/ClientAdapter.cpp
merceyz: Done

merceyz:
> @merceyz I've granted you access to all of the CefSharp repositories (assuming you accept the invitation of course).

Thank you, though I haven't gotten an invitation yet I appreciate it.

amaitland:
> I haven't gotten an invitation yet

Check https://github.com/cefsharp

Tried sending you another one, would only let me edit the existing invite, so it's def been sent.

merceyz:
> Check https://github.com/cefsharp

There it was, cheers

> Tried sending you another one, would only let me edit the existing invite, so it's def been sent.

Didn't get an E-mail nor a notification, so I suppose one would have to know the invite was sent and where to find it

amaitland:
> Didn't get an E-mail nor a notification, so I suppose one would have to know the invite was sent and where to find it

It said an `email` was sent, maybe it got lost along the way.

merceyz:
> It said an email was sent, maybe it got lost along the way.

Probably

Would you like this to be a part of `57`?

amaitland:
> Would you like this to be a part of 57?

I'm reluctant to merge anything other than bug fixes this late in the release process. Plan is to generate the `57.0.0` release packages early next week.

PR #2010, opened 2017-04-07
Made it possible to register a "one-time use" ResourceHandler

merceyz:
- Added a `ResourceFactoryItem` class
- Added a overload to `RegisterHandler` that takes a bool representing whether or not the handler should be used only once
- Made `GetResourceHandler` remove the `ResourceFactoryItem` if it was specified as a `one-time use` handler
- Made LoadHtml register a `one-time use` handler

### DIFF
--- a/CefSharp/DefaultResourceHandlerFactory.cs
+++ b/CefSharp/DefaultResourceHandlerFactory.cs
@@ -18,7 +18,7 @@ namespace CefSharp
         /// <summary>
         /// Resource handler thread safe dictionary
         /// </summary>
-        public ConcurrentDictionary<string, IResourceHandler> Handlers { get; private set; }
+        public ConcurrentDictionary<string, ResourceFactoryItem> Handlers { get; private set; }
 
         /// <summary>
         /// Create a new instance of DefaultResourceHandlerFactory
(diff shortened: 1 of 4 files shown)


amaitland:
Follow up commit is https://github.com/cefsharp/CefSharp/commit/d03e4a44bdc8f911d81d745cc8a10964ca7b1424

Are you in a hurry to see this merged into `cefsharp/57`?

CefSharp/DefaultResourceHandlerFactory.cs
```diff
@@ -18,43 +18,57 @@ public class DefaultResourceHandlerFactory : IResourceHandlerFactory
...
             return false;
         }
 
         /// <summary>
+        /// Register a persistant handler for the specified Url
+        /// </summary>
+        /// <param name="url">url</param>
+        /// <param name="handler">handler</param>
+        /// <returns>returns true if the Url was successfully parsed into a Uri otherwise false</returns>
+        public virtual bool RegisterHandler(string url, IResourceHandler handler)
```
amaitland: Removed unnecessary overload, can use a default value instead.

CefSharp/ResourceFactoryItem.cs
```diff
@@ -0,0 +1,27 @@
+﻿// Copyright © 2010-2017 The CefSharp Authors. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style license that can be found in the LICENSE file.
+
+namespace CefSharp
+{
+    public class ResourceFactoryItem
```
amaitland: Renamed class, as it's only used in the context of `DefaultResourceHandlerFactory`.

amaitland:
Marking this as `breaking-change`, `DefaultResourceHandlerFactory.Handlers` was publicly accessible, and it's changed type. It's unlikely anyone was using it directly, it was however possible.

CefSharp/WebBrowserExtensions.cs
```diff
@@ -309,7 +309,7 @@ public static bool LoadHtml(this IWebBrowser browser, string html, string url, E
                 throw new Exception("LoadHtml can only be used with the default IResourceHandlerFactory(DefaultResourceHandlerFactory) implementation");
             }
 
-            if (resourceHandler.RegisterHandler(url, ResourceHandler.FromString(html, encoding, true)))
+            if (resourceHandler.RegisterHandler(url, ResourceHandler.FromString(html, encoding, true), false))
```
amaitland: Whilst this change makes sense it is however a subtle change in behavior and could potentially bring some existing users unstuck. There's now a `oneTimeUse` param with a default value as part of `LoadHtml`.

merceyz:
> Are you in a hurry to see this merged into cefsharp/57?

I build CefSharp from master so that's all up to you :+1:

PR #2012, opened 2017-04-08
Added support for PDF scaling

GrabzIt:
This connects up the PDF scaling however it doesn't seem to scale the PDF. So I think the key used to scale the PDF in CEF maybe wrong.


amaitland:
> So I think the key used to scale the PDF in CEF maybe wrong.

Hmm, that is unfortunate. This has been merged into `master`. I might hold off on merging into `57` until it's working.

GrabzIt:
It mustn't have been released properly in version 57. It now works in version 59. This case can be marked as resolved.

Issue #1915, opened 2017-01-17
Right click menu is opened in the left corner


shenhavba:
Bug Report

**What version of the product are you using?**

CefSharp.Common, CefSharp.Wpf all version 55

**What version are you using? Nuget? CI Nuget? build from a branch? If so which branch? e.g. 53.0.1
What architecture x86 or x64?**

build from branch 55.0.0-pre01 - x86

**On what operating system?**

windows 10

**Are you using WinForms, WPF or OffScreen?**

wpf

**What steps will reproduce the problem?**

     1. open the browser(not in full size)
     2.right click on the browser page - the menu is opened outside the browser and can't be clolsed.

**Please provide detailed information here, enough for someone else to reprodce your problem.
Please no binary (zip, etc) links, fork the MinimalExample and push your changes to GitHub. (Alternatively use a code sharing service list Gist or Pastebin).
What is the expected output? What do you see instead?**

The expected behavior is that the right click menu will open inside the bowser area and when clicking on the browser again, the menu should be closed.

Its working fine in version 47.

**Does this problem also occur in the CEF Sample Application from http://opensource.spotify.com/cefbuilds/index.html?**

No.

merceyz:
https://github.com/cefsharp/CefSharp/blob/master/ISSUE_TEMPLATE.md

amaitland:
The menu unable to be closed is #1795 

In reality you probably need a custom `WPF` context menu to resolve this, which is blocked on #1767 for which someone will actually need to take ownership and see the issue resolved in `CEF`.

ashu3011:
You can see this happening in the CefSharp  minimal example. Basically the offset is taken from the corner of the browser window but the menu is displayed using the same offset from the Monitor top left.

amaitland:
A complete implementation of `CefRenderHandler::GetScreenPoint` is required to fix this. Needs to handle `DPI` scaling as well.

amaitland:
The experimental implementation I had previous commented out is now available for testing on `master`.

https://github.com/cefsharp/CefSharp/commit/ef27fb5cf0165985b1299a8316e49e48742ee576

Testing at different DPIs is required. To dismiss the menu it's important to see #1795

PR #2022, opened 2017-04-14
Get zoom level from the browser host instead

### LINKED ISSUE #1778: Zooming Reset Problem
# 

Zooming Reset Problem

**Version :** 45.0.0
**Architecture :** x86
**Operating System :** Win 8
**Platform :** WPF

**Problem :** 
**Step 1 :** Extract the attached sample project and add the cefshrap wpf 45.0.0 package.
**Step 2 :** Run the project.
**Step 3 :** Right click and Click the Zoom In Browser 1. It's automatically Zoom In on Browser 2.
**Step 4 :** Click the Browser 2 tab and right click. Select Reset. The Browser not reset. If I go to the Browser 1 and click the reset option. It's reset both the browser.

**Expected Output:**
If I Zoom In the Browser 1, The Browser 2 also Zoom In. It's Fine.
If I Reset the Browser 2 The both Browser need to be reset.
And also I have added the Mouse Wheel with Ctrl button Zoom In. It's working fine. But when Zoom In or Out the Scrollbar also moving. The Zoom In and Zoom out only working when we focused the browser not working on mouse over.

merceyz:
Resolves #1778 

- Get zoom level from the browser host instead of `ZoomLevelProperty`

### DIFF
--- a/CefSharp.Wpf/ChromiumWebBrowser.cs
+++ b/CefSharp.Wpf/ChromiumWebBrowser.cs
@@ -1157,7 +1157,7 @@ namespace CefSharp.Wpf
         {
             get
             {
-                using (var getZoomLevel = this.GetZoomLevelAsync())
+                using (var getZoomLevel = browser.GetHost().GetZoomLevelAsync())
                 {
                     getZoomLevel.Wait();
                     return getZoomLevel.Result;


amaitland:
Is it safe to Wait for the task to execute? I really dislike Wait, I've seen far too many deadlocks

merceyz:
In my tests it didn't deadlock, it can of course happen.

In my application (background process doing work) i use Wait on JS execution and haven't seen a deadlock yet

amaitland:
I'm still not keen on this. It's also we would be blocking the UI thread.

merceyz:
Another solution is to have the zoom commands send the new zoom value to all other browser instances on the same domain

amaitland:
Is a fix actually required? The simplest workaround is to isolate each browser and let them zoom independently.

merceyz:
If one wants the `shared domain zoom` to work properly yes, if not, no

amaitland:
How about providing an example of your idea as an attached property? Get some real world feedback. There is nothing stopping a user from implementing their own solution.

merceyz:
> How about providing an example of your idea as an attached property?

I'm not quite sure what you mean, could you attempt to clarify?

amaitland:
Using this in conjunction with my proposed fix for #1915 causes a deadlock. The simplest fix may well be updating `ZoomLevelProperty` when the browser becomes visible. Something like:

```c#
if(isVisible)
{
	browser.GetHost().GetZoomLevelAsync().ContinueWith(previous =>
	{
		if (!IsDisposed)
		{
			SetCurrentValue(ZoomLevelProperty, previous.Result);
		}
	}, TaskContinuationOptions.OnlyOnRanToCompletion);
}
```

The `ZoomReset` issue still needs to be addressed, changing `ZoomLevelProperty` from using the property changed to the `CoerceValueCallback` is probably the easiest fix (always called even if the values are the same).

> EDIT:
> I could make CefBrowserHostWrapper::GetZoomLevelOnUI() public, then the task wouldn't be needed

Adding `IBrowserHost.GetZoomLevel` is fine with me. I think the `GetZoomLevelOnUI` method should be `private` with a new `public` method being added that includes the standard disposed checks. You would only be able to use that version when `MultiThreadedMessageLoop = false` though.

merceyz:
I don't have any plans to do anything with this PR, so i'm closing this.

PR #2025, opened 2017-04-19
Support serialization of CLR Classes

joaompneves:
Bound objects (using RegisterAsyncJsObject) now can have methods that return Structs and Classes (before only structs were allowed). This PR introduces support for classes (returned to the JS side) serialization (only public fields will be serialized). Check the Binding examples: asyncStructsArray and asyncClassesArray.


amaitland:
> Support CLR Classes serialization when before only Structs were allowed

In what context? Assuming you mean allow `IJavascriptCallback` to use a classes? Please be more descriptive, if you weren't the author of this `PR` would you be able to read the description and know what the changes were did?

> CLR Classes can now also be returned to the JS side.

What does this mean exactly?

CefSharp/ModelBinding/ModelBindingExtensions.cs
```diff
@@ -36,7 +36,7 @@ public static T CreateInstance<T>(this Type type, bool nonPublic = false)
         /// <param name="nonPublic"><see langword="true"/> if a non-public constructor can be used, otherwise <see langword="false"/>.</param>
         public static object CreateInstance(this Type type, bool nonPublic = false)
         {
-            return CreateInstanceInternal(type, nonPublic);
+            return Activator.CreateInstance(type, nonPublic);
```
amaitland: What is the purpose of this change? How does it relate to `structs`?

amaitland:
Please provide an example that demos your changes.

CefSharp/ModelBinding/ModelBindingExtensions.cs
joaompneves: Structs have implicit public parameterless constructors and the CreateInstanceInternal wasn't able to find that constructor. So I replaced with the Activator.CreateInstance which does the job right.

amaitland: Much clearer, thank you.

amaitland:
Will look at merging this soon, we need to add support for `Properties` as well, currently only `Fields` are supported which was fine for when only supporting `Structs`. Should be trivial to add to the serializer.

joaompneves:
Can't the properties support be made in another pull request and merge this one?

amaitland:
> Can't the properties support be made in another pull request and merge this one?

I'd prefer for this to be complete. What's the advantage of merging this now?

joaompneves:
I  would like to move to cefsharp master other than maintaining my own branch with fixes. But I don't have bandwidth to do those changes now.

perlun:
@joaompneves This one has gone a bit stale. Would you have a chance to update it to resolve the conflicts?

joaompneves:
Not working with recent changes

perlun:
@amaitland Do you think this is superseded by the recent JSB changes you did? I am thinking about 7a0b06ad0b9fdb0470fc02c04cbeea69bd31e7ff.

(Note: haven't looked at the PR myself.)

amaitland:
Support for returning a `Class` from an `async bound method` has been added in https://github.com/cefsharp/CefSharp/commit/c303e1ca0f5996be3130d8d9f7ed0bbd36f877b1

Returning properties and fields for both structs and classes is now possible (`async bound objects only`)

PR #2043, opened 2017-05-11
Update Readme with correct version of VC++ Version

Hammerstad:
Hello.

I was unsure if a small fix like this requires an issue. Please tell me if I have to create one.

The table with cefsharp versions contains an error: CefSharp 37 needs VC++ Redist version 2012, not 2013.
I've verified this on Windows 7 and Windows Server 2012 r2.


amaitland:
No issue required, thanks.

PR #2052, opened 2017-05-17
Fix DefaultBinder conversions from Dictionary<string,object>

joaompneves:
The following example will work and not crash:

binder = new DefaultBinder(new DefaultFieldNameConverter());
obj = new Dictionary<string, object> { { "key", "value" } };
binder.Bind(new [] {  obj }, typeof(Dictionary<string, object>[])); => returns [ obj ]
binder.Bind(obj, typeof(Dictionary<string, object>)); => returns obj

1) Dictionaries were wrongly considered collections.
2) Some conversions involving Dictionaries caused NullReferenceExceptions to be thrown

### DIFF
--- a/CefSharp/ModelBinding/DefaultBinder.cs
+++ b/CefSharp/ModelBinding/DefaultBinder.cs
@@ -72,6 +72,10 @@ namespace CefSharp.ModelBinding
                 {
                     throw new ArgumentException("When modelType is an enumerable it must specify the type.", "modelType");
                 }
+            } 
+            else if (modelType.IsGenericObject())
+            {
+                return obj;
             }
 
             var bindingContext = this.CreateBindingContext(obj, modelType, genericType);
@@ -92,7 +96,7 @@ namespace CefSharp.ModelBinding
 
                     if (val != null)
                     {
-                        if (val.GetType() == typeof(Dictionary<string, object>))
+                        if (val.GetType().IsGenericObject())
                         {
                             var subModel = Bind(val, genericType);
                             model.Add(subModel);
@@ -144,10 +148,8 @@ namespace CefSharp.ModelBinding
                 return;
             }
 
-            Type dictionaryType = typeof(Dictionary<string, object>);
-
             //If the type is a dictionary and the PropertyType isn't then we'll bind.
-            if (obj.GetType() == dictionaryType && modelProperty.PropertyType != dictionaryType)
+            if (obj.GetType().IsGenericObject() && !modelProperty.PropertyType.IsGenericObject())
             {
                 //We have a sub dictionary, attempt to bind it to the class
                 var model = Bind(obj, modelProperty.PropertyType);
@@ -190,7 +192,7 @@ namespace CefSharp.ModelBinding
 
         protected virtual object GetValue(string propertyName, BindingContext context)
         {
-            if (context.Object.GetType() == typeof(Dictionary<string, object>))
+            if (context.Object.GetType().IsGenericObject())
             {
                 var dictionary = (Dictionary<string, object>)context.Object;
                 if (dictionary.ContainsKey(propertyName))
--- a/CefSharp/ModelBinding/ModelBindingExtensions.cs
+++ b/CefSharp/ModelBinding/ModelBindingExtensions.cs
@@ -74,7 +74,8 @@ namespace CefSharp.ModelBinding
         /// </summary>
         /// <param name="source">source type</param>
         /// <returns>return true if is a generic js object</returns>
-        public static bool IsGenericObject(this Type source) {
+        public static bool IsGenericObject(this Type source) 
+        {
             return source == typeof(Dictionary<string, object>);
         }
 
--- a/CefSharp/ModelBinding/ModelBindingExtensions.cs
+++ b/CefSharp/ModelBinding/ModelBindingExtensions.cs
@@ -56,7 +56,6 @@ namespace CefSharp.ModelBinding
         /// <returns><see langword="true" /> if the type is an array, otherwise <see langword="false" />.</returns>
         public static bool IsArray(this Type source)
         {
-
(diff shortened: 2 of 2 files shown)


amaitland:
What's the go here? Is there actually a bug?

joaompneves:
There is but I have to take a deep look at it, that's why I closed it for now. I'll re-open later.

CefSharp/ModelBinding/DefaultBinder.cs
```diff
@@ -72,6 +72,10 @@ public virtual object Bind(object obj, Type modelType)
                 {
                     throw new ArgumentException("When modelType is an enumerable it must specify the type.", "modelType");
                 }
+            } 
+            else if (modelType.IsGenericObject())
```
amaitland: Type checking here seems a little weak, I think we'd be better off immediately after the `obj == null` check to using something like 

```c#
//If the object can be directly assigned to the modelType then return immediately. 
if(modelType.IsAssignableFrom(obj.GetType()))
{
	return obj;
}
```

CefSharp/ModelBinding/ModelBindingExtensions.cs
```diff
@@ -67,7 +66,17 @@ public static bool IsArray(this Type source)
...
-            return source.IsCollection() || source.IsArray() || source.IsEnumerable();
+            return !source.IsGenericObject() && (source.IsCollection() || source.IsArray() || source.IsEnumerable());
+        }
+
+        /// <summary>
+        /// The type is a generic javascript object (dictionary)
+        /// </summary>
+        /// <param name="source">source type</param>
+        /// <returns>return true if is a generic js object</returns>
+        public static bool IsGenericObject(this Type source)
```
amaitland: The name of this method is confusing, I look at it and think of `.Net generics`, I don't actually think it's required either if we make a few other simplifications.

CefSharp/ModelBinding/ModelBindingExtensions.cs
```diff
@@ -94,7 +103,7 @@ public static bool IsAssignableToGenericType(this Type givenType, Type genericTy
         /// </summary>
         /// <param name="source">The type to check.</param>
         /// <returns><see langword="true" /> if the type is an collection, otherwise <see langword="false" />.</returns>
-        public static bool IsCollection(this Type source)
+        private static bool IsCollection(this Type source)
```
amaitland: Reason for making this private is?

CefSharp/ModelBinding/DefaultBinder.cs
```diff
@@ -144,10 +148,8 @@ protected virtual void BindValue(BindingMemberInfo modelProperty, object obj, Bi
                 return;
             }
 
-            Type dictionaryType = typeof(Dictionary<string, object>);
-
             //If the type is a dictionary and the PropertyType isn't then we'll bind.
-            if (obj.GetType() == dictionaryType && modelProperty.PropertyType != dictionaryType)
+            if (obj.GetType().IsGenericObject() && !modelProperty.PropertyType.IsGenericObject())
```
amaitland: As the two `if` branches here are duplicated I think we can simplify the whole `BindValue` method. We should be able to simplify things quite a bit, using something like the following.

```c#
protected virtual void BindValue(BindingMemberInfo modelProperty, object obj, BindingContext context)
{
	if(obj == null)
	{
		return;
	}

	if (modelProperty.PropertyType.IsAssignableFrom(obj.GetType()))
	{
		//Simply set the property
		modelProperty.SetValue(context.Model, obj);
	}
	else
	{
		//Cannot directly set the property attempt to bind
		var model = Bind(obj, modelProperty.PropertyType);

		modelProperty.SetValue(context.Model, model);
	}
}
```

CefSharp/ModelBinding/DefaultBinder.cs
```diff
@@ -190,7 +192,7 @@ protected virtual object CreateModel(Type modelType, Type genericType)
 
         protected virtual object GetValue(string propertyName, BindingContext context)
         {
-            if (context.Object.GetType() == typeof(Dictionary<string, object>))
+            if (context.Object.GetType().IsGenericObject())
```
amaitland: Please revert this and remove `IsGenericObject` extension method.

CefSharp/ModelBinding/ModelBindingExtensions.cs
```diff
@@ -67,7 +66,17 @@ public static bool IsArray(this Type source)
         /// <returns>return if collection, array or enumerable</returns>
         public static bool IsCollectionOrArray(this Type source)
         {
-            return source.IsCollection() || source.IsArray() || source.IsEnumerable();
+            return !source.IsGenericObject() && (source.IsCollection() || source.IsArray() || source.IsEnumerable());
```
amaitland: I'm hopeful we can revert this change with the other changes I've proposed.

CefSharp/ModelBinding/DefaultBinder.cs
```diff
@@ -92,7 +96,7 @@ public virtual object Bind(object obj, Type modelType)
 
                     if (val != null)
                     {
-                        if (val.GetType() == typeof(Dictionary<string, object>))
+                        if (val.GetType().IsGenericObject())
```
amaitland: Please revert this and remove `IsGenericObject` extension method.

amaitland:
I've merged my proposed changes in commit https://github.com/cefsharp/CefSharp/commit/0d48e30418d1ac940f18098bbd7d9574a188951e

Code is greatly simplified. If your still experiencing a problem I believe the simplifications are a better starting point for any further bug fixes. Where possible please include a test case that reproduces your problem, would make my life a lot easier, thank you.

joaompneves:
It works as expected, great!
Where should I add the test case? 
In the PR description I added a snippet of code (a test case) that was crashing before.

PR #2056, opened 2017-05-29
Prevent accelerator keys to work when typing text

joaompneves:
When you have buttons with accelerator keys defined in a window with CefSharp and you press the key of the accelerator while the web view is focused, the key won't be sent to the web view.
Expected the web view to receive the key event when focused.
This pull request reestablishes that behavior.

Ex: If you have a button _Go (the accelerator key is G) and you type G inside the web view, the web view won't get the event.


amaitland:
Please provide an example that can be used for testing.

amaitland:
Without looking into this in detail, from a maintenance point of view I'm not keen on adding a third piece of keyboard handling logic, having two is one more than I'd like as it is.

amaitland:
I've got no problem with removing the `SourceHook` and replacing with a higher level `WPF` implementation.

joaompneves:
Removed SourceHook and ended up with 2 key handling methods, one for key down/up to handle non-textual key presses and another for textual key presses.

CefSharp.Wpf/ChromiumWebBrowser.cs
```diff
@@ -1972,25 +1898,43 @@ protected override void OnPreviewKeyUp(KeyEventArgs e)
...
+                var browserHost = browser.GetHost();
+                for (int i = 0; i < e.Text.Length; i++) 
                 {
-                    browser.GetHost().SendKeyEvent(message, virtualKey, (int)modifiers);
-                    e.Handled = true;
+                    browserHost.SendKeyEvent((int)WM.CHAR, e.Text[i], 0);
                 }
+                e.Handled = true;
             }
+            base.OnTextInput(e);
```
amaitland: Should `base.OnTextInput(e);` be `base.OnPreviewTextInput(e);`?

CefSharp.Wpf/ChromiumWebBrowser.cs
```diff
@@ -1972,25 +1898,43 @@ protected override void OnPreviewKeyUp(KeyEventArgs e)
...
+        }
 
-                if(browser != null)
+        /// <summary>
+        /// Handles the <see cref="E:PreviewTextInput" /> event.
+        /// </summary>
+        /// <param name="e">The <see cref="TextCompositionEventArgs"/> instance containing the event data.</param>
+        protected override void OnPreviewTextInput(TextCompositionEventArgs e) 
+        {
+            if (browser != null)
```
amaitland: We need to check if (!e.Handled)` here, handle the case where an overriding implementation chooses to handle the input.

joaompneves:
This is now again ready to merge.

joaompneves:
Any plans to merge this PR?

perlun:
@amaitland Since you looked at this before, maybe it makes sense that you look at the latest changes as well. (If you don't have time, let me know as usual and we'll try to make it work in other ways.)

amaitland:
Firstly it conflicts with the changes proposed in #2103, whilst I like the idea of removing the source hook and having a higher level solution it's unclear if a `sourcehook` is required to implement `IME` support. I would hope there are higher level `APIs` that we can tap into, I just don't have any experience to know one way or the other.

Secondly there's the question of testing, I'm assuming you've deployed this to a sizeable user base? Testing keyboard input is difficult, so many combinations to factor in. If this `PR` can be passed as production ready then I'm happy to see it merged.

joaompneves:
This change has been deployed and its running for 6 months now. 
I didn't make any automatic tests cause its a bit hard to simulate those scenarios in an automatic unit testing environment.

Regarding the IME support, I would like to have support for it, although I think its more important to have the standard keyboard flow handling working correctly and then have IME support.

amaitland:
> This change has been deployed and its running for 6 months now.

How many users? Are they all using the same `locale` or different?

Whilst I'm the first to admit the current keyboard implementation has a few problems, it's also pretty stable, it''s rare that someone raises an issue, so that adds to my reluctance in merging.

@perlun Thoughts?

joaompneves:
Around 200 users, as the current software version using this CefSharp version is still in beta stage. In a few weeks/months this user base will grow to thousands as the software reaches other release stages. Currently we have users testing in portuguese and english locales (at least).

joaompneves:
Any plans for this?

amaitland:
If you can make the behavior configurable then the risk of merging is trivial and I'm happy to include this. You can default to the new behavior, long as the old is there as a fall back.

joaompneves:
Added the ability to fallback to the old behavior using a property. Please review.

Tested both approaches against: http://en.key-test.ru/

amaitland:
Thanks 👍 Assuming this passes basic testing I'll include in the `63` milestone.

amaitland:
Merged with changes see https://github.com/cefsharp/CefSharp/commit/a2451960d38c395a2a3145fb7244c2e101c19627

Anyone can implement their own `IWpfKeyboardHandler`, so in the case of #2103 it can easily be integrated implemented in user code.

PR #2067, opened 2017-06-12
Prevent VS from always rebuilding projects that reference CefSharp when using AnyCPU configuration

joaompneves:
Fixed the problem of Visual Studio Ignoring the Condition on the Reference tag [https://stackoverflow.com/questions/28851338/why-the-condition-attribute-doesnt-work-for-the-itemgroup-element]. This will stop VS from always rebuilding projects that reference CefSharp when using AnyCPU configuration

### DIFF
--- a/NuGet/CefSharp.Common.props
+++ b/NuGet/CefSharp.Common.props
@@ -1,29 +1,41 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <ItemGroup>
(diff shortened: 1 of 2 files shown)


merceyz:
Should apply this to all the projects, AnyCPU should probably be it's own folder.

If you build with AnyCPU with your current changes and run it seems like it will crash depending on what CEF files was last copied
https://github.com/cefsharp/CefSharp/issues/1714

joaompneves:
Fixed remaining projects. Not getting the behavior you described. This is just a workaround for a VS known compilation bug and the run-time behavior should be unaffected.

merceyz:
Merged the complete version from https://github.com/cefsharp/CefSharp/pull/2083

PR #2071, opened 2017-06-19
change labels of badges

304NotModified:
Personally I find 3 the same badges confusing. So I changed the labels: 

before:
image

after:
image


merceyz:
Cheers

304NotModified:
@merceyz 

Would it also be OK to replace the links with the badges? (under stable and pre-release)

image

IMO It's conventional to press on the badge there, but currently that will show a new page with only the badge:
image

merceyz:
Yeah, that should be fine

PR #2072, opened 2017-06-19
Badges for all releases

304NotModified:
Badges for all release links.

I also changed the label "build status", as this section isn't describing the build status?

Maybe the layout is a bit poor now. Please your opinion.

image

[full preview](https://github.com/304NotModified/CefSharp/blob/patch-2/README.md)

If it's OK, I could do a git squash first


perlun:
> If it's OK, I could do a git squash first

Looks fine! I think I'll just squash it here in the web UI. Thanks for contributing!

Issue #2077, opened 2017-06-28
Visual Studio keeps rebuilding projects that reference CefSharp NuGet packages and build AnyCPU platform


ymccready:
- **What version of the product are you using?**
57.0.0.0, via Nuget in Visual Studio 2015

- **What architecture x86 or x64?**
AnyCPU

- **On what operating system?**
Win 8.1

- **Are you using `WinForms`, `WPF` or `OffScreen`?**
WPF and OffScreen, but probably applied to WinForms

- **What steps will reproduce the problem?**
  - Add Nuget Package for OffScreen (for example) to a new project in VS2015.
  - Add `<CefSharpAnyCpuSupport>true</CefSharpAnyCpuSupport>` as detailed here: https://github.com/cefsharp/CefSharp/issues/1714
  - Build the project: Project builds with success. 
  - Build the project again: Project builds again.

- **What is the expected output? What do you see instead?**
The project not rebuild again and be "up-to-date" according to VS. 

- **Please provide any additional information below.**
If you set the VS build verbosity to _Diagnostic_, you will get the following message on the first line of the log when re-building the project:
  >1>Project 'CefSharpTest' is not up to date. CopyLocal reference ...\CefSharpTest\bin\Debug\CefSharp.OffScreen.dll' is missing from output location.

  *Visual Studio is expecting the dll to be in the output because it did not properly evaluate the references with conditions in the CefSharp.OffScreen.props file*

- Workaround
I hard coded the references to the dlls in my project and commented the import directives for the _.props_ files. This will probably have to be redone if we upgrade Package versions.

ymccready:
This was also reported in PR #2067 with a viable but incomplete solution.

Kingtwo:
@ymccready But I builded with x86 or x64，And somebody can run，somebody not。I looked up some information，that said：“Cefsharp needs to install VC++ 2013 Redist on the user's computer”。
So, what's wrong? thank you

merceyz:
Fixed with https://github.com/cefsharp/CefSharp/commit/2bdaa17d6474d900ef7bd7e235cdb0f37b359b2e

PR #2083, opened 2017-07-06
Prevent VS from always rebuilding projects that reference CefSharp when using AnyCPU platform, All .props files

ymccready:
Realized I should create pr on master. Sorry about that.

I'm also having the same issue as @joaompneves with Visual Studio rebuilding every time projects that depend on CefSharp and use AnyCPU platform are built. I opened Issue #2077 to add more details and repro steps.

I took @joaompneves' PR #2067 that fixes this issue and applied it to all the other .props files so that the issue is fixed for all references (common, core, Wpf, WinForms, OffScreen).

I hope this helps,


merceyz:
Thanks

Issue #1262, opened 2015-09-10
WPF IME Windows 8.1/10


ay2015:
I use the project from official  website and version '41.0.1' ,when I design a wonderful window,I have to set variable ''AllowsTransparency" to "true", “WindowStyle” to "None",I edited the ‘MainWindow.xaml’ ,set variable ''AllowsTransparency" to "true", “WindowStyle” to "None",and run.
the problem:  input tag from html can't use other IME,but character is Ok，how can I solve it?  
Thanks for you

amaitland:
> the problem: input tag from html can't use other IME,but character is Ok，

What is other `IME`? What is `character`? Can you describe your problem in more detail? Some screenshots maybe?

ay2015:
IME：input method such as Sogou Input software, When I wanna input Chinese Character,However, the effect is as follows:
error1: http://www.wpfui.com/share/1111.jpg

amaitland:
> I edited the ‘MainWindow.xaml’ ,set variable ''AllowsTransparency" to "true", “WindowStyle” to "None",

Are you using the `41` branch specifically for your testing?
https://github.com/cefsharp/CefSharp/tree/cefsharp/41

amaitland:
Your post references `AllowsTransparency` a lot, is that actually relevant to the problem? Does the problem reproduce straight out of the box without editing the example your using?

ay2015:
I don't edit straight  your code, I only download the source code, then unzip it. 
After Executed the 'Build.bat' file in the Root Directory of project,I open the code by Visual Studio 2015,
Build the solution，Afterwards,I change the website's address to one address which contains input text tag of html,and clicked the  'go' button.
And then input method can't input other characters such as Chinese.

amaitland:
> I don't edit straight your code, I only download the source code,

From which branch?

ay2015:
Branch: master
I download from this address    "https://codeload.github.com/cefsharp/CefSharp/zip/master"

amaitland:
Try with https://github.com/cefsharp/CefSharp/tree/cefsharp/41

Their is an `IME` bug for which `41` has a workaround. It's supposed to be solved in newer versions, though it may be a problem in your case.

https://github.com/cefsharp/CefSharp/commit/7eafe6211c5d152915df28224b1925e4d2c01d02

ay2015:
Dear Author:
         I try your demo of wpf on different operating systems,the result is as follows:
         Win7: IME problem is not exits,
[...]

PR #2103, opened 2017-07-31
fixed WPF osr ime support

starts2000:
fixed @ #1262


merceyz:
Please write a more detailed description so it's easy to see what is done in this PR.

You seem to have changed things unrelated to IME support, please revert that.
Spaces not tabs (1 tab -> 4 spaces)
Use the same formatting style as the rest of the project

ashu3011:
Tried this fix on CefSharp.WPF.Example. There are 2 issue I encountered:
**Issue1:**
To recreate:
1.  Run CefSharp.WPF.Example. 
2. Keyboard Language should be set to any IME supporting language,
3. Type something in browser IME works fine,
4. Change tab
5. Type again

Result: IME doen't get displayed and you are typing in English

**Issue 2**

Null reference exception in chromiumwebbrowser.cs  if the browser window is not displayed yet. The presentationsource returns null which causes the exception when you try to get handle while setting up _imewin

ashu3011:
For Issue 1: I don't get nu;ll reference exception any more but one of my browser is not properly initialized and thus I can not type in it.

Issue 2 still persists. In CefSharp.wpf example just open a  new tab which will default to Google.com. Type a search query, IME will not appear.

ashu3011:
Works perfectly with Cef.WPF.Example . I am still having trouble with my xwt application but I think it has to do something with my implementation of it. The issue is that IME does not show up on initial load of the browser but works perfectly after I get away from it and come back to it. It persists to work rest of the time. I will look deeper and see if I can figure it out. Thanks.

ashu3011:
IME for Japanese language is not at the correct position, it appears at the bottom right of the screen. Also the other IME do not correct position with Scale (Display settings on windows). Thus with few settings, depending on the monitors resolution, they appear on top of the input where IME is used and thus make it tough for users.

CefSharp.BrowserSubprocess/CefSharp.BrowserSubprocess.csproj
```diff
@@ -111,13 +111,8 @@
...
   <PropertyGroup>
-    <PostBuildEvent>if $(PlatformName) == x86 (
-  "$(DevEnvDir)..\..\VC\bin\editbin" /largeaddressaware /TSAWARE "$(TargetPath)"
-  call "$(DevEnvDir)..\Tools\vsvars32.bat"
-  sn -R "$(TargetPath)" "$(ProjectDir)..\CefSharp.snk"
-) else (
-  "$(DevEnvDir)..\..\VC\bin\amd64\editbin" /TSAWARE "$(TargetPath)"
-)</PostBuildEvent>
+    <PostBuildEvent>
+    </PostBuildEvent>
```
perlun: This part should be reverted, since it disables strong named signing.

CefSharp.Core/Internals/CefBrowserHostWrapper.cpp
```diff
@@ -1,4 +1,4 @@
-// Copyright � 2010-2017 The CefSharp Authors. All rights reserved.
+// Copyright ?2010-2017 The CefSharp Authors. All rights reserved.
```
perlun: This and other similar changes should be reverted - the copyright sign should remain in the previous encoding. (UTF-8 I presume)

amaitland:
This `PR` conflicts with the changes proposed in #2056

amaitland:
#2056 has been merged. With those changes you should be able to implement your own `IWpfKeyboardHandler` and create a solution that requires no changes to the core `CefSharp` code. (You'll likely need to create a class that inherits from `ChromiumWebBrowser`).

Looking at the code it can be **greatly** simplified, there's lots of code there that has no relevance to `CefSharp`

CefSharp.Core/Internals/CefBrowserHostWrapper.cpp
```diff
@@ -336,12 +336,30 @@ void CefBrowserHostWrapper::ImeSetComposition(String^ text, cli::array<Compositi
     _browserHost->ImeSetComposition(StringUtils::ToNative(text), underlinesVector, CefRange(), range);
 }
 
+void CefBrowserHostWrapper::ImeSetComposition(
```
amaitland: These can likely be removed and you just get a reference to the underlying `CefBrowserHost` rather than adding code.

It's my understanding that some of the params are only used on `Mac` and can be left out of any `Windows` implementation.

CefSharp.Wpf/ChromiumWebBrowser.cs
```diff
@@ -471,7 +496,7 @@ private void NoInliningConstructor()
 
             PresentationSource.AddSourceChangedHandler(this, PresentationSourceChangedHandler);
 
-            RenderOptions.SetBitmapScalingMode(this, BitmapScalingMode.HighQuality);
+            RenderOptions.SetBitmapScalingMode(this, BitmapScalingMode.NearestNeighbor);
```
amaitland: Revert this, you can change it in your own code if you wish to adjust the quality.

CefSharp.Wpf/ChromiumWebBrowser.cs
```diff
@@ -828,6 +859,24 @@ void IRenderWebBrowser.OnImeCompositionRangeChanged(Range selectedRange, Rect[]
...
+                UiThreadRunAsync(() =>
+                {
+                    IList<Rect> rects = new List<Rect>();
+                    foreach (var item in characterBounds)
+                    {
+                        GeneralTransform generalTransform1 = TransformToAncestor(this);
+
+                        if (CleanupElement != null)
+                        {
+                            Point point = this.TransformToAncestor(CleanupElement).Transform(new Point(0, 0));
```
amaitland: Using `CleanupElement` here really isn't a reliably method to access the parent `Window`, which is what I'm assuming your doing. I'm sure you can find a better way of doing this.

CefSharp.Wpf/ChromiumWebBrowser.cs
```diff
@@ -988,6 +1037,17 @@ void IWebBrowserInternal.OnAfterBrowserCreated(IBrowser browser)
...
                 if (!IsDisposed)
                 {
+                    if (_imeWin != null)
+                    {
+                        _imeWin.Dispose();
+                    }
+
+                    if (source != null)
+                    {
+                        _imeWin = new OsrImeWin(source.Handle, this.browser);
```
amaitland: If you create `OsrImeWin` when focus is gained and destroyed it when focus Is lost, is there a reason it's created here?

CefSharp.Wpf/ChromiumWebBrowser.cs
```diff
@@ -1859,37 +1954,15 @@ protected virtual IntPtr SourceHook(IntPtr hWnd, int message, IntPtr wParam, Int
                 return IntPtr.Zero;
             }
 
-            switch ((WM)message)
+            if (!IsDisposed && _imeWin != null && IsKeyboardFocused && browser != null)
```
amaitland: You can now implement your own `IWpfKeyboardHandler` to capture and forward key events.

CefSharp.Core/OsrImeHandler.h
```diff
@@ -0,0 +1,117 @@
...
+#include <windows.h>
+//#include <vector>
+
+//#include "include/internal/cef_types_wrappers.h"
+
+namespace CefSharp {
+
+	// Handles IME for the native parent window that hosts an off-screen browser.
+	// This object is only accessed on the CEF UI thread.
+	class OsrImeHandler {
```
amaitland: Unless there is a specific reason this is a separate class I think this should be merged into `OsrImeWin`

CefSharp.Core/OsrImeHandler.h
```diff
@@ -0,0 +1,117 @@
+// Copyright 2016 The Chromium Embedded Framework Authors. Portions copyright
+// 2013 The Chromium Authors. All rights reserved. Use of this source code is
+// governed by a BSD-style license that can be found in the LICENSE file.
+
+#ifndef CEF_TESTS_CEFCLIENT_BROWSER_OSR_IME_HANDLER_WIN_H_
+#define CEF_TESTS_CEFCLIENT_BROWSER_OSR_IME_HANDLER_WIN_H_
```
amaitland: Remove this.

amaitland:
Closing as I have no plans to merge this.

See https://github.com/cefsharp/CefSharp/issues/1262#issuecomment-445088428 for an alternative.

Issue #2106, opened 2017-08-11
Fatal Exception if ChromiumWebBrowser in loaded but not visible


jamespearce2006:
In version 57 (WPF), if a ChromiumWebBrowser is loaded, but is not yet part of the visual tree, an unhandled exception is thrown:

=======================
```
8/11/2017 9:46:27 AM
Exception Details:  This Visual is not connected to a PresentationSource. (System.InvalidOperationException)

   at System.Windows.Media.Visual.PointToScreen(Point point)
   at CefSharp.Wpf.ChromiumWebBrowser.OnLoaded(Object sender, RoutedEventArgs routedEventArgs)
   at System.Windows.EventRoute.InvokeHandlersImpl(Object source, RoutedEventArgs args, Boolean reRaised)
   at System.Windows.UIElement.RaiseEventImpl(DependencyObject sender, RoutedEventArgs args)
   at System.Windows.BroadcastEventHelper.BroadcastEvent(DependencyObject root, RoutedEvent routedEvent)
   at System.Windows.BroadcastEventHelper.BroadcastLoadedEvent(Object root)
   at MS.Internal.LoadedOrUnloadedOperation.DoWork()
   at System.Windows.Media.MediaContext.FireLoadedPendingCallbacks()
   at System.Windows.Media.MediaContext.FireInvokeOnRenderCallbacks()
   at System.Windows.Media.MediaContext.RenderMessageHandlerCore(Object resizedCompositionTarget)
   at System.Windows.Media.MediaContext.RenderMessageHandler(Object resizedCompositionTarget)
   at System.Windows.Interop.HwndTarget.OnResize()
   at System.Windows.Interop.HwndTarget.HandleMessage(WindowMessage msg, IntPtr wparam, IntPtr lparam)
   at System.Windows.Interop.HwndSource.HwndTargetFilterMessage(IntPtr hwnd, Int32 msg, IntPtr wParam, IntPtr lParam, Boolean& handled)
   at MS.Win32.HwndWrapper.WndProc(IntPtr hwnd, Int32 msg, IntPtr wParam, IntPtr lParam, Boolean& handled)
   at MS.Win32.HwndSubclass.DispatcherCallbackOperation(Object o)
   at System.Windows.Threading.ExceptionWrapper.InternalRealCall(Delegate callback, Object args, Int32 numArgs)
[...]

PR #2107, opened 2017-08-11
Prevent unhandled exception in PointToScreen when not yet visible

jamespearce2006:
Fix for #2106.
Checks that the visual is connected to a presentation source before calling PointToScreen()
Updates the browserScreenLocation value when the presentation source changes
Fix compiler warning for XML comment

### DIFF
--- a/CefSharp.Wpf/ChromiumWebBrowser.cs
+++ b/CefSharp.Wpf/ChromiumWebBrowser.cs
@@ -83,7 +83,6 @@ namespace CefSharp.Wpf
         /// </summary>
         private int disposeCount;
         /// <summary>
-        /// <summary>
         /// Location of the control on the screen, relative to Top/Left
         /// Used to calculate GetScreenPoint
         /// We're unable to call PointToScreen directly due to treading restrictions
@@ -1547,6 +1546,8 @@ namespace CefSharp.Wpf
                         window.StateChanged += WindowStateChanged;
                         window.LocationChanged += OnWindowLocationChanged;
                     }
+
+                    updateBrowserSecreenLocation();
                 }
             }
             else if (args.OldSource != null)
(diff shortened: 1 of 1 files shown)


CefSharp.Wpf/ChromiumWebBrowser.cs
```diff
@@ -1588,11 +1589,19 @@ private void WindowStateChanged(object sender, EventArgs e)
             } 
         }
 
+        private void updateBrowserSecreenLocation()
```
merceyz: Please correct the spelling of the function

jamespearce2006:
Done

CefSharp.Wpf/ChromiumWebBrowser.cs
```diff
@@ -1743,7 +1752,7 @@ private void OnLoaded(object sender, RoutedEventArgs routedEventArgs)
             tooltipTimer.IsEnabled = false;
 
             //Initial value for screen location
-            browserScreenLocation = PointToScreen(new Point());
+            updateBrowserScreenLocation();
```
merceyz: `browserScreenLocation` would need a default value here. Perhaps make a call to `updateBrowserScreenLocation` once the control is part of a `visual tree` (assuming there is an event for that)

merceyz:
Thank you

CefSharp.Wpf/ChromiumWebBrowser.cs
```diff
@@ -1588,11 +1589,19 @@ private void WindowStateChanged(object sender, EventArgs e)
             } 
         }
 
+        private void updateBrowserScreenLocation()
```
amaitland: Should start with a capital.

CefSharp.Wpf/ChromiumWebBrowser.cs
```diff
@@ -1588,11 +1589,19 @@ private void WindowStateChanged(object sender, EventArgs e)
...
+        {
+            if (PresentationSource.FromVisual(this) != null)
```
amaitland: This is a very expensive check to make (have a look at the `.Net` reference source if you want to know more, it's a few levels deep in the functions calls when it starts walking up the tree from memory).

`PresentationSourceChangedHandler` already tracks additions/removal if the `HwndSource`, you can probably check if `source != null` instead.

Issue #1983, opened 2017-03-10
Feature Request - Upgrade to VS2015(VC++ 2015)


amaitland:
Historically we've only supported two version of `Visual Studio` at any one time, it's currently `VS2013` and `VS2015` with default builds generated using `VS2013`, which ties us to `VC++ 2013`. With the release of `VS2017` just recently it's time to consider our roadmap and when we move from `VS2013` as our default to `VS2015`. 

The only complication with this is the `VC++` version being tired to the `VS` version. So `VS2015` means the new requirement for `VC++2015`.

I'm not in a quite rush to make this change as `VC2017` is very new, would prefer that it stabilise for a bit, so this is something to consider over the next `3-6mths`.

The `cef.sdk` packages would be updated so by default it would only be possible to build using `VS2015/17` as they get too large including three `debug` and `release` builds of the `libcefwrapper`

Feedback welcome.

mountgellert:
+1

peters:
@amaitland I agree with your reasoning. Would you mind postponing the transition to `VS2015` post CefSharp 57 release? A couple of weeks ago we had to manually update the .NET framework and c++ runtime for all our payment kiosks. Even with TeamViewer that was large effort 😊

amaitland:
> Would you mind postponing the transition to VS2015 post CefSharp 57 release?

@peters I was thinking more like version `59` or `61`. No plans to update before `57` 😄

peters:
@amaitland Awesome 🍻

amaitland:
Just for clarity, at this stage we'll look at upgrading to `VS2015` after `59` has been branched, so the first release will be `61`.

amaitland:
Just as a note, when adding support for `VS2017` `editbin` is in a different location, See https://gitter.im/cefsharp/CefSharp?at=592af5b00a783b6c0ae28259 for details.

amaitland:
> VC Runtime in MSVC 2017 is binary compatible with 2015

https://blogs.msdn.microsoft.com/vcblog/2017/03/07/binary-compatibility-and-pain-free-upgrade-why-moving-to-visual-studio-2017-is-almost-too-easy/

Looks like `VS2017` brings a lot of advancements, including the ability to target `VC++2015.3`

DickvdBrink:
https://groups.google.com/a/chromium.org/d/msg/chromium-dev/Y3OEIKkdlu0/TCcT1SvwAwAJ
[...]

Issue #1782, opened 2016-08-25
System.NullReferenceException Accessing Registered C# Object


JWallaceDev:
We have a client experiencing a System.NullReferenceException:

![exception](https://cloud.githubusercontent.com/assets/14850075/17973971/3804407c-6adc-11e6-9a29-44744739a78a.jpg)

This is isolated to 1 machine and isn't repeatable on any other machine at their company, or on any of our dev machines.

We are getting a bit stuck for ideas on possible causes and looking for some ideas to at least replicate the issue.

Based on the exception stack have you any suggestions?

It looks like the exception is occurring the first ever time the JS attempts to access a C# object registered via RegisterJsObject.

Currently suspect the way that the C# classes are registered and accessed.

A developer has created a demo app - which exhibits the same problem for the client.

This file contains the C# class being registered and the code to register it:

http://pastebin.com/s7HWZBBQ

And this contains the JavaScript accessing the C#:

http://pastebin.com/7LF54fFy

- **What version of the product are you using?**
   Build of branch 51.0.0

- **What architecture x86 or x64?**
  x64

- **On what operating system?**
     Win7

- **Are you using `WinForms`, `WPF` or `OffScreen`?**
WinForms

- **What steps will reproduce the problem?**

We are currently unable to reproduce the problem outside of a remote clients machine.

When they launch a demo app containing the code pasted above, the null reference exception occurs when trying to access the registered C# object. The exception can be seen in Dev Tools.

- **What is the expected output? What do you see instead?**

No exception and the embedded browser loads correctly.

- **Please provide any additional information below.**
    - A stack trace if available, any Exception information.

    - Does the cef log provide any relevant information? 
No - with verbose logging on we get very similar/almost identical results to those on a machine where everything works.

I've attached the log anyway:

    - Any other background information that's relevant? Are you doing something out of the ordinary? 3rd party controls?

- Client has disabled/removed all anti-virus
- Client has successfully used the application on a different machine in their organisation.

- **Does this problem also occur in the `Cef` TestApp:

No it doesn't.

rlmcneary2:
@JWallaceDev - issues are for problem reports and change requests only; please ask questions on Gitter. Suggest you register your objects **before** you add the CWB to your form.

amaitland:
> Based on the exception stack have you any suggestions?

@JWallaceDev Have you tried opening the image you posted? It's too small for me to read, probably others as well. Ideally just post a plain text version.

[...]

Issue #1203, opened 2015-08-17
Javascript Bound objects no longer registered after cross-site navigation


MIvoylov:
When you call the method named "LoadUrl" losted all registered JsObject.

What I do:
1. Create a browser: new ChromiumWebBrowser (null).
2. Registering JsObject .: browser.RegisterAsyncJsObject
3. Call the method LoadUrl. Registered object "JsObject" available on the page.
4. The method LoadUrl is called again with a different address. On the loaded page registered object "JsObject" is not available

How can re-register JsObject or prevent registserd objects clean?

amaitland:
What version are you using? Head of `master`?

amaitland:
@cefsharp-ms @arsher Any thoughts? I moved the register code into `OnRenderViewReady` see https://github.com/cefsharp/CefSharp/commit/0e2078e288fbde041812d49f874ea773e6302da1#diff-2731f23ba4e8fcf96f8501a49c6d687eR512

> Called on the browser process UI thread when the render view associated
  // with |browser| is ready to receive/handle IPC messages in the render

I reverted this and it made no difference. Looking into it further I'm not surprised. As the change of domain creates a new `render process`. I haven't come across a suitable event handler, open to suggestions. I tried quickly changing to send an `IPC` message when the process was created, still ended up with what looked like a timing issue between when the context was created and the `IPC` message was received.

I'll look at this in more detail later, open to suggestions if someone comes up with a solution.

cefsharp-ms:
I was giving this a thought and some research but it seems like CEF doesn't have any events to notify us when a browser instances switches render processes :disappointed: . Currently the only solution I can think of is to keep a global list of bound objects not per browser and send each one of them to every new process just like it's done with the extensions and schemes. What do you think?

amaitland:
> Currently the only solution I can think of is to keep a global list of bound objects not per browser and send each one of them to every new process just like it's done with the extensions and schemes. What do you think?

Maybe as a last resort. I'm not really a fan of that approach. I might ask on `ceforum` see if anyone else has some advice.

cefsharp-ms:
[...]

PR #2114, opened 2017-08-18
Update to Cef Version 59.0.0

GrabzIt:
Note that there is one breaking change as offscreen windows are now automatically transparent due to this cef change: https://bitbucket.org/chromiumembedded/cef/commits/3f71138d64586d774f645a6db5283b5205ea9df4

To make a offscreen window non-transparent you need to set the background colour as shown below were it is set to white:

browserSettings.BackgroundColor = 0255255255;

### DIFF
--- a/CefSharp.BrowserSubprocess.Core/CefSharp.BrowserSubprocess.Core.vcxproj
+++ b/CefSharp.BrowserSubprocess.Core/CefSharp.BrowserSubprocess.Core.vcxproj
@@ -226,10 +226,4 @@
   </ItemGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
   <ImportGroup Label="ExtensionTargets" />
-  <Target Name="EnsureNuGetPackageBuildImports" BeforeTargets="PrepareForBuild">
-    <PropertyGroup>
-      <ErrorText>This project references NuGet package(s) that are missing on this computer. Use NuGet Package Restore to download them.  For more information, see http://go.microsoft.com/fwlink/?LinkID=322105. The missing file is {0}.</ErrorText>
-    </PropertyGroup>
-    <Error Condition="!Exists('..\packages\cef.sdk.3.3071.1649\build\cef.sdk.props')" Text="$([System.String]::Format('$(ErrorText)', '..\packages\cef.sdk.3.3071.1649\build\cef.sdk.props'))" />
-  </Target>
 </Project>
--- a/CefSharp.BrowserSubprocess.Core/CefSharp.BrowserSubprocess.Core.vcxproj
+++ b/CefSharp.BrowserSubprocess.Core/CefSharp.BrowserSubprocess.Core.vcxproj
@@ -1,6 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project DefaultTargets="Build" ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <Import Project="..\packages\cef.sdk.3.3029.1611\build\cef.sdk.props" Condition="Exists('..\packages\cef.sdk.3.3029.1611\build\cef.sdk.props')" />
+  <Import Project="..\packages\cef.sdk.3.3071.1634\build\cef.sdk.props" Condition="Exists('..\packages\cef.sdk.3.3071.1634\build\cef.sdk.props')" />
   <ItemGroup Label="ProjectConfigurations">
     <ProjectConfiguration Include="Debug|Win32">
       <Configuration>Debug</Configuration>
@@ -66,9 +66,7 @@
   <ImportGroup Condition="'$(Configuration)|$(Platform)'=='Release|x64'" Label="PropertySheets">
     <Import Project="$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props" Condition="exists('$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props')" Label="LocalAppDataPlatform" />
   </ImportGroup>
-  <PropertyGroup Label="UserMacros">
-    <NuGetPackageImportStamp>0c266be5</NuGetPackageImportStamp>
-  </PropertyGroup>
+  <PropertyGroup Label="UserMacros" />
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">
     <LinkIncremental>true</LinkIncremental>
     <LinkKeyFile>..\CefSharp.snk</LinkKeyFile>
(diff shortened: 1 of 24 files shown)


GrabzIt:
This will require the cef version in appveyor to be updated to 3.3071.1649

sireneweb:
any news about this update ?

CefSharp.BrowserSubprocess.Core/CefSharp.BrowserSubprocess.Core.vcxproj
```diff
@@ -221,11 +219,17 @@
...
-    <None Include="packages.config" />
+    <ResourceCompile Include="Resource.rc" />
   </ItemGroup>
   <ItemGroup>
-    <ResourceCompile Include="Resource.rc" />
+    <None Include="packages.config" />
   </ItemGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
   <ImportGroup Label="ExtensionTargets" />
+  <Target Name="EnsureNuGetPackageBuildImports" BeforeTargets="PrepareForBuild">
```
amaitland: The `EnsureNuGetPackageBuildImports` targets need to be removed from this file and the others.

See the following for background
http://blog.davidebbo.com/2014/01/the-right-way-to-restore-nuget-packages.html
https://www.xavierdecoster.com/post/2014/03/06/migrate-away-from-msbuild-based-nuget-package-restore.html

build.ps1
```diff
@@ -1,11 +1,11 @@
 param(
     [ValidateSet("vs2013", "vs2015", "nupkg-only", "gitlink")]
     [Parameter(Position = 0)] 
-    [string] $Target = "vs2013",
+    [string] $Target = "vs2015",
```
amaitland: Are you deliberately upgrading the project to `VC++ 2015`? The documentation would need to be updated if this is the path your planning on taking, See #1983 the previous discussion on the topic.

CefSharp.Core/ManagedCefBrowserAdapter.cpp
```diff
@@ -78,7 +77,7 @@ void ManagedCefBrowserAdapter::CreateBrowser(BrowserSettings^ browserSettings, R
     CefString addressNative = StringUtils::ToNative(address);
 
     CefBrowserHost::CreateBrowser(window, _clientAdapter.get(), addressNative,
-        *browserSettings->_browserSettings, requestContext);
+        *browserSettings->_browserSettings, *requestContext);
```
amaitland: Was there an actual compiler error or were you just going by the `Intellisense` warning that there was a problem? The intellisense warning is safe to ignore.

As it's a `ref counted pointer`, I'd suggest changing this to `static_cast<CefRefPtr<CefRequestContext>>(requestContext)` instead. Passing in as a pointer should be safe in this instance, in general you shouldn't pass a `ref counted pointer` around as a normal `pointer`.

amaitland:
> This will require the cef version in appveyor to be updated to 3.3071.1649

@GrabzIt If you `tag` a release on the `cef-binary` project following the same structure as the others `appveyor` will build and upload packages to `Myget`.

Few passing comments made inline, best of luck 👍

amaitland:
> any news about this update ?

Further releases are blocked on #1203 

At this rate it's looking like you'll have to wait for `Chromium` to re-add support for the `Process Per Tab` process model.

GrabzIt:
Hi @amaitland I will make the changes. However someone raised the possibility of #1562 or fixing the blocking issue. What do you think?

CefSharp.Core/ManagedCefBrowserAdapter.cpp
GrabzIt: Will do I will need to read up on the difference.

build.ps1
GrabzIt: Sorry, no I wasn't I just compiled it in 2015. I will fix the target.

amaitland:
> However someone raised the possibility of #1562 or fixing the blocking issue. What do you think?

I find it very unlikely, you should investigate for yourself.

CefSharp.Core/ManagedCefBrowserAdapter.cpp
amaitland: https://bitbucket.org/chromiumembedded/cef/wiki/GeneralUsage.md#markdown-header-reference-counting
https://www.chromium.org/developers/smart-pointer-guidelines

GrabzIt:
I ended up moving to the async javascript objects which doesn't have the issue. The code has been in production for a month so far with no problems.

GrabzIt:
Thanks for all the background information. I have made the requested changes.

amaitland:
> I ended up moving to the async javascript objects which doesn't have the issue. The code has been in production for a month so far with no problems.

The problem described in #1203 affects **both** `sync` and `async` versions. They share the same code for large parts.

CefSharp.Core/Internals/CefBrowserHostWrapper.cpp
```diff
@@ -20,7 +20,7 @@ void CefBrowserHostWrapper::DragTargetDragEnter(IDragData^ dragData, MouseEvent
 
     auto dragDataWrapper = static_cast<CefDragDataWrapper^>(dragData);
     dragDataWrapper->ResetFileContents(); // Recommended by documentation to reset before calling DragEnter
-    _browserHost->DragTargetDragEnter(dragDataWrapper, GetCefMouseEvent(mouseEvent), (CefBrowserHost::DragOperationsMask) allowedOperations);
+    _browserHost->DragTargetDragEnter(*dragDataWrapper, GetCefMouseEvent(mouseEvent), (CefBrowserHost::DragOperationsMask) allowedOperations);
```
amaitland: This pointer reference should also be changed.

GrabzIt:
> The problem described in #1203 affects both sync and async versions. They share the same code for large parts.
It may be because my browser instances are so short lived that I am getting away with it.

I have updated the pointer in the mouse drag code as requested.

amaitland:
> It may be because my browser instances are so short lived that I am getting away with it.

More like your navigating to websites within the same domain???

GrabzIt:
> More like your navigating to websites within the same domain???

Actually I had exactly that problem! On my US servers calls to www.bbc.co.uk would go to www.bbc.com and break. It took two days of working on the issue, but I eventually read this case #1782 where he said he had a similar issue with losing registered objects and switched to async that fixed the issue.

So I converted all of my C# JavaScript calls using the async methods and made all of my JavaScript async as well (using let and async keywords etc) and it worked. I haven't had anymore of those problems.

amaitland:
> Actually I had exactly that problem!

And your using a version greater than `57`? In versions greater if the browser process switches to a new sub process, the bound objects are never registered in the new process..  The `sync` and `async `versions use the same code to transmit the objects, unless `ClientAdapter::OnRenderViewReady` is now being called for the creation of **every** process then it's not possible for one to work and the other to be broken. It maybe possible with a newer `Chromium` version the process switching is less aggressive and your domains aren't spawning a new process.

https://github.com/cefsharp/CefSharp/blob/d1681162a68adb8af47328aaa3c15e5841c9b126/CefSharp.Core/Internals/ClientAdapter.cpp#L562

GrabzIt:
Oh ok, maybe that is what it is. I am now using version 60, but the fix also worked on version 59.

amaitland:
> Oh ok, maybe that is what it is. I am now using version 60, but the fix also worked on version 59.

Tried your new branch and still breaks as expected, even using bbc.co.uk and then navigating to bbc.com the render process switches and the bound objects are no longer available. Unclear what your doing differently, 

Anyways, I've spent way more time than I was planning on this, so that's it from me for now. I'll check back in a few months (probably when I have time off over Christmas). Best of luck getting a release out 👍

GrabzIt:
Thanks for all your help! You are a legend!

I am not sure why it is working on my version. One possibility is that I am setting the --renderer-process-limit to 1. So maybe that is related.

I haven't made any more changes to the cefsharp code to  handle it though. I will keep an eye on the #1203 case and release when it is fixed. You are right it is better to be safe than breaking everyone's projects!

PR #2130, opened 2017-08-29
Change the targets to BeforeBuild

nisbus:
When the copying of files is done after build and you sign a clickonce distribution the copied files will not match the manifests hash anymore.
Changing it to BeforeBuild got it working again for me.


perlun:
Makes sense, thanks! Merging this one.

Issue #2154, opened 2017-10-05
Buggy and difficult to test AsyncExtensions 


alfein:
- **What version of the product are you using?**
    - What version are you using? Nuget? CI Nuget? build from a branch? If so which branch?  Nuget 57.0.0

- **What architecture x86 or x64?** x64

- **On what operating system?** Win7

- **Are you using `WinForms`, `WPF` or `OffScreen`?**  WPF

- **What steps will reproduce the problem?**
AsyncExtensions class is a static class that implements in C# some of the async functions that used to be part of the  ICookieManager interface. 

If ICookieManager.SetCookie or ICookieManager.DeleteCookies or ICookieManager.VisitAllCookies or ICookieManager.FlushStore returns false the AsyncExtensions functions return null.  This violates expecteations because if our code uses await on the async methods, the await throws nullref.  You should return a Task<false> instead.

The Task...Callback wrappers

* TaskCompletionCallback
* TaskDeleteCookiesCallback
* TaskCookieVisitor
* TaskSetCookieCallback

all use your TrySetResultAsync extension method  method instead of TrySetResult, and you drop the async task itself.  This prevents us from completing the task synchronously during unit tests!  We used to be able to to this, becuase we could mock the actual async methods on the interface, but now, the only thing we can mock is the SetCookie method on CookieManager:

            _cookieManager.SetCookie(Arg.Any<string>(), Arg.Any<CefSharp.Cookie>(), Arg.Any<ISetCookieCallback>()).Returns(ci =>
            {
                var callback = (ISetCookieCallback)ci[2];

                var tcs = (TaskCompletionSource<bool>) typeof(TaskSetCookieCallback).GetField("taskCompletionSource", System.Reflection.BindingFlags.NonPublic | System.Reflection.BindingFlags.Instance).GetValue(callback);
                tcs.SetResult(true);
                callback.OnComplete(true);
                return true;
            });

And we have to do reflection as above to force the task completion to be synchrounous.  You can make OnComplete return a Task which we can await it?

amaitland:
> If ICookieManager.SetCookie or ICookieManager.DeleteCookies or ICookieManager.VisitAllCookies or ICookieManager.FlushStore returns false the AsyncExtensions functions return null. This violates expecteations becuase if our code uses await on the async methods, the await throws nullref. 

That's a fair point, you should submit a `PR`.

alfein:
Done

alfein:
[...]

PR #2155, opened 2017-10-07
Fixing Async methods to always return Tasks.  Making them easier to unit test.

alfein:
Fixing issue: Buggy and difficult to test AsyncExtensions #2154

Modified TaskExtensions to always return a Task, because it's not a good practice to return a null, since clients could be awaiting the result, which assumes that a Task is always returned.

Had to expose Dispose() method directly on the callback for the Callbacks and the Visitor classes. Calling dispose if the result of the operation is a false.

Also exposed a Task that's created when attempting to set the result of the task asynchronously - need this to await in unit tests.  

Replaced Task.Factory.StartNew with its more modern equivalent (Task.Run).

### DIFF
--- a/CefSharp/Callback/ICompletionCallback.cs
+++ b/CefSharp/Callback/ICompletionCallback.cs
@@ -3,7 +3,6 @@
 // Use of this source code is governed by a BSD-style license that can be found in the LICENSE file.
 
 using System;
-using System.Threading.Tasks;
 
 namespace CefSharp
 {
@@ -17,9 +16,5 @@ namespace CefSharp
         /// </summary>
         void OnComplete();
 
-        /// <summary>
-        /// Returns a Task that was used to set the result of the CookieCallback 
-        /// </summary>
-        Task SetResultTask { get; }
     }
 }
--- a/CefSharp/Callback/ICompletionCallback.cs
+++ b/CefSharp/Callback/ICompletionCallback.cs
@@ -3,6 +3,8 @@
 // Use of this source code is governed by a BSD-style license that can be found in the LICENSE file.
 
 using System;
+using System.Threading.Tasks;
+
 namespace CefSharp
 {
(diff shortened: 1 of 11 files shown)


CefSharp/Callback/ICompletionCallback.cs
```diff
@@ -14,5 +16,10 @@ public interface ICompletionCallback : IDisposable
         /// Method that will be called once the task is complete. 
         /// </summary>
         void OnComplete();
+
+        /// <summary>
+        /// Returns a Task that was used to set the result of the CookieCallback 
+        /// </summary>
+        Task SetResultTask { get; }
```
amaitland: The Task had no place being part of the call back and needs to be removed from all instances

amaitland:
Please give this PR a meaningful name

Will give a proper review in a few months, more changes will be required

CefSharp/Callback/ICompletionCallback.cs
alfein: Then how do I wait for the result to be set?

amaitland: How is that relevant to the interfaces?

alfein: Removed.  Also pull in your change.

amaitland:
Closing this as I don't believe it's necessary to expose the inner Task for testing purposes. See https://github.com/cefsharp/CefSharp/blob/master/CefSharp.Test/Framework/AsyncExtensionFacts.cs#L37 for an example of how a callback can be tested, mimics the real world usage scenario where Continue and Dispose are always called on one of the CEF threads. If you are unhappy with this you are totally free to implement the callback interface in your code anyway you like. By all means create a new PR that adds more test cases. 

The underlying bug where the extensions incorrectly returned null was fixed a while ago.

Issue #1437, opened 2015-11-23
ChromiumWebBrowser URL changed event


xmedeko:
I would like to have an event to inform for every URL change in `ChromiumWebBrowser`, event for HTTP 302 (and similar redirects). I have tried events `FrameLoadStart`, `FrameLoadEnd`, but they do not pass an URL which results into HTTP 302.

amaitland:
Check out `IRequestHandler`.

https://github.com/cefsharp/CefSharp/blob/cefsharp/43/CefSharp/IRequestHandler.cs#L146

Also please include which version of `CefSharp` your using.

xmedeko:
I use CefSharp 41. I see the this project develops rapidly. `IRequestHandler.OnBeforeBrowse` did a job. Thanks.

Note: some abstract class implementing IRequestHandler with default behaviour would be handy, when I need override just one method. (I know there is and example).

amaitland:
> I use CefSharp 41. I see the this project develops rapidly.

Yes, `41` is a few versions out of date.

> Note: some abstract class implementing IRequestHandler with default behaviour would be handy, when I need override just one method. (I know there is and example).

Your welcome to submit a `PR` implementing such.

PR #2157, opened 2017-10-09
Added default implementation of IRequestHandler

martinekvili:
This PR was requested by @amaitland in issue #1437, but as far as I was able to tell, it was never implemented.


perlun:
@martinekvili Thanks for the PR. Could you resolve the conflicts? We can then review it.

martinekvili:
@perlun Thanks, the conflicts should be sorted now.

perlun:
@amaitland Since you requested the original change, I think it would make most sense if you review this. Let me know if you are unable to do this and we'll let someone else give it a look.

amaitland:
@perlun Seems reasonable, will have a look 👍

CefSharp/DefaultRequestHandler.cs
```diff
@@ -0,0 +1,96 @@
+﻿// Copyright © 2010-2017 The CefSharp Authors. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style license that can be found in the LICENSE file.
+
+using System.Security.Cryptography.X509Certificates;
+
+namespace CefSharp
```
amaitland: At a minimum this should be in the `Handler` folder. Probably the `CefSharp.Handler` namespace as well. (Ideally all the handlers should be in their own namespace, moving the interfaces is a massive breaking change, so probably won't happen).

CefSharp/DefaultRequestHandler.cs
```diff
@@ -0,0 +1,96 @@
...
+{
+    /// <summary>
+    /// Default implementation of <see cref="IRequestHandler"/>.
+    /// This class provides default implementations of the methods from <see cref="IRequestHandler"/>,
+    /// therefore providing a convenience base class for any custom request handler.
+    /// </summary>
+    public class DefaultRequestHandler : IRequestHandler
+    {
+        public virtual bool GetAuthCredentials(IWebBrowser browserControl, IBrowser browser, IFrame frame, bool isProxy, string host,
```
amaitland: Ideally the `xmldoc` would be copied from the parent `IRequestHandler`. `GhostDoc` is handy when it comes to this sort of thing.

martinekvili:
@amaitland Thank you very much for the feedback! I think I implemented all the requested changes.

perlun:
@amaitland AppVeyor has stalled on this one for some reason, but I'm fine with you merging this nonetheless.

PR #2172, opened 2017-10-24
Add support UrlRequestFlags

AgapovAlexsey:
Can load page with skip cache


CefSharp.Core/Internals/CefRequestWrapper.cpp
```diff
@@ -153,6 +153,13 @@ namespace CefSharp
             return _wrappedRequest->IsReadOnly();
         }
 
+		void CefRequestWrapper::SetFlags(CefSharp::UrlRequestFlags flags)
```
jankurianski: Please change from tabs to spaces.

CefSharp.Core/Internals/CefRequestWrapper.h
```diff
@@ -51,6 +51,7 @@ namespace CefSharp
             virtual property TransitionType TransitionType { CefSharp::TransitionType get(); }
             virtual property IPostData^ PostData { IPostData^ get(); }
             virtual property bool IsReadOnly { bool get(); }
+			virtual void SetFlags(CefSharp::UrlRequestFlags flags);
```
jankurianski: Tabs to spaces.

CefSharp/Enums/UrlRequestFlags.cs
```diff
@@ -0,0 +1,41 @@
+﻿using System;
```
jankurianski: Please add copyright header as seen in `UrlRequestStatus.cs`

CefSharp/Enums/UrlRequestFlags.cs
```diff
@@ -0,0 +1,41 @@
...
+
+namespace CefSharp
+{
+    [Flags]
```
jankurianski: Please add summary comment as in `UrlRequestStatus.cs`.  The CEF docs say:

> Flags used to customize the behavior of CefURLRequest.

CefSharp/IRequest.cs
```diff
@@ -91,6 +92,12 @@ public interface IRequest : IDisposable
         bool IsReadOnly { get; }
 
         /// <summary>
+        /// Set the flags used in combination with CefURLRequest
+        /// </summary>
+        /// <param name="flags">Request Flags</param>
+        void SetFlags(UrlRequestFlags flags);
```
jankurianski: See line `28` in this file?  I think it was someone's earlier intention that the flags be implemented via a `Flags` property with `get`/`set` accessors.

What do you think about implementing the getter and changing the interface to be a property instead of methods?

AgapovAlexsey: I have tried implemented  getFlags, but it always return 0.

jankurianski: OK fair enough.  Could you please post a comment when you are able to change the formatting and comments?

AgapovAlexsey: Yes. I will do it 28.10.2017

AgapovAlexsey: why is it  fail? :)

perlun:
@jankurianski @AgapovAlexsey What is the status of this one now?

AgapovAlexsey:
@perlun I did all, but appveyor failed. May be i did something wrong?

jankurianski:
I think appveyor built this PR at a time when a nuget package was failing to install.  It looks good now.

Thanks for fixing formatting and adding changes @AgapovAlexsey - next time please leave a comment on the PR once you have pushed further changes to your branch, because GitHub does not notify the reviewer and it is easy to miss.

perlun:
> Thanks for fixing formatting and adding changes @AgapovAlexsey - next time please leave a comment on the PR once you have pushed further changes to your branch, because GitHub does not notify the reviewer and it is easy to miss.

We can also make AppVeyor send a message to the PR once it is finished with its work. I'll activate that setting now.

PR #2178, opened 2017-10-27
initial vs2017 build support (with vs2015 tooling)

japj:
This implementation uses VS2017 with the VS2015 toolset (use the VS2017 installer to install this toolset).

I extracted the PostBuild script into the CefSharp.props file and tried to keep the behaviour the same for non-VS2017 (however, I cannot test that without VS2013 and VS2015).

Note though that the CefSharp.Test OffScreenBrowserBasicFacts tests are failing and I am unsure if that is expected (since they were only recently added and I don't see a succesfull appveyor build yet).


japj:
I redid the pull request in #2179, closing this one

perlun:
(Note: technically, you don't have to redo a PR. You can just force-push the branch and make changes to the original PR that way.)

PR #2179, opened 2017-10-29
VS2017 build support (using the VS2015 tooling)

japj:
note: I redid the pull request for a more clean version.

Added support for VS2017, using the VS2015 toolset (use the VS2017 installer to install this toolset).
I also added a config file for the "Microsoft Child Process Debugging Power Tool" which will allow automatic debugging of child processes.

I updated the nuget url, since appveyor could not install xunit.abstractions 2.0.1 with the old nuget.

I extracted the PostBuild script into the CefSharp.props file and tried to keep the behaviour the same for non-VS2017 (however, I cannot test that without VS2013 and VS2015, but it succeeds on appveyor).

Note though that the CefSharp.Test OffScreenBrowserBasicFacts tests are failing and I am unsure if that is expected (since they were only recently added and I don't see a succesfull appveyor build yet).

### DIFF
--- a/CefSharp.BrowserSubprocess/CefSharp.BrowserSubprocess.csproj
+++ b/CefSharp.BrowserSubprocess/CefSharp.BrowserSubprocess.csproj
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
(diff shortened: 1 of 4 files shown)


amaitland:
Changed to target `v141` toolset, so no longer be necessary to install `VC++2015` tooling when using `VS2017`. Commit https://github.com/cefsharp/CefSharp/commit/05f0fa025afcf2f65221cd46ab00fdbacb787f7f

McXX:
The new version still uses VS2015 libs from cef.sdk package as VisualStudioProductVersion was not changed.

McXX:
Sorry, in fact there's no 2017 version of cef.sdk libraries yet.

PR #2198, opened 2017-11-15
Update CONTRIBUTING.md

perlun:
Added note about short commit messages.

I believe this is a good habit, since it makes `git log` output etc much easier to skim through. What do you think?

(9407ba18 is an example of a bad commit message from this POV, but it seems to be very common.)


amaitland:
50 characters seems too short to me, the `GitHub` web interface truncates at roughly 70 characters. What software are you using that has a limit of 50?

perlun:
Funnily enough, GitHub. 😄 

image

I can live with 70 characters also, so please change it to that if you prefer it. The importance for me personally is to get a better output from `git log --graph --oneline --all --decorate --tags` which is a command I have aliased to `git tree`, and gives a really nice output:

image

But this works best if we try to keep the first line as a reasonably short summary. (The example above doesn't look that great, because of the long commit messages.)

Issue #1479, opened 2015-12-09
Proprietary Codec Support (H264, MP3)


Feko:
Hello,

I'm trying to play audio, using <audio> tag and CefSharp. No success so far.
After long searches, seems like I need ffmpegsumo.dll from the same exact version as CEF. (https://github.com/cefsharp/CefSharp/issues/444)
But, I'm using CEF 3.2526.1359, and as described here: https://codereview.chromium.org/1141703002 , CEF3 now statically links ffmpeg... What is "statically link"? What should I do to be able to play audio?

Thanks in advice!

Feko:
Just as a test, I tried another release of CefSharp: Release 43, CEF 3.2357.1287
This time, with ffmpegsumo.dll provided by default. Same issue as described in https://github.com/cefsharp/CefSharp/issues/444 Audio controls appear, but speaker is disabled, and no sound.
Tried downloading CEF release from cefbuilds.com (3.2357.1287.g861c26e_windows32) and using those dlls... same thing. What am I doing wrong?

amaitland:
> What is "statically link"?

The code has been compiled into the `libcef.dll`, so there is no longer a need for a separate dll.

> After long searches, seems like I need ffmpegsumo.dll from the same exact version as CEF. (#444)

Did you try the example `URL` linked in #444? It plays fine for me?

> Just as a test, I tried another release of CefSharp: Release 43, CEF 3.2357.1287

For this release you would require `ffmpegsumo.dll`

> What am I doing wrong?

Are you trying to play audio that requires a proprietary codec? e.g. `mp3`? Licensing restrictions prohibit `CEF/CefSharp` from distributing binary's that provide support.

Feko:
Hi, thanks for your reply :)
Oh sorry, I didn't tried that example, I was working on a test page I've created.
I was trying to play MP3/M4A... After your answer, I tried OGG Vorbis on CefSharp 43 + ffmpegsumo.dll, and it worked just fine!!
But in this project, I really need support to those proprietary formats... How can I achieve that?

Thanks!

amaitland:
> But in this project, I really need support to those proprietary formats... How can I achieve that?

Quite a lot of effort is required. See https://groups.google.com/d/msg/cefsharp/BJLMXl9c204/HMJlp8mZzF0J for a guide (little outdated).

You'd also need to look at acquiring the relevant licenses to use/distribute those codecs.

amaitland:
[...]

PR #2204, opened 2017-12-03
[DO NOT MERGE] Use experimental cef.sdk and cef.redist packages

perlun:
These packages are custom-built by me, and includes proprietary codecs. The legal status of using these packages is questionable, so use them at your own risk.

They were built using this build script, and these Chromium flags:

```bat
set CEF_USE_GN=1
set GN_DEFINES=is_official_build=true proprietary_codecs=true ffmpeg_branding=Chrome
set GYP_DEFINES=buildtype=Official
set GYP_MSVS_VERSION=2015
set CEF_ARCHIVE_FORMAT=tar.bz2
set DEPOT_TOOLS_UPDATE=0

python ..\automate\automate-git.py --download-dir=. --depot-tools-dir=..\depot_tools --minimal-distrib --client-distrib --branch=3202 --force-build --no-debug-build
```

See https://github.com/cefsharp/cef-binary/pull/62 for the CEF version being used here.

This is a temporary workaround for #1479. Once CEF 63 has landed, we should make an official CefSharp version based on that; this is definitely just for the time being. **You may run into legal trouble if you use these packages**, especially in certain jurisdictions.

The unofficial packages are here: https://www.myget.org/feed/Packages/perlun

### DIFF
--- a/CefSharp.BrowserSubprocess.Core/CefSharp.BrowserSubprocess.Core.vcxproj
+++ b/CefSharp.BrowserSubprocess.Core/CefSharp.BrowserSubprocess.Core.vcxproj
@@ -1,6 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project DefaultTargets="Build" ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <Import Project="..\packages\cef.sdk.3.3202.1686\build\cef.sdk.props" Condition="Exists('..\packages\cef.sdk.3.3202.1686\build\cef.sdk.props')" />
+  <Import Project="..\packages\cef.sdk.3.3202.1690-proprietary-codecs\build\cef.sdk.props" Condition="Exists('..\packages\cef.sdk.3.3202.1690-proprietary-codecs\build\cef.sdk.props')" />
   <ItemGroup Label="ProjectConfigurations">
     <ProjectConfiguration Include="Debug|Win32">
       <Configuration>Debug</Configuration>
@@ -230,6 +230,6 @@
     <PropertyGroup>
       <ErrorText>This project references NuGet package(s) that are missing on this computer. Use NuGet Package Restore to download them.  For more information, see http://go.microsoft.com/fwlink/?LinkID=322105. The missing file is {0}.</ErrorText>
     </PropertyGroup>
-    <Error Condition="!Exists('..\packages\cef.sdk.3.3202.1686\build\cef.sdk.props')" Text="$([System.String]::Format('$(ErrorText)', '..\packages\cef.sdk.3.3202.1686\build\cef.sdk.props'))" />
+    <Error Condition="!Exists('..\packages\cef.sdk.3.3202.1690-proprietary-codecs\build\cef.sdk.props')" Text="$([System.String]::Format('$(ErrorText)', '..\packages\cef.sdk.3.3202.1690-proprietary-codecs\build\cef.sdk.props'))" />
   </Target>
 </Project>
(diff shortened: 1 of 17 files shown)


mitchcapper:
It seeps the depency of CEfSharp.WPF on CefSharp.Common is not quite correct:
Attempting to resolve dependencies for package 'CefSharp.Wpf.62.0.0-proprietary-codecs' with DependencyBehavior 'Lowest'
Unable to find a version of 'CefSharp.Common' that is compatible with 'CefSharp.Wpf 62.0.0-proprietary-codecs constraint: CefSharp.Common (= 62.0.0-CI2368)'.

Not a big deal installing both packages manually still works.

amaitland:
I see you published these to Nuget.org, which has my name against the packages, I wish to make it very clear that I had nothing to do with providing packages that contain proprietary codes and won't take any responsibility for any legal liability.

perlun:
@amaitland 

> I see you published these to Nuget.org, which has my name against the packages

To make that clear - yes, it had nothing to do whatsoever with you (unfortunately the packages published via MyGet are tagged that way, probably because the MyGet connection was set up using your user name.) So I take the full responsibility for this and any legal action for this should be brought towards me. Sorry that your name was used, it was not my intention here.

(Having that said, this is not a matter of breaking any copyrights to the best of my knowledge - the use of patented algorithms is the big issue here.)

The reason why I needed to push this to NuGet is because of a customer who needed these packages. I did clearly mark them as a "prerelease" though, indicating that this is **not** a mainstream release and it can **not** become such a one, because of these mentioned legal reasons.

@mitchcapper 

> It seeps the depency of CEfSharp.WPF on CefSharp.Common is not quite correct:

Thanks, I noted that also while doing a test upgrade in a project - published new packages with fixed dependencies.

amaitland:
> Sorry that your name was used, it was not my intention here.

@perlun Whilst it was not your direct intention that is the case unfortunately. 

> (Having that said, this is not a matter of breaking any copyrights to the best of my knowledge - the use of patented algorithms is the big issue here.)

I have yet to seek legal advice regarding distribution of `CEF` with `proprietary codec` support as I had never planned to do so. As the unfortunate result of my `Nuget.org` credentials being used to push those packages, I have unlisted those versions, removed the links above and revoked my `API` keys. Whilst some may as this is an overreaction it is a very clear and simple choice for me. As I've made it clear that I am aware of the packages taking swift action should limit any legal liability on my part. If there was ever a problem and I was found to have done nothing, that's a different story.

> The reason why I needed to push this to NuGet is because of a customer who needed these packages. I did clearly mark them as a "prerelease" though, indicating that this is not a mainstream release and it can not become such a one, because of these mentioned legal reasons.

You really should be using `MyGet` as your playground and not `Nuget.org`, it's very easy to create your own feed and publish packages there. I guess I made things too easy with automation of building and pushing packages.

vwmberry95:
I've been watching this closely for the past week or so, and for those of us that have legal H264 licensing, thank you!  I've had a hell of a time getting proprietary codecs enabled with the stable release of CefSharp on v57.

perlun:
> I've been watching this closely for the past week or so, and for those of us that have legal H264 licensing, thank you!

You're welcome. One thing worth checking out is that `ffmpeg_branding=Chrome` _might_ mean that the Chromium code is compiled with GPL licensed libraries included. These can _not_ be combined with non-GPL libraries without very serious legal consequences.

I haven't found the conclusive list of compiler flags being used for `ffmpeg` with this setting; if anyone else finds it, please paste the link here so we can better analyze this.

mitchcapper:
@perlun calling avutil_configuration (https://ffmpeg.org/doxygen/1.2/group__lavu__ver.html) should return the exact compile args for the library.

perlun:
Closing this for now, it's gotten stale anyway. Once #1479 has landed, we can all shout for joy. 🎉 Probably a few months away though.

PR #2218, opened 2017-12-26
Dont discard null Values from arrays

joaompneves:
When passing arrays between .Net and JS null values were discarded.
Eg: 
[ "a", null, "c" ] -> [ "a", "c" ]

Fixed behavior:
[ "a", null, "c" ] -> [ "a", null, "c" ]


amaitland:
Please add a unit test or a functional test.

Note that `BindingTest.html` has been rewritten to use `QUnit`.

joaompneves:
Added a test as requested

CefSharp.Example/AsyncBoundObject.cs
```diff
@@ -64,6 +64,10 @@ public JsSerializableClass[] ReturnClassesArray(string name)
             };
         }
 
+        public string[] ReceiveArray(string[] arg) {
```
amaitland: Braces on new line.

Probably be better if this method was called something like `EchoArray`

CefSharp.Example/Resources/BindingTest.html
```diff
@@ -216,6 +216,19 @@
                 });
             });
 
+            QUnit.test("Async call (receiveArray):", function (assert) {
```
amaitland: Braces on new line, same goes for below

CefSharp/ModelBinding/DefaultBinder.cs
```diff
@@ -114,17 +114,14 @@ public virtual object Bind(object obj, Type modelType)
                 {
                     var val = GetValue(bindingContext, i);
 
-                    if (val != null)
+                    if (val != null && typeof(IDictionary<string, object>).IsAssignableFrom(val.GetType()))
```
amaitland: I don't think this change reads very well, think something like the following would read better, also added some extra checks.

- Assuming it's possible to have a multi dimensional array, haven't tested that yet.
- If it's an array of value types and you pass in null then use `Activator.CreateInstance` to return the default value for the type. Trying to avoid throwing exceptions as currently there is no clearly defined place to catch them.



```c#
//If the value is null then we'll add null to the collection,
if(val == null)
{
	//For value types like int we'll create the default value and assign that as we cannot assign null
	model.Add(genericType.IsValueType ? Activator.CreateInstance(genericType) : null);
}
else
{
	var valueType = val.GetType();
	//If the collection item is a list or dictionary then we'll attempt to bind it
	if (typeof(IDictionary<string, object>).IsAssignableFrom(valueType) ||
		typeof(IList<object>).IsAssignableFrom(valueType))
	{
		var subModel = Bind(val, genericType);
		model.Add(subModel);
	}
	else
	{ 
		model.Add(val);
	}
}
```

CefSharp.Example/Resources/BindingTest.html
joaompneves: Unfortunatelly I have different settings on VS, and have to change all these by hand. :(

CefSharp/ModelBinding/DefaultBinder.cs
joaompneves: Done, added tests to cover these scenarios

amaitland:
Thanks for updating 👍 Still a few formatting issues, will fix those after merge. Probably next early next week.

PR #2219, opened 2017-12-27
Fix enums, integer and doubles binding behavior

joaompneves:
Bind behavior of the default binder was broken for enums. 
Eg: trying to convert an integer to an enum always resulted in the enum default value (0).

Convertion from doubles to ints or ints to doubles was also broken. Now it tries to convert with precision loss if applicable.

JS -> .Net
1 -> B (enum { A, B } )
2.4 -> 2 (int)
2.6 -> 3 (int)
2 -> 2 (double)


perlun:
Looks good to me. @amaitland, I think you're the one who's done most of the work on this file so I wouldn't mind you reviewing it also before we merge.

amaitland:
@perlun I haven't had a chance to see the unit tests run, so was waiting to merge. I had added it to the `63` milestone as I do have plans to merge it.

perlun:
> @perlun I haven't had a chance to see the unit tests run, so was waiting to merge. I had added it to the 63 milestone as I do have plans to merge it.

Ah, sorry to rush you! 😄 Thanks for the merge.

Regarding the test runs - would it make sense if we added the test output to AppVeyor, so we could more easily see results of tests (without having to download the branch and test it locally)?

amaitland:
@perlun Excellent suggestion, feel like creating an issue to represent the idea?

perlun:
Done - see #2227.

PR #2225, opened 2018-01-06
Request to add a github link to the "Link" list.

mattkol:
Hi All,

This is is the link to a project I have been working on for sometime now, and would like to share with the community. The project uses CefSharp, CefGlue.

I hope it meets the required standards.

Thank you.


perlun:
Thanks for the suggestion @mattkol! Looks like an interesting project. I would suggest we add another section called `## Projects using CefSharp` or something like that. That makes it a bit more clear that this is not _specifically_ related to CEF or similar, but more a "use case" for the project.

If you're fine with that, just go ahead and amend the PR and I'll give a 👍 afterwards.

mattkol:
@perlun the PR is updated as you suggested.
Thanks for looking into this.

PR #2231, opened 2018-01-10
Honor js camel case setting

joaompneves:
When an object is registered with the BindingOptions.CamelCaseJavascriptNames on, this setting is not completly honored. Before the fix, this setting only converts the names of the object being registered but not the names of the objects that cross managed boundaries.

This changes that behavior and honors the setting provided.
If you think that this change might great impact on existing consumers I can add an extra property. What do you think?


amaitland:
Huge impact on existing users, massive breaking change, as it stands there is no chance that I'd merge this. 

Even if you make it configurable I'm not keen on this change. Someone else might be interested, you'll have to convince them it's worth the time.

joaompneves:
I agree with you that this is a huge breaking change, thats why I would like to control the behavior with a side setting.

The current implementation is **misleading**, since the **camel case behavior is not honored**. And there's no way with current mechanisms to change that behavior. One can override the default binder to change inbound data, but cannot override outbound data serialization (at least in efficient way).
This forces consumers to have code that is against target languages conventions (JS and C#). Either we use camel case fields in C# or Pascal Case fields in JS. In any case, code linters need to be turned off which or any third party software used needs to be manipulated or wrapped to make the changes needed. Neither seems acceptable.

Regarding change proposed, its not a big deal (if you look closely). The majority of the changes is to pass the setting value to the serialization methods. Even in a maintenance perspective I don't think it introduces great concerns.

amaitland:
> The current implementation is misleading, since the camel case behavior is not honored.

It's taken 2+yrs for this to be mentioned as an issue,

> The change proposed, its not a big deal (if you look closely).

There are `21` files changed, so it's hardly trivial.

> Even in a maintenance perspective I don't think it introduces great concerns.

I'm still not interested. I dislike the whole `CamelCase` feature, it's only there as a legacy. If someone else is keen to review and merge this then I'm not stopping anyone.

perlun:
> I'm still not interested. I dislike the whole CamelCase feature, it's only there as a legacy. If someone else is keen to review and merge this then I'm not stopping anyone.

I'm probably the one that introduced it a few years ago... 😄 Like @joaompneves, I dislike having to use the "wrong" casing in either land (C#/.NET or JS.) But I also understand this might be an expensive feature to carry around in our dear codebase.

My _suggestion_ would be to either:

1. Go with @joaompneves ideas and try harder to make it work "as one would expect", _or_
2. Kill the whole thing, and just let things be named whatever they are named. This is the KISS approach, arguably.

Should we consider doing option 2?

amaitland:
Replaced by #3141

PR #2232, opened 2018-01-12
tooltip updates it position when the tooltip text is changed

DanielBretzigheimer:


CefSharp.Wpf/ChromiumWebBrowser.cs
```diff
@@ -1,4 +1,4 @@
-﻿// Copyright © 2010-2017 The CefSharp Authors. All rights reserved.
+// Copyright © 2010-2017 The CefSharp Authors. All rights reserved.
```
amaitland: Revert unnecessary formatting changes please.

CefSharp.Wpf/ChromiumWebBrowser.cs
```diff
@@ -1967,6 +1967,10 @@ private void UpdateTooltip(string text)
             }
             else
             {
+                // hide old tooltip before showing the new one to update the position
+                if (toolTip.Content as string != text)
```
amaitland: Is the if statement even required? Is there a scenario where the same tooltip text is sent through?

CefSharp.Wpf/ChromiumWebBrowser.cs
DanielBretzigheimer: Sorry but my git client does not show me a change at this line (and as you see the two lines are identical). So I am not able to undo this line.

CefSharp.Wpf/ChromiumWebBrowser.cs
DanielBretzigheimer: It is possible to get called when the tooltip was hidden because the mouse was over an area where no tooltip is shown. But in this case the tooltip.IsOpen is already false so no need to set it. But yeah, technically it can be set to IsOpen every time.

CefSharp.Wpf/ChromiumWebBrowser.cs
amaitland: Your editor has likely changed the files codepage.

The lines may look identical, if you did a binary diff they'd be different.

DanielBretzigheimer: should be fixed now

PR #2239, opened 2018-01-18
WPF - ChromiumWebBrowser inherit from Control rather than UserControl

amaitland:
Based on #1499

Additionally Popup is created in xaml
Canvas used to absolute position the two images, one on top of the other
No Mouse.Capture required
Much simpler implementation, not very well tested yet.


amaitland:
Once this is tested with on a `High DPI` screen it will be merged, won't be included in the `cefsharp/63` branch though.

amaitland:
Will look at merging this after `65` has been branched.

amaitland:
Follow up https://github.com/cefsharp/CefSharp/commit/61e94115d8c12e35f4f9fcc2879fc6f974104356

PR #2244, opened 2018-01-19
Support uncaught exception handling

joaompneves:
Render Process Message Handler now supports uncaught exception handling. This is useful when developers need to be notified when a script error occurs and need to get information about the exception message and its stacktrace.

Introduced the OnUncaughtException method in RenderProcessMessageHandler.


CefSharp.Core/Internals/Messaging/Messages.h
```diff
@@ -1,4 +1,4 @@
-// Copyright � 2010-2017 The CefSharp Authors. All rights reserved.
+// Copyright © 2010-2017 The CefSharp Authors. All rights reserved.
```
amaitland: Looks like you've changed the file encoding, unless absolutely necessary please revert.

CefSharp.Example/RenderProcessMessageHandler.cs
```diff
@@ -26,5 +26,10 @@ void IRenderProcessMessageHandler.OnContextReleased(IWebBrowser browserControl,
         {
             //The V8Context is about to be released, use this notification to cancel any long running tasks your might have
         }
+
+        public void OnUncaughtException(IWebBrowser browserControl, IBrowser browser, IFrame frame, JavascriptException exception)
```
amaitland: For consistency please change to explicit interface implementation.

CefSharp/Handler/IRenderProcessMessageHandler.cs
```diff
@@ -37,9 +37,18 @@ public interface IRenderProcessMessageHandler
...
         /// </summary>
         /// <param name="browserControl">The ChromiumWebBrowser control</param>
-        /// <param name="browser">the browser object</param>
+        /// <param name="browser">The browser object</param>
         /// <param name="frame">The frame object</param>
         /// <param name="node">An object with information about the node (if any) that has focus.</param>
         void OnFocusedNodeChanged(IWebBrowser browserControl, IBrowser browser, IFrame frame, IDomNode node);
+
+        /// <summary>
+        /// Invoked when an uncaught javascript exception occurs.
```
amaitland: > Called for global uncaught exceptions in a frame. Execution of this callback is disabled by default. To enable set CefSettings.uncaught_exception_stack_size > 0. 

Summary should be more closely based on the `CEF API Doc`

CefSharp/JavascriptException.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Copyright © 2010-2017 The CefSharp Authors. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style license that can be found in the LICENSE file.
+
+using System.Runtime.Serialization;
+
+namespace CefSharp
+{
+    [DataContract]
```
amaitland: Unless there is a reason `DataContract` and `DataMember` should be removed.

CefSharp.Core/Internals/Messaging/Messages.h
amaitland: Unless absolutely required revert file encoding change.

CefSharp/JavascriptException.cs
joaompneves: Just followed the same pattern that JavascriptResponse has

amaitland: JavaScriptResponse is used by the WCF service, hence the attributes.

amaitland: Do you have time to remove the attributes? Can be merged when that's done.

joaompneves: Done

Issue #2206, opened 2017-12-07
WinForms, cant add UserControl in design mode


ilbarzo:
I try to add a UserControl browser in a form. By adding it on code behind, I have no problem.
I can Build and use it without problem.
If I try to add the UserControl in design mode, by dragging the UC from toolbox I have this error:
![2017-12-07_10-56-04](https://user-images.githubusercontent.com/2872864/33709483-d4f5680c-db3d-11e7-9582-73be4451f1d7.jpg)

Environment:
Windows 10 x64
Visual Studio 2015
C++ redistributable installed from 2005 to 2015, both x86 and x64
CefSharp from NuGet master 57

Any idea?

avspeed:
I am not sure if that is fully implemented yet
Take a look at this issue
https://github.com/cefsharp/CefSharp/pull/1946

merceyz:
Please use the `Bug Report` template provided at https://github.com/cefsharp/CefSharp/blob/master/ISSUE_TEMPLATE.md#bug-report

Fork the https://github.com/cefsharp/CefSharp.MinimalExample project and provide an example that reproduces your problem.

mitchcapper:
The other thing you can do is fire up procmon and see where it is looking for the cefsharp dll's (maybe the project folder if not the output folder) and manually copy them there.

merceyz:
That isn't the problem in this case.
It's not supposed to load those files while in designer mode.

I tested and also got the same error so something has changed since the PR that fixed this

ilbarzo:
Hi @merceyz , I have forked the MinimalExample as suggested, added a winform and a usercontrol to reproduce the issue.
committed here: [https://github.com/ilbarzo/CefSharp.MinimalExample/commit/1029c87c7eb64bfee524b308fbef3e0cd94e7646](url)

tnx for support

amaitland:
Adding the `ChromiumWebBrowser` control using the designer is working fine for me. @merceyz Are you seeing problems with the control or adding a `User Control`?

> committed here: https://github.com/ilbarzo/CefSharp.MinimalExample/commit/1029c87c7eb64bfee524b308fbef3e0cd94e7646

@ilbarzo Your example has references to classes from `CefSharp.Core` for which you basically have three options for:

- Move all references to classes `CefSharp.Core` outside of any control that will be used by the designer
[...]

PR #2261, opened 2018-02-03
WinForms: Prevent designer from setting default values

merceyz:
This PR solves the issue described in https://github.com/cefsharp/CefSharp/issues/2206#issuecomment-362769476

The changes in this PR prevents the designer from setting the default value of specific properties which caused it to attempt to load `CefSharp.Core` causing an exception to be thrown

### DIFF
--- a/CefSharp.WinForms/ChromiumWebBrowser.cs
+++ b/CefSharp.WinForms/ChromiumWebBrowser.cs
@@ -73,7 +73,7 @@ namespace CefSharp.WinForms
         /// Gets or sets the request context.
         /// </summary>
         /// <value>The request context.</value>
-        [Browsable(false), EditorBrowsable(EditorBrowsableState.Never)]
+        [Browsable(false), EditorBrowsable(EditorBrowsableState.Never), DefaultValue(null)]
         public IRequestContext RequestContext
         {
             get { return requestContext; }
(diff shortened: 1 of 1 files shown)


amaitland:
Should we do the same with other properties like `BrowserSettings`?

merceyz:
We could but we don't have to. These were the only ones causing a issue.

If you add a control to a form the designer adds this code, so these would be the ones to add it to
```
this.chromiumWebBrowser1.BrowserSettings = null;
this.chromiumWebBrowser1.DialogHandler = null;
this.chromiumWebBrowser1.DisplayHandler = null;
this.chromiumWebBrowser1.DownloadHandler = null;
this.chromiumWebBrowser1.DragHandler = null;
this.chromiumWebBrowser1.FindHandler = null;
this.chromiumWebBrowser1.FocusHandler = null;
this.chromiumWebBrowser1.GeolocationHandler = null;
this.chromiumWebBrowser1.IsActivating = false;
this.chromiumWebBrowser1.JsDialogHandler = null;
this.chromiumWebBrowser1.KeyboardHandler = null;
this.chromiumWebBrowser1.LifeSpanHandler = null;
this.chromiumWebBrowser1.LoadHandler = null;
this.chromiumWebBrowser1.MenuHandler = null;
```

PR #2265, opened 2018-02-05
WinForms: Prevent designer from setting any default values

merceyz:
Continuation of https://github.com/cefsharp/CefSharp/pull/2261

This PR sets the default values for the rest of the properties that aren't visible to the designer


amaitland:
Thanks 👍 

Also merged into https://github.com/cefsharp/CefSharp/tree/cefsharp/63

PR #2280, opened 2018-02-18
Split CefFileDialogMode into CefFileDialogMode and CefFileDialogFlags

### LINKED ISSUE #2277: CefFileDialogMode should probably not have [Flags]
master at be50b75648eb75c4dfda38b3fa5c2b7ffcf93ca6

x86

Windows 10 1709 x64

WPF

Try to use `CefFileDialogMode` enum (see [CefFileDialogMode.cs](https://github.com/cefsharp/CefSharp/blob/be50b75648eb75c4dfda38b3fa5c2b7ffcf93ca6/CefSharp/Enums/CefFileDialogMode.cs#L12))

Due to how the enum is structured, I expect it to be a regular enum. However, it is marked as a flag enum.

See below.

Not applicable

---

The `CefFileDialogMode` enum has (since [ef21c492b67438cc100af09554ee6da717af7389](https://github.com/cefsharp/CefSharp/commit/ef21c492b67438cc100af09554ee6da717af7389#diff-e71583cdc041117b3d4197ea35cfb86d)) the `[Flags]` attribute. However, when looking at the values, that can’t be right. Here’s what a decompiler produces:

```csharp
[Flags]
public enum CefFileDialogMode
{
  Open = 0,
  OpenMultiple = 1,
  OpenFolder = 2,
  Save = OpenFolder | OpenMultiple, // 0x00000003
  TypeMask = 255, // 0x000000FF
  OverwritePrompt = 16777216, // 0x01000000
  HideReadOnly = 33554432, // 0x02000000
}
```

Surely, `Save` is not a combination of `OpenMultiple` and `OpenFolder`.

It appears the enum is somehow a mix of flags and non-flag values. While that may be appropriate in native code, I don’t think it is in .NET.

I’ll provide a pull request on this topic this weekend if that’s acceptable.

fuzzykiller:
Resolves #2277 by splitting the enum.

In CEF, `cef_file_dialog_mode_t` is a mix of flags and non-flags. In the .NET managed world, a combination like that rather unusual. As such, my suggestion is to split the enum into `CefFileDialogMode` and `CefFileDialogFlags`.

### DIFF
--- a/CefSharp.Core/Internals/ClientAdapter.cpp
+++ b/CefSharp.Core/Internals/ClientAdapter.cpp
@@ -1019,7 +1019,19 @@ namespace CefSharp
             auto browserWrapper = GetBrowserWrapper(browser->GetIdentifier(), browser->IsPopup());
             auto callbackWrapper = gcnew CefFileDialogCallbackWrapper(callback);
 
-            return handler->OnFileDialog(_browserControl, browserWrapper, (CefFileDialogMode)mode, StringUtils::ToClr(title), StringUtils::ToClr(default_file_path), StringUtils::ToClr(accept_filters), selected_accept_filter, callbackWrapper);
+            auto dialogMode = mode & FileDialogMode::FILE_DIALOG_TYPE_MASK;
+            auto dialogFlags = mode & ~FileDialogMode::FILE_DIALOG_TYPE_MASK;
+
+            return handler->OnFileDialog(
+                _browserControl,
+                browserWrapper,
+                (CefFileDialogMode)dialogMode,
+                (CefFileDialogFlags)dialogFlags,
+                StringUtils::ToClr(title),
+                StringUtils::ToClr(default_file_path),
+                StringUtils::ToClr(accept_filters),
+                selected_accept_filter,
+                callbackWrapper);
         }
 
         bool ClientAdapter::OnDragEnter(CefRefPtr<CefBrowser> browser, CefRefPtr<CefDragData> dragData, DragOperationsMask mask)
--- a/CefSharp.Example/TempFileDialogHandler.cs
+++ b/CefSharp.Example/TempFileDialogHandler.cs
@@ -9,10 +9,10 @@ namespace CefSharp.Example
 {
     public class TempFileDialogHandler : IDialogHandler
     {
-        public bool OnFileDialog(IWebBrowser browserControl, IBrowser browser, CefFileDialogMode mode, string title, string defaultFilePath, List<string> acceptFilters, int selectedAcceptFilter, IFileDialogCallback callback)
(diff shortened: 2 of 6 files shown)


fuzzykiller:
@merceyz I’m not sure what you might be referring to, could you please clarify?

amaitland:
Happy to merge a change that removes the `[Flags`] attribute, the proposed changes in this `PR` should be discussed in more detail, I have some other bugs that need to be taken care of as a priority before I can get to this.

fuzzykiller:
Simply removing the `[Flags]` attribute isn’t the right way. There *are* flags in that enum, after all. That’s why I chose to split the enum.

If that’s not acceptable (it’s a breaking change after all!), I’m also fine with that. However, the `[Flags]` attribute needs to stay in that case.

amaitland:
Removing the flags attribute is the simplest option, you  can perform the same bit operations as you did in this PR in c#. It really depends on how often the flags are actually used, for which I don't have time to investigate.

fuzzykiller:
Well, yes. On a simple open dialog triggered by `<input type="file">`, both `OverwritePrompt` and `HideReadOnly` are set. Doesn’t make any sense but that’s how it is. So simple comparisons with mutually exclusive values (which is what an absent `[Flags]` implies) won’t work. As such, `[Flags]` *must* stay unless the enum is split.

amaitland:
Flags must stay if the flag values are actually used, which was the part that I didn't have time to look into. If the flags were only set on rare edge cases then I would happily have removed the attrib

Is there a case when no flag is set? How will your change behave?

fuzzykiller:
After looking at the CEF and Chromium source code for a few hours I still wasn’t able to find out where those values are set. 😢 As such, I cannot say whether there’s a case where flags aren’t set.

However, I noticed the new `CefFileDialogFlags` enum is missing a `None` member. I will update the pull request accordingly.

If no flag is set, the `OnFileDialog` method will receive one of the `CefFileDialogMode` values and `CefFileDialogFlags.None`.

I agree that this adds additional maintenance overhead. However, I believe the improved development experience is worth it.

amaitland:
> After looking at the CEF and Chromium source code for a few hours I still wasn’t able to find out where those values are set. 😢 As such, I cannot say whether there’s a case where flags aren’t set.

Quick check and it appears that they're `CEF` specific, It's possible they might always be set for a `Save Dialog`, though I cannot say that definitively yet, I've only had a very brief check. There is only one case that I can find where it's set to anything other than true and I haven't followed that code path through in any detail.

- https://bitbucket.org/chromiumembedded/cef/src/56bbee8efbf072441f5f9824622f940ce94fc57b/libcef/browser/file_dialog_runner.h?at=master&fileviewer=file-view-default#file_dialog_runner.h-26
- https://bitbucket.org/chromiumembedded/cef/src/56bbee8efbf072441f5f9824622f940ce94fc57b/libcef/browser/file_dialog_manager.cc?at=master&fileviewer=file-view-default#file_dialog_manager.cc-195
- https://bitbucket.org/chromiumembedded/cef/src/56bbee8efbf072441f5f9824622f940ce94fc57b/libcef/browser/file_dialog_manager.cc?at=master&fileviewer=file-view-default#file_dialog_manager.cc-283

PR #2282, opened 2018-02-21
Fix for: The added or subtracted value results in an un-representable…

GrabzIt:
… DateTime.

In version 63.0.1 large epochs are causing the error quite frequently:

The added or subtracted value results in an un-representable DateTime.
Parameter name: value
   at System.DateTime.AddTicks(Int64 value)
   at System.DateTime.Add(Double value, Int32 scale)
   at CefSharp.Internals.DateTimeUtils.FromCefTime(Double epoch)
   at CefSharp.ResourceHandlerWrapper.GetCookie(ResourceHandlerWrapper* , CefStructBase<CefCookieTraits>* cookie)
   at CefSharp.ResourceHandlerWrapper.CanGetCookie(ResourceHandlerWrapper* , CefStructBase<CefCookieTraits>* cookie)

If it's possible it would be awesome if this could be released as minor release for version 63.

### DIFF
--- a/CefSharp/Internals/DateTimeUtils.cs
+++ b/CefSharp/Internals/DateTimeUtils.cs
@@ -20,7 +20,14 @@ namespace CefSharp.Internals
         {
             if (epoch > 0)
             {
-                return FirstOfTheFirstNineteenSeventy.AddSeconds(epoch).ToLocalTime();
+                try
+                {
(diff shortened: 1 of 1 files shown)


GrabzIt:
No longer required as @amaitland has fixed the problem.

PR #2303, opened 2018-03-07
Serialize Dictionary type to V8 object

WarWithinMe:
Handle Dictionary type, otherwise it will be treated as class/structs and cause exception.

### DIFF
--- a/CefSharp.Core/Internals/Serialization/V8Serialization.cpp
+++ b/CefSharp.Core/Internals/Serialization/V8Serialization.cpp
@@ -111,6 +111,18 @@ namespace CefSharp
                     }
                     list->SetList(index, subList);
                 }
+                // Serialize dictionary to CefDictionary (key,value pairs)
+                else if (System::Collections::IDictionary::typeid->IsAssignableFrom(type))
+                {
+                    auto subDict = CefDictionaryValue::Create();
+                    auto dict = (System::Collections::IDictionary^) obj;
+                    for each (System::Collections::DictionaryEntry kvp in dict)
+                    {
+                        auto fieldName = StringUtils::ToNative(Convert::ToString(kvp.Key));
+                        SerializeV8SimpleObject(subDict, fieldName, kvp.Value, seen);
+                    }
+                    list->SetDictionary(index, subDict);
+                }
                 // Serialize class/structs to CefDictionary (key,value pairs)
                 else if (!type->IsPrimitive && !type->IsEnum)
                 {
--- a/CefSharp.Example/AsyncBoundObject.cs
+++ b/CefSharp.Example/AsyncBoundObject.cs
@@ -3,6 +3,7 @@
 // Use of this source code is governed by a BSD-style license that can be found in the LICENSE file.
 
 using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Text;
@@ -99,7 +100,8 @@ namespace CefSharp.Example
             return builder.ToString();
         }
 
-        public Dictionary<string, int> MethodReturnsDictionary1() {
+        public Dictionary<string, int> MethodReturnsDictionary1()
+        {
             return new Dictionary<string, int>()
             {
                 {"five", 5},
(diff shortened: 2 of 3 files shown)


amaitland:
Thanks, will need a unit test and probably some changes to the type checking. 

Will review in detail later.

CefSharp.Core/Internals/Serialization/V8Serialization.cpp
```diff
@@ -111,6 +111,18 @@ namespace CefSharp
                     }
                     list->SetList(index, subList);
                 }
+                // Serialize dictionary to CefDictionary (key,value pairs)
+                else if (type->IsGenericType && type->GenericTypeArguments->Length == 2 && type->Name->StartsWith("Dict"))
```
amaitland: Can just check for dictionary using https://msdn.microsoft.com/en-us/library/system.type.isassignablefrom(v=vs.110).aspx

CefSharp.Core/Internals/Serialization/V8Serialization.cpp
```diff
@@ -111,6 +111,18 @@ namespace CefSharp
...
+                {
+                    auto subDict = CefDictionaryValue::Create();
+                    auto dict = (System::Collections::IDictionary^) obj;
+                    for each (System::Collections::DictionaryEntry kvp in dict)
+                    {
+                        auto fieldName = StringUtils::ToNative((String ^)kvp.Key);
```
amaitland: I think using Convert.ToString is probably the go instead of casting

CefSharp.Core/Internals/Serialization/V8Serialization.cpp
WarWithinMe: If the user returns an object of Dictionary with more specific parameters, for example, `Dictionary<string, string>` instead of `Dictionary<string, object>`, `IsAssignableFrom` returns false:

```
typeof(Dictionary<string, object>).IsAssignableFrom(typeof(Dictionary<string, string>)) === false
````

amaitland: A chick for just straight `IDictionary` should be sufficient, not need to limit things to the generic version.

amaitland: You are later casting to `IDitionary`, so the generic checks aren't required. And using `Convert.ToString` should gracefully handle the type conversion of the key .

https://msdn.microsoft.com/en-us/library/system.convert.tostring(v=vs.110).aspx

WarWithinMe: Don't have any experience with managed C++ and get confused with C#.... Will make a commit later

amaitland: It will be something like `type->IsAssignableFrom(System::Collections::IDictionary::typeid)`

WarWithinMe:
Before adding a testcase  in BindingTest.html, I need to know how to run the test first. Is there any ReadMe or Wiki I can refer to?

amaitland:
Open `custom://cefsharp/BindingTest.html` in either the `WinForms` or `WPF` examples.

The default registered object is https://github.com/cefsharp/CefSharp/blob/master/CefSharp.Example/AsyncBoundObject.cs

WarWithinMe:
@amaitland Please review.

WarWithinMe:
The latest force push should have removed all the unintentional modifications.

amaitland:
I'd like to see one of the test methods changed to just a straight `IDictionary` and for consistency braces need to be on new lines, other than that it looks fine 👍

WarWithinMe:
Sure thing.

WarWithinMe:
Hopefully I have ensured consistent bracket styles with the latest commit. (Quite hard, 'cause not having a format tool). If not, please modify them as suited after the PR

amaitland:
Thanks 👍

amaitland:
https://github.com/cefsharp/CefSharp/commit/1542fa3cfcbc95a6ce69130b2ee539763bfbcffc

PR #2311, opened 2018-03-12
Fixed crash in Writable Bitmap Factory when browser is resized

joaompneves:
When resizing the browser CreateOrUpdateBitmap from WritableBitmapFactory throws exception when using an invalid image handle.
Fixed that problem preventing the bitmap update to occur with invalid image handles. The update will eventually occur later with the freshly created handle.

### DIFF
--- a/CefSharp.Wpf/Rendering/WritableBitmapFactory.cs
+++ b/CefSharp.Wpf/Rendering/WritableBitmapFactory.cs
@@ -115,6 +115,11 @@ namespace CefSharp.Wpf.Rendering
             {
                 lock (lockObject)
                 {
+                    if (backBufferHandle.IsInvalid) 
+                    {
+                        return; // bailout, the handle was recreated
+                    }
+
                     if (createNewBitmap)
                     {
                         if (image.Source != null)


Antonyo:
@joaompneves you might want to include additional info to recreate this issue.  

This issue happens while continuously resizing a browser window with a  WritableBitmap back buffer so it's only recreatable in `High DPI`.  It's due to a race condition between releasing the back buffer memory in a worker thread in a resize operation and copying the pixels to the back buffer in the UI Thread from the previous resize operation. Avoid copying pixels if the back buffer is not valid seems a valid solution since the back buffer will be copied again in the latest resize operation with the correct size but I would consider a safer check like [this](https://github.com/Antonyo/CefSharp/commit/3edb5e71366ac2f2ac2208b511283055f2db4d1b)

The exception stack trace is:

```System.ArgumentNullException
  HResult=0x80004003
  Message=Value cannot be null.
Parameter name: sourceBuffer
  Source=PresentationCore
  StackTrace:
   at System.Windows.Media.Imaging.WriteableBitmap.WritePixelsImpl(Int32Rect sourceRect, IntPtr sourceBuffer, Int32 sourceBufferSize, Int32 sourceBufferStride, Int32 destinationX, Int32 destinationY, Boolean backwardsCompat)
   at System.Windows.Media.Imaging.WriteableBitmap.WritePixels(Int32Rect sourceRect, IntPtr sourceBuffer, Int32 sourceBufferSize, Int32 sourceBufferStride, Int32 destinationX, Int32 destinationY)
   at CefSharp.Wpf.Rendering.WritableBitmapFactory.<>c__DisplayClass17_0.<CreateOrUpdateBitmap>b__0() in C:\Work\Projects\Cefsharp\CefSharp.Wpf\Rendering\WritableBitmapFactory.cs:line 146
   at System.Windows.Threading.ExceptionWrapper.InternalRealCall(Delegate callback, Object args, Int32 numArgs)
   at System.Windows.Threading.ExceptionWrapper.TryCatchWhen(Object source, Delegate callback, Object args, Int32 numArgs, Delegate catchHandler)
   at System.Windows.Threading.DispatcherOperation.InvokeImpl()
   at System.Windows.Threading.DispatcherOperation.InvokeInSecurityContext(Object state)
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state)
   at MS.Internal.CulturePreservingExecutionContext.Run(CulturePreservingExecutionContext executionContext, ContextCallback callback, Object state)
   at System.Windows.Threading.DispatcherOperation.Invoke()
   at System.Windows.Threading.Dispatcher.ProcessQueue()
   at System.Windows.Threading.Dispatcher.WndProcHook(IntPtr hwnd, Int32 msg, IntPtr wParam, IntPtr lParam, Boolean& handled)
   at MS.Win32.HwndWrapper.WndProc(IntPtr hwnd, Int32 msg, IntPtr wParam, IntPtr lParam, Boolean& handled)
   at MS.Win32.HwndSubclass.DispatcherCallbackOperation(Object o)
   at System.Windows.Threading.ExceptionWrapper.InternalRealCall(Delegate callback, Object args, Int32 numArgs)
   at System.Windows.Threading.ExceptionWrapper.TryCatchWhen(Object source, Delegate callback, Object args, Int32 numArgs, Delegate catchHandler)
   at System.Windows.Threading.Dispatcher.LegacyInvokeImpl(DispatcherPriority priority, TimeSpan timeout, Delegate method, Object args, Int32 numArgs)
   at MS.Win32.HwndSubclass.SubclassWndProc(IntPtr hwnd, Int32 msg, IntPtr wParam, IntPtr lParam)
```

This PR should be referenced in #2237

It should be also be possible to recreate with an `InteropBitmapFactory` since the pattern is the same but I have not been able so far.

joaompneves:
Looking at the InteropBitmap implementation I think the problem does not occur because the access to the handle is always inside the locked region. So that might also be another way (probably better) to fix this problem, I mean, get rid of the backBufferHandle variable.

Antonyo:
@joaompneves not using the variable is even worst because then it ends up writing the bitmap from a backbuffer with different size. Actually that's what's happening in the InteropBitmap implementation which results in visual artifacts. I will take a look at it this weekend when I have some free time.

amaitland:
Thanks, Checks added in https://github.com/cefsharp/CefSharp/commit/59c6da399cf0a7922bac4278dca4eb0e2951eac2

More refactoring to follow, see #2237 for details.

Issue #2258, opened 2018-02-01
WPF Improve Scrollbar/Scroll Drag Usability


amaitland:
Follow up to #2060

It should be possible to capture the mouse outside of the control to improve usability.

Need to capture mouse and then release the capture.

Possibly `Mouse.PreviewMouseUpOutsideCapturedElement` attached event might be usable here.

 https://msdn.microsoft.com/en-us/library/system.windows.input.mouse.previewmouseupoutsidecapturedelement(v=vs.110).aspx

amaitland:
First attempt at fixing this see commit https://github.com/cefsharp/CefSharp/commit/c843317e338e572fb485b31b6ec3c9c7a517ca53

amaitland:
Changes reverted, seems when leaving the `Window` forces `Capture` to be lost. Appears to be a limitation of `WPF`, can possibly deal with this behaviour and capture scroll when staying within `Window`. To capture when leaving the `Window` might require some `Win32 API` calls.

amaitland:
> Changes reverted, seems when leaving the `Window` forces `Capture` to be lost

For clarity, this behaviour reintroduces #2060, detailed steps to reproduce in https://github.com/cefsharp/CefSharp/pull/2871#issuecomment-524189434

amaitland:
Looking at the `WPF` source it looks like a standard [SetCapture](https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setcapture) with a [ReleaseCapture](https://docs.microsoft.com/en-gb/windows/win32/api/winuser/nf-winuser-releasecapture)

https://referencesource.microsoft.com/#WindowsBase/Shared/MS/Win32/SafeNativeMethodsCLR.cs,486
https://referencesource.microsoft.com/#WindowsBase/Shared/MS/Win32/SafeNativeMethodsCLR.cs,272

So it's actually likely the `WPF Window` has the capture, it's just not forwarding the messages onto our control, probably because it's outside the window. More digging through the source required to verify this.

amaitland:
Thanks to the excellent detective work of @mol  it appears the Capture related issues are only if the machine has a touch screen. Described roghly in https://github.com/dotnet/wpf/issues/1323 (I was having issues way before `1809` was released, the workaround proposed resolves the issues, so assuming it's the same).

[...]

PR #2325, opened 2018-03-21
Captures mouse for left mouse button down events to fix scrolling isssue

jgyo:
Small modification to CefSharp.Wpf project in ChromiumWebBrowser.cs to fix scrolling issue when the mouse leaves the browser control. The modification does the following:

1. Captures the mouse when the left mouse button down event is received.
2. Release the mouse when any mouse button up event is received.
3. Removes browser.GetHost().SendMouseClickEvent call in the OnMouseLeave method which caused a premature loss of the mouse.

With this change, the scrollbar behaves in the same way it behaves in Chrome where leaving the scrollbar either to the left or right does not lose the mouse and scrolling continues normally.

### DIFF
--- a/CefSharp.Wpf/ChromiumWebBrowser.cs
+++ b/CefSharp.Wpf/ChromiumWebBrowser.cs
@@ -2057,6 +2057,15 @@ namespace CefSharp.Wpf
             Focus();
             OnMouseButton(e);
 
+
+            // Capture the mouse only for the left button down event.
+            // Note: This code must follow OnMouseButton().
+            if (e.ChangedButton == MouseButton.Left)
+            {
+                // Capture mouse so browser can see mouse moving outside of the controll
+                Mouse.Capture(this);
+            }
+
             base.OnMouseDown(e);
         }
 
@@ -2067,6 +2076,9 @@ namespace CefSharp.Wpf
         protected override void OnMouseUp(MouseButtonEventArgs e)
         {
             OnMouseButton(e);
+
+            // Release the mouse for all button up events.
+            Mouse.Capture(null);
 
             base.OnMouseUp(e);
         }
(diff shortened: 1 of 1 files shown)


amaitland:
> Small modification to CefSharp.Wpf project in ChromiumWebBrowser.cs to fix scrolling issue when the mouse leaves the browser control.

Issue #2258

> Captures the mouse when the left mouse button down event is received.

This seems questionable, is it wise/safe/advisable to capture the mouse for every left mouse down?

> Release the mouse when any mouse button up event is received.

Releasing for every up, not just the left ones doesn't seem sensible either.

> Removes browser.GetHost().SendMouseClickEvent call in the OnMouseLeave method which caused a premature loss of the mouse.

That is deliberate and a short term fix until #2258 is finally resolved.

Honestly I don't like your solution, and maybe you know more about the inner working of `Mouse.Capture` than I do, and can say it's perfectly safe to do so? Personally I've had a horrible experience with `Mouse.Capture` before that crashed randomly for no reason and now it's being called for **every** left click. To me it makes sense to only capture the mouse on leave, then look at releasing with `Mouse.PreviewMouseUpOutsideCapturedElement` is called (or something similar). 

Short term you can of course subclass `ChromiumWebBrowser` and override `OnMouseLeave/Down/Up`, add the calls yourself, and basically copy paste `OnMouseLeave` and leave out the `SendMouseClickEvent` calls (Haven't tested, though hopefully all the relevant extension methods are accessible).

jgyo:
I haven't had the problems you describe with Mouse.Capture.

The reason for issue #2258 it seems from my testing is that the mouse is being released prematurely without the browser knowing that fact. So when the mouse is over the scroll bar, and it can see the mouse, it thinks that scrolling is still occuring so it moves the scroll bar.

My fix insures that the mouse remains captured during all of the scrolling so the problem described in #2258 does not occur. My fix makes the temporary fix mute.

amaitland:
> I haven't had the problems you describe with Mouse.Capture.

How much testing have you done? See https://github.com/cefsharp/CefSharp/issues/1723#issuecomment-229236482 for details on the problem we've had previously.

`Mouse.Capture` in my opinion is took expensive a call to make for every click, see https://referencesource.microsoft.com/#PresentationCore/Core/CSharp/System/Windows/Input/MouseDevice.cs,342

I haven't stepped into the source to see exactly which code path it goes through.

> My fix insures that the mouse remains captured during all of the scrolling so the problem described in #2258 does not occur. My fix makes the temporary fix mute.

It also captures for every left click and calls releases for every click unless I'm missing something? 

There should be enough extension points to implement this fix in a sub-classed `ChromiumWebBrowser`.

amaitland:
Quick check and you are unable to select a drop down menu item. You can load `custom://cefsharp/TooltipTest.html` for testing in the `CefSharp.Wpf.Example` project.

I know I may have come across strongly, harsh even, it's nothing personal, I just have a grudge against `Mouse.Capture`! 😄

jgyo:
Thanks for checking. I understand.

amaitland:
Give https://github.com/cefsharp/CefSharp/commit/c843317e338e572fb485b31b6ec3c9c7a517ca53 a go

amaitland:
> Give c843317 a go

If you experience any problems with the changes made in https://github.com/cefsharp/CefSharp/commit/c843317e338e572fb485b31b6ec3c9c7a517ca53 then please comment on #2258

mol:
@amaitland turns out the missing `Mouse.Capture(this);` in OnMouseDown is the cause of both this issue with the scroller not moving when moving up and down with the mouse down on the right of the scroller and this one I reported yesterday: https://github.com/cefsharp/CefSharp/issues/2870

If I put back the capture both of those issues are resolved. I'll look closer at these pull requests and linked issues to see if I can find another way to fix those issues.

Issue #2339, opened 2018-03-29
EvaluateScriptAsync (IWebBrowser, String,Object[]) incorrectly handles json string argument


hunkydoryrepair:
In 63.0.2 via NuGet, which passing arguments to a javascript function, strings are expanded.  I have a valid json string, passed in from javascript. Passing it back to ExecuteScriptAsync, any occurance of "\\n" is replaced by "\n", which causes JSON.parse to throw an exception.

I have verified using .charCodeAt and [] between c# and javascript. For example:

in C#:
```
String param = "[\"\\n\"]";
ASSERT(param[2] == '\\');   //  == 92
ctlWebBrowser.ExecuteScriptAsync("HostCallback", {param});
```

in html:
```

<script>
function HostCallback( param )
{
       console.log(param.charCodeAt(2));   //  <--- outputs 10
}

</script>

```

That is certainly unexpected behavior, and I don't necessarily have a good workaround.  I could replace('\n','\\n') on the javascript side, but I'm not sure if there are other characters being substituted.

amaitland:
Looks like your using the `ExecuteScriptAsync` extension method, the source is https://github.com/cefsharp/CefSharp/blob/cefsharp/63/CefSharp/WebBrowserExtensions.cs#L197

The method is only provided for convenience, and isn't totally bullet proof. Your welcome to submit a `PR` with a `Test Case` and improvements.

Otherwise you should use https://github.com/cefsharp/CefSharp/blob/cefsharp/63/CefSharp/WebBrowserExtensions.cs#L209 and build your own string.

hunkydoryrepair:
In CefSharp/WebBrowserExtensions.cs, it appears the strings are just being encapsulated in single quotes and then passed to javascript interpreter, so any escape sequences are going to be unencoded. It is difficult for me to imagine a scenario where this doesn't cause problems. Any string that contains single quotes will break it.  Any strings should be json encoded rather than just put into single quotes.

I don't know the codebase well, I assume there is a specific json encoder that is used throughout CefSharp?

amaitland:
> It is difficult for me to imagine a scenario where this doesn't cause problems.

I don't remember why, you can go through the history and see if there are any clues. From memory this is the first issue I've seen regarding that particular method in a couple of years, so either nobody is using the method, or they're using it for simpler things. 

> Any strings should be json encoded rather than just put into single quotes.

Your welcome to submit a `PR` and some test cases.

[...]

PR #2340, opened 2018-03-29
#2339 Encode string parameters for EvaluateScriptAsync (IWebBrowser, String,Object[]) 

hunkydoryrepair:
Added string encoding and test to ensure strings pass to js and comes back unchanged.
Test can easily be modified with other string features


amaitland:
Resolves #2339

PR #2349, opened 2018-04-09
 Fix non-thread-safety of TrySetResultAsync

shkdee:
The extension method `TaskExtensions.TrySetResultAsync` has non-deterministic results when called twice on the same instance, even on the same thread: it creates one task for each call, which may not run in their creation order.
This PR adds a wrapper class around `TaskCompletionSource` (`AsyncTaskCompletionSource`), that takes care of calling `TrySetResult` in an async fashion while ensuring that only the first call is executed. I would really recommend using it instead of `TaskCompletionSource` each time `TaskExtensions.TrySetResultAsync` is used (even if it is called only once in any logical workflow), because this non-determinism can cause very subtle and hard-to-track bugs.

This PR also updates the test case that could be violated with the previous behavior (but that was hard to trigger with only one test).

### DIFF
--- a/CefSharp/Internals/AsyncTaskCompletionSource.cs
+++ b/CefSharp/Internals/AsyncTaskCompletionSource.cs
@@ -1,0 +1,35 @@
+﻿using System;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace CefSharp
+{
+    /// <summary>
+    /// Wraps <see cref="TaskCompletionSource{T}"/> by providing a way to set its result in an async fashion.
+    /// This prevents the Task Continuation being executed sync on the same thread, which may be required when
+    /// continuations must not run on CEF UI threads.
+    /// </summary>
+    /// <typeparam name="T"></typeparam>
+    public class AsyncTaskCompletionSource<T>
+    {
+        private readonly TaskCompletionSource<T> tcs = new TaskCompletionSource<T>();
+        private int resultSet = 0;
+
+        /// <summary>
+        /// Gets the <see cref="Task{TResult}"/> created by this <see cref="AsyncTaskCompletionSource{T}"/>
+        /// </summary>
+        /// <seealso cref="TaskCompletionSource{TResult}.Task"/>
+        public Task<T> Task => tcs.Task;
+
+        /// <summary>
+        /// Set the TaskCompletionSource in an async fashion. This prevents the Task Continuation being executed sync on the same thread
+        /// This is required otherwise contintinuations will happen on CEF UI threads
+        /// </summary>
+        /// <param name="result">result</param>
+        public void TrySetResultAsync(T result)
+        {
+            if (Interlocked.Exchange(ref resultSet, 1) == 0)
+                System.Threading.Tasks.Task.Factory.StartNew(() => tcs.TrySetResult(result), CancellationToken.None, TaskCreationOptions.None, TaskScheduler.Default);
+        }
+    }
+}
--- a/CefSharp/Internals/AsyncTaskCompletionSource.cs
+++ b/CefSharp/Internals/AsyncTaskCompletionSource.cs
@@ -1,4 +1,8 @@
-﻿using System;
+﻿// Copyright © 2010-2017 The CefSharp Authors. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style license that can be found in the LICENSE file.
+
+using System;
 using System.Threading;
 using System.Threading.Tasks;
 
@@ -22,14 +26,15 @@ namespace CefSharp
         public Task<T> Task => tcs.Task;
 
         /// <summary>
-        /// Set the TaskCompletionSource in an async fashion. This prevents the Task Continuation being executed sync on the same thread
-        /// This is required otherwise contintinuations will happen on CEF UI threads
+        /// Set the TaskCompletionSource in an async fashion. This prevents the Task Continuation being executed
+        /// sync on the same thread. This is required otherwise contintinuations will happen on CEF UI threads.
         /// </summary>
         /// <param name="result">result</param>
         public void TrySetResultAsync(T result)
         {
             if (Interlocked.Exchange(ref resultSet, 1) == 0)
-                System.Threading.Tasks.Task.Factory.StartNew(() => tcs.TrySetResult(result), CancellationToken.None, TaskCreationOptions.None, TaskScheduler.Default);
+                System.Threading.Tasks.Task.Factory.StartNew(() => tcs.TrySetResult(result),
+                    CancellationToken.None, TaskCreationOptions.None, TaskScheduler.Default);
         }
     }
 }
--- a/CefSharp/Internals/TaskExtensions.cs
+++ b/CefSharp/Internals/TaskExtensions.cs
@@ -89,8 +89,14 @@ namespace CefSharp.Internals
         }
 
         /// <summary>
+        /// <para>
         /// Set the TaskCompletionSource in an async fashion. This prevents the Task Continuation being executed sync on the same thread
-        /// This is required otherwise contintinuations will happen on CEF UI threads
+        /// This is required otherwise contintinuations will happen on CEF UI threads.
+        /// </para>
+        /// <para>
+        /// USAGE WARNING: if this method is called twice on the same task completion source, the result it not deterministic.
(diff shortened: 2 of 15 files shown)


amaitland:
Ideally we'd update to .Net 4.6 and remove TrySetResultAsync. 

Other than the single test case what real world use case are you experiencing an issue with? Which CEF callback is calling TrySetResultAsync twice?

CefSharp/Internals/TaskExtensions.cs
```diff
@@ -89,8 +89,14 @@ public static TaskCompletionSource<TResult> WithTimeout<TResult>(this TaskComple
         }
 
         /// <summary>
+        /// <para>
```
amaitland: If we go with your proposal, can this method be deleted?

CefSharp/Internals/AsyncTaskCompletionSource.cs
```diff
@@ -0,0 +1,40 @@
...
+        public Task<T> Task => tcs.Task;
+
+        /// <summary>
+        /// Set the TaskCompletionSource in an async fashion. This prevents the Task Continuation being executed
+        /// sync on the same thread. This is required otherwise contintinuations will happen on CEF UI threads.
+        /// </summary>
+        /// <param name="result">result</param>
+        public void TrySetResultAsync(T result)
+        {
+            if (Interlocked.Exchange(ref resultSet, 1) == 0)
```
amaitland: Code consistently, braces required even for single line if block.

CefSharp/Internals/AsyncTaskCompletionSource.cs
```diff
@@ -0,0 +1,40 @@
...
+
+namespace CefSharp
+{
+    /// <summary>
+    /// Wraps <see cref="TaskCompletionSource{T}"/> by providing a way to set its result in an async fashion.
+    /// This prevents the Task Continuation being executed sync on the same thread, which may be required when
+    /// continuations must not run on CEF UI threads.
+    /// </summary>
+    /// <typeparam name="T"></typeparam>
+    public class AsyncTaskCompletionSource<T>
```
amaitland: This can probably be made internal.

CefSharp/Internals/TaskExtensions.cs
shkdee: I guess it could, it's just still used in a few places in the examples.

CefSharp/Internals/AsyncTaskCompletionSource.cs
```diff
@@ -0,0 +1,40 @@
...
+        private int resultSet = 0;
+
+        /// <summary>
+        /// Gets the <see cref="Task{TResult}"/> created by this <see cref="AsyncTaskCompletionSource{T}"/>
+        /// </summary>
+        /// <seealso cref="TaskCompletionSource{TResult}.Task"/>
+        public Task<T> Task => tcs.Task;
+
+        /// <summary>
+        /// Set the TaskCompletionSource in an async fashion. This prevents the Task Continuation being executed
```
amaitland: This needs rewording. I had to read it three times, your new sentence about sync on same thread could be clearer. Should probably also link to a relevant resource, I should have done that previously.

CefSharp/Internals/AsyncTaskCompletionSource.cs
amaitland: Last I checked IntetnalsVisibleTo was working for all the c# projects, just doesn't work with the VC++ ones, so we do have to leave it public if it's going to be used in C++. 

https://github.com/cefsharp/CefSharp/blob/master/CefSharp/Properties/AssemblyInfo.cs#L26

CefSharp/Internals/TaskExtensions.cs
amaitland: We can look at upgrading the example projects to .Net 4.6 if required. 

Only the core projects need to remain as .Net 4.5.2

shkdee:
Well as you wish, that would be a good solution too.

I found that bug when using `AsyncExtensions.SetCookieAsync(this ICookieManager cookieManager, string url, Cookie cookie);` that would sometimes randomly return false, due to this bug (it uses `TaskSetCookieCallback` internally, which is one of the callback implementations modified by this PR). I did not try to trigger the bug for the other ones but in any case, the method `TaskExtensions.TrySetResultAsync` is flawed.

CefSharp/Internals/TaskExtensions.cs
shkdee: I'll let you do that then.
I also just noticed it was used in two places in the `CefSharp.Core` project, and I'm much less familiar with C++/CLI to handle that case.

CefSharp/Internals/AsyncTaskCompletionSource.cs
shkdee: Oh sorry, I missed the InternalsVisibleTo. Well then it's ok to make it internal except if we also want to replace the two usages of TaskExtensions::TrySetResultAsync which are in the C++ project.

amaitland: Ok no worries, will have to remain public.

CefSharp/Internals/AsyncTaskCompletionSource.cs
shkdee: By "a relevant resource", you mean regarding executing code on CEF UI threads ? So something like the general usage guide of the wiki ?

CefSharp/Internals/AsyncTaskCompletionSource.cs
shkdee: Fixed.

amaitland:
Upgrading to .Net 4.6 currently isn't an option for the core projects. 

It looks like there are only a couple of classes that have the potential for this bug. Looks like only the callbacks have the potential to make multiple calls. I guess the question is should we limit the scope of this PR to only the relevant classes? 

Will have a think about it, comments welcome.

shkdee:
Well, for that PR only, it would be easier and much less breaking to limit it to relevant classes.

But in the long term, for the CefSharp project, my opinion on the matter is that the current method `TaskExtensions.TrySetResultAsync` is a bit dangerous : it does not behave like the .Net framework method `TaskCompletionSource<>.TrySetResult` it intends to replace (because of this bug), so it's misleading and error-prone (it's really easy to call it twice if you are not very careful, because that would not bother you for the origin method). As such, it should be removed/obsoleted as much as possible.

amaitland:
TrySetResultAsync is only intended to be used within the context of CefSharp it's self, so the scope for error in usage is very small. I think upgrading the example projects to .Net 4.6 and removing is usage from those examples makes a lot of sense, we shouldn't be encouraging others to use the same method. 

If TrySetResultAsync does remain, then it should be renamed for clarity, it's sole purpose is to ensure the continuation happens on the thread pool and not on the CEF UI thread.

amaitland:
I feel this needs more work that I have time for just at the moment, to resolve the issue with the callbacks I've applied a very simple bool check, see commit https://github.com/cefsharp/CefSharp/commit/43c1ff3ed0fe418763f0f07c3124711f0657fa5e for details.

The complete and dispose methods should always be called on the same thread (dispose is slightly different to the standard .net dispose in this context as it's being called from unmanged code when the parent ref pointer goes out of scope). This should resolve the short term issue in the next release and I'll come back to this when I have some time.

amaitland:
I'll resolve the merge conflicts when it comes time to merge this.

CefSharp/Internals/AsyncTaskCompletionSource.cs
JanEggers: guys why dont u use the default TaskCompletionSource and create it with TaskCreationOptions.RunContinuationsAsync ?

amaitland: > guys why dont u use the default TaskCompletionSource and create it with TaskCreationOptions.RunContinuationsAsync ?

TaskCreationOptions.RunContinuationsAsync was added in .Net 4.6

We are still supporting .Net 4.5.2 (no plans to change this)

One day I'll get around to updating the .Net Core builds to use RunContinuationsAsync

amaitland:
#4482 updates to `.Net 4.6.2` and removes `TrySetResultAsync`.

PR #2372, opened 2018-04-25
V8Serialization - Replace usage of Stack with HashSet

### LINKED ISSUE #2370: SerializeV8SimpleObject performance question
Looking at the current master (9eb19756ac8791f3bc716629d9a01a0f8d6c5990):

https://github.com/cefsharp/CefSharp/blob/9eb19756ac8791f3bc716629d9a01a0f8d6c5990/CefSharp.Core/Internals/Serialization/V8Serialization.cpp#L28-L36

Here a `Stack` is used to (presumably) prevent infinite loops when object graphs contain cycles. via the `seen` parameter.

The only operations called on this type are `Push` and `Contains`.

A `HashSet` would have better performance here, swapping O(N) for O(1) on the `Contains` test.

drewnoakes:
Fixes #2370.

Also rename the local from `seen` to `ancestors` as it is not the set of all seen objects, just those in the chain from current to root.

Added a brief comment to this effect.

In issue 2370 I missed the call to `Pop` and thought this collection had _every_ seen object in it which could be quite bad for performance when calling `Contains`. Given it actually only contains ancestors I doubt the perf gain here is considerable, and I confess that I haven't benchmarked.

I almost didn't file this PR. Feel free to close it. I won't mind.

### DIFF
--- a/CefSharp.Core/Internals/Serialization/V8Serialization.cpp
+++ b/CefSharp.Core/Internals/Serialization/V8Serialization.cpp
@@ -145,7 +145,7 @@ namespace CefSharp
                         SerializeV8SimpleObject(subDict, propertyName, propertyValue, seen);
                     }
                     list->SetDictionary(index, subDict);
-                } 
+                }
                 else
                 {
                     throw gcnew NotSupportedException("Unable to serialize Type");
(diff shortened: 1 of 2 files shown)


amaitland:
Performance wise it's hard to say without testing. Possibly with a small number of objects in the `Stack` the reference comparison might be quicker than the hash computation. 

I think the readability improvement make this worth merging 👍  Will wait until after `65` has been branched.

PR #2373, opened 2018-04-25
Call OnContextCreated and OnContextReleased for each frame

sukhovy:
Call `IRenderProcessMessageHandler.OnContextCreated` and `IRenderProcessMessageHandler.OnContextReleased` for each frame. It allows to inject JS code for each frame


amaitland:
Thanks, this will need a bit of work before it can be merged. Upgrading users will now experience a very subtle change in behavior. There's also more code changes to the `SetCanExecuteJavascriptOnMainFrame` calls in `ClientAdapter.cpp` will need to be set only on main frame. The examples will also need to be updated, including some additional comments.

This `PR` will need some notes and potentially an example for `Upgrading` users.

Until this is sorted, did you try execute your scripts in `FrameLoadStart`? I don't usually recommend this approach though might be viable for your usage scenario.

CefSharp/Handler/IRenderProcessMessageHandler.cs
```diff
@@ -14,7 +14,7 @@ public interface IRenderProcessMessageHandler
         /// OnContextCreated is called in the Render process immediately after a CefV8Context is created.
         /// An IPC message is immediately sent to notify the context has been created
         /// (should be safe to execute javascript). If the page has no javascript then no V8Context will be created
-        /// and as a result this method will not be called. Currently only called for the Main frame <see cref="IFrame.IsMain"/>
```
amaitland: Will to mention this will be called for every `v8context` and to check `IFrame.IsMain` to determine if the it's the main context.

CefSharp/Handler/IRenderProcessMessageHandler.cs
```diff
@@ -25,7 +25,7 @@ public interface IRenderProcessMessageHandler
         /// OnContextReleased is called in the Render process immediately before the CefV8Context is released.
         /// An IPC message is immediately sent to notify the context has been released
         /// (cannot execute javascript this point). If the page had no javascript then the context would not have been created
-        /// and as a result this method will not be called. Currently only called for the Main frame <see cref="IFrame.IsMain"/>
+        /// and as a result this method will not be called.
```
amaitland: Similar update here, called per frame

CefSharp.Example/RenderProcessMessageHandler.cs
```diff
@@ -17,9 +17,12 @@ void IRenderProcessMessageHandler.OnFocusedNodeChanged(IWebBrowser browserContro
 
         void IRenderProcessMessageHandler.OnContextCreated(IWebBrowser browserControl, IBrowser browser, IFrame frame)
         {
-            const string script = "document.addEventListener('DOMContentLoaded', function(){ alert('DomLoaded'); });";
+            if (frame.IsMain)
```
amaitland: I think we should probably add a comment here regarding multiple calls to this method, as not everyone reads the xml doc.

AgapovAlexsey:
it be merged?

sukhovy:
@amaitland In our product, we implemented the necessary functionality without these changes. So now they are no longer needed for us. Therefore, this request can be closed or accepted. Maybe someone else needs this feature.

amaitland:
Follow up commit https://github.com/cefsharp/CefSharp/commit/19a56bd7f92cb4b2ddfd93060f7a9a0cdcab5975

Issue #2359, opened 2018-04-13
CefSharp.BrowserSubprocess.exe dangling


joaompneves:
- **What version of the product are you using?**
    - 65 but happens in previous versions also at least since 57

- **What architecture x86 or x64?**
    - Happens in both architecures

- **On what operating system?**
    - Win10

- **Are you using `WinForms`, `WPF` or `OffScreen`?**
    - WPF

- **What steps will reproduce the problem?**
    - I don't know how to reproduce but it happens frequently

- **What is the expected output? What do you see instead?**
   - CefSharp.BrowserSubprocess.exe does not close and keeps around consuming 100% of 1 CPU core after the main application process closed 

- **Please provide any additional information below.**
    - The stack trace of the process main thread consuming the CPU:
0, libcef.dll!IsSandboxedProcess+0x19a0bac
1, 0xfbe730088e478a
2, 0xc9f26501c193b3a
3, 0x4107d6671bc664d4
4, 0x4107d6671c193b3a
5, 0x4107d6671bc664d4
6, libcef.dll!IsSandboxedProcess+0x19a0bac (No unwind info)
7, libcef.dll!cef_zip_reader_create+0x1042499 (No unwind info)
8, libcef.dll!cef_zip_reader_create+0x104201a (No unwind info)
9, libcef.dll!RelaunchChromeBrowserWithNewCommandLineIfNeeded+0x17869 (No unwind info)
10, libcef.dll!GetHandleVerifier+0x408c9 (No unwind info)
11, libcef.dll!cef_zip_reader_create+0x10458b3 (No unwind info)
12, libcef.dll!IsSandboxedProcess+0x1448bb3 (No unwind info)
13, libcef.dll!GetHandleVerifier+0x408c9 (No unwind info)
14, libcef.dll!GetHandleVerifier+0x5b023 (No unwind info)
15, libcef.dll!GetHandleVerifier+0x9206 (No unwind info)
16, libcef.dll!GetHandleVerifier+0x9547 (No unwind info)
17, libcef.dll!GetHandleVerifier+0x965e (No unwind info)
18, libcef.dll!GetHandleVerifier+0x5cea7 (No unwind info)
19, libcef.dll!GetHandleVerifier+0x8f4f (No unwind info)
20, libcef.dll!cef_time_delta+0x997ff (No unwind info)
21, libcef.dll!IsSandboxedProcess+0xfc1383 (No unwind info)
[...]

PR #2375, opened 2018-04-26
Close CefSharp.BrowserSubprocess.exe if parent process exits

joaompneves:
Following the issue #2359 here is the "solution".

I tested this solution making the following changes in CefSharp.Wpf.Example\Views\BrowserTabView.xaml.cs:

```
browser.RenderProcessMessageHandler = new RenderProcessMessageHandler();

browser.FrameLoadStart += (sender, args) => {
    browser.Dispatcher.BeginInvoke((Action) (() => {
        Thread.Sleep(10);
        Environment.Exit(0);
    }));
};
```

### DIFF
--- a/CefSharp.BrowserSubprocess/Program.cs
+++ b/CefSharp.BrowserSubprocess/Program.cs
@@ -21,7 +21,7 @@ namespace CefSharp.BrowserSubprocess
             var type = args.GetArgumentValue(CefSharpArguments.SubProcessTypeArgument);
             var parentProcessId = int.Parse(args.GetArgumentValue(CefSharpArguments.HostProcessIdArgument));
 
-            Task.Run(() => AwaitParentProcessExit(parentProcessId));
+            Task.Factory.StartNew(() => AwaitParentProcessExit(parentProcessId), TaskCreationOptions.LongRunning);
 
             //Use our custom subProcess provides features like EvaluateJavascript
             if (type == "renderer")
@@ -53,9 +53,8 @@ namespace CefSharp.BrowserSubprocess
             }
             catch (Exception e) 
             {
-                //main process probably died already, bailout
+                //main process probably died already
                 Debug.WriteLine(e);
-                return;
             }
 
             Task.Delay(1000); //wait a bit before exiting
--- a/CefSharp.BrowserSubprocess/Program.cs
+++ b/CefSharp.BrowserSubprocess/Program.cs
@@ -3,9 +3,8 @@
 // Use of this source code is governed by a BSD-style license that can be found in the LICENSE file.
 
 using System;
-using System.Collections.Generic;
 using System.Diagnostics;
-using System.Linq;
+using System.Threading.Tasks;
 using CefSharp.Internals;
 
 namespace CefSharp.BrowserSubprocess
@@ -19,16 +18,16 @@ namespace CefSharp.BrowserSubprocess
             SubProcess.EnableHighDPISupport();
 
             int result;
+            var type = args.GetArgumentValue(CefSharpArguments.SubProcessTypeArgument);
+            var parentProcessId = int.Parse(args.GetArgumentValue(CefSharpArguments.HostProcessIdArgument));
 
-            const string typePrefix = "--type=";
-            var typeArgument = args.SingleOrDefault(arg => arg.StartsWith(typePrefix));
-            var type = typeArgument.Substring(typePrefix.Length);
+            Task.Run(() => AwaitParentProcessExit(parentProcessId));
 
             //Use our custom subProcess provides features like EvaluateJavascript
             if (type == "renderer")
             {
                 var wcfEnabled = args.HasArgument(CefSharpArguments.WcfEnabledArgument);
-                var subProcess = wcfEnabled ? new WcfEnabledSubProcess(args) : new SubProcess(args);
+                var subProcess = wcfEnabled ? new WcfEnabledSubProcess(parentProcessId, args) : new SubProcess(args);
 
                 using (subProcess)
                 {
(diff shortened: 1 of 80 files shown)


CefSharp.BrowserSubprocess/Program.cs
```diff
@@ -19,16 +18,16 @@ public static int Main(string[] args)
             SubProcess.EnableHighDPISupport();
 
             int result;
+            var type = args.GetArgumentValue(CefSharpArguments.SubProcessTypeArgument);
+            var parentProcessId = int.Parse(args.GetArgumentValue(CefSharpArguments.HostProcessIdArgument));
 
-            const string typePrefix = "--type=";
-            var typeArgument = args.SingleOrDefault(arg => arg.StartsWith(typePrefix));
-            var type = typeArgument.Substring(typePrefix.Length);
+            Task.Run(() => AwaitParentProcessExit(parentProcessId));
```
amaitland: Whilst the `ThreadPool` won't be used in the render process, it just doesn't seem appropriate to use a vanilla `Task.Run`, should probably add `TaskCreationOptions.LongRunning`, from memory it should spawn a new `Thread` and a quick check on `stackoverflow` suggests it will.

https://stackoverflow.com/a/10298761

CefSharp.BrowserSubprocess/Program.cs
```diff
@@ -41,8 +40,29 @@ public static int Main(string[] args)
...
+            try 
+            {
+                var parentProcess = Process.GetProcessById(parentProcessId);
+                parentProcess.WaitForExit();
+            }
+            catch (Exception e) 
+            {
+                //main process probably died already, bailout
+                Debug.WriteLine(e);
+                return;
```
amaitland: What's the purpose of this return?

CefSharp.BrowserSubprocess/Program.cs
```diff
@@ -41,8 +40,29 @@ public static int Main(string[] args)
...
+            }
+
+            Task.Delay(1000); //wait a bit before exiting
+
+            Debug.WriteLine("BrowserSubprocess shutting down forcibly.");
```
amaitland: Should probably log to the native `CEF Logging`, trivial to add an wrapper method.

CefSharp/CefCustomScheme.cs
```diff
@@ -119,13 +119,12 @@ public CefCustomScheme()
         /// <returns>list of scheme objects</returns>
         public static List<CefCustomScheme> ParseCommandLineArguments(IEnumerable<string> args)
         {
-            var schemes = args.FirstOrDefault(a => a.StartsWith(CefSharpArguments.CustomSchemeArgument));
+            var schemes = args.GetArgumentValue(CefSharpArguments.CustomSchemeArgument);
```
amaitland: Have you tested these changes?

CefSharp.BrowserSubprocess/Program.cs
joaompneves: I though about adding that, but since this process is running mostly unmanaged code and is not using the ThreadPool, I skipped it. But its a small change, I will do it.

CefSharp.BrowserSubprocess/Program.cs
joaompneves: I assumed that the process would die by it self, but I'll remove it... its safer.

CefSharp.BrowserSubprocess/Program.cs
joaompneves: I just copied the same pattern that was already there.

CefSharp/CefCustomScheme.cs
joaompneves: Yes... and I even spotted an existing "bug". Some of the parameters sent to the process are duplicated, that's why the FirstOrDefault works but not not SIngleOrDefault.

amaitland: Which paramaters are duplicated?

joaompneves:
Those defined in CefSharpArguments

   João Neves

CefSharp/CefCustomScheme.cs
joaompneves: Those defined in CefSharpArguments

amaitland: Interesting, thanks

amaitland:
I don't have time to properly review this before the `65.0.0-pre01` release which will be out shortly.

I've manually cherry picked some pieces that are included in commit https://github.com/cefsharp/CefSharp/commit/f2fcbf0add188c9c5bd411c47ba798b608a105c7

The process id will always be passed through and I've updated the command line parsing. This should give anyone wishing to try this feature out the ability to create their own `BrowserSubprocess` with minimal effort and include the `Task` that waits for the parent to exit.

When it comes time to merge this I'll resolve the conflicts.

CefSharp.BrowserSubprocess/Program.cs
```diff
@@ -19,16 +18,16 @@ public static int Main(string[] args)
             SubProcess.EnableHighDPISupport();
 
             int result;
+            var type = args.GetArgumentValue(CefSharpArguments.SubProcessTypeArgument);
+            var parentProcessId = int.Parse(args.GetArgumentValue(CefSharpArguments.HostProcessIdArgument));
 
-            const string typePrefix = "--type=";
-            var typeArgument = args.SingleOrDefault(arg => arg.StartsWith(typePrefix));
-            var type = typeArgument.Substring(typePrefix.Length);
+            Task.Factory.StartNew(() => AwaitParentProcessExit(parentProcessId), TaskCreationOptions.LongRunning);
```
kronic: Does not take a thread
 ```csharp
var parentProcess = Process.GetProcessById(parentProcessId)
if(parentProcess != null)
{
   parentProcess.EnableRaisingEvents = true;
   parentProcess.Exited += (o,a) => 
   {
     Debug.WriteLine("BrowserSubprocess shutting down forcibly.");
     Environment.Exit(0);
   }
}

CefSharp.BrowserSubprocess/Program.cs
```diff
@@ -41,8 +40,28 @@ public static int Main(string[] args)
...
+                var parentProcess = Process.GetProcessById(parentProcessId);
+                parentProcess.WaitForExit();
+            }
+            catch (Exception e) 
+            {
+                //main process probably died already
+                Debug.WriteLine(e);
+            }
+
+            Task.Delay(1000); //wait a bit before exiting
```
kronic:
```csharp
await Task.Delay(1000);

PR #2388, opened 2018-05-10
Fixed issue with separate window using WPF.ChromiumWebBrowser

krupineg:
Application.Exit is dispatched
Otherwise if we try to create browser and put it into separate window
with separate dispatcher thread we have unhandled exception in static
constructor

### DIFF
--- a/CefSharp.Wpf/ChromiumWebBrowser.cs
+++ b/CefSharp.Wpf/ChromiumWebBrowser.cs
@@ -400,7 +400,10 @@ namespace CefSharp.Wpf
 
                 if (app != null)
                 {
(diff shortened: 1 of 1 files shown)


amaitland:
Thanks for the contribution, firstly all `PR's` need to be targeted against `master`. 

I think this use case is better handled in your own code, simply set `CefSharpSettings.ShutdownOnExit = false` before creating your first `ChromiumWebBrowser` instance. Then handle calling `Cef.Shutdown` yourself.

Remember that `Cef.Initialize` and `Cef.Shutdown` need to be called on the same `Thread`.

PR #2397, opened 2018-05-22
Bound objects null

mbragg12:
Null checking to reduce chance of hard crash in OnProcessMessageReceived


amaitland:
When exactly are you seeing it crash? If you referring to the missed changes that I didn't merge into the `cefsharp/65` branch then this wouldn't prevent that crash, the additional params were missing. Honestly I think it should crash as it's a development problem.

amaitland:
Actually I stand corrected, I would prevent the crash. I still think the crash is better than logging a message as this is a dev issue not a prod issue.

mbragg12:
Fair enough.  I had the code stubbed in when I was trying to figure what was going on, so figured I would offer it up.
Your call, I can merge it or just kill the PR.

PR #2416, opened 2018-06-12
Release window event handlers on Dispose

henrihs:
Make sure event handlers related to `Window`, e.g. `WindowStateChanged`, is released during Dispose.

### DIFF
--- a/CefSharp.Wpf/ChromiumWebBrowser.cs
+++ b/CefSharp.Wpf/ChromiumWebBrowser.cs
@@ -41,6 +41,10 @@ namespace CefSharp.Wpf
         /// </summary>
         private HwndSource source;
         /// <summary>
+        /// The source window bound to window related EventHandlers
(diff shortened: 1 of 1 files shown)


CefSharp.Wpf/ChromiumWebBrowser.cs
```diff
@@ -556,6 +560,12 @@ protected virtual void Dispose(bool isDisposing)
                     }
 
                     PresentationSource.RemoveSourceChangedHandler(this, PresentationSourceChangedHandler);
+                    // Release window event listeners if SourceChanged event wasn't received before disposal
+                    if (sourceWindow != null)
```
amaitland: Should this be set to `null` here as well?

henrihs: Should I squash these commits?

PR #2424, opened 2018-06-20
Mouse capture bug

AngryShaDoW:
If parent control have some logic with mouse capturing and if user will
pass over browser this logic will be broken.

### DIFF
--- a/CefSharp.Wpf/ChromiumWebBrowser.cs
+++ b/CefSharp.Wpf/ChromiumWebBrowser.cs
@@ -57,13 +57,6 @@ namespace CefSharp.Wpf
         /// </summary>
         private ManagedCefBrowserAdapter managedCefBrowserAdapter;
         /// <summary>
-        /// Track the state of the Mouse.Capture set in OnMouseLeave, when the user releases
-        /// the left button it's important we release the capture. It's important we track the mouse
-        /// capture state ourselves and not just to a Mouse.Captured == this check, as we use
-        /// Mouse.Capture for the popup that hosts dropdown menus
-        /// </summary>
-        private bool mouseCapturedInOnMouseLeave;
-        /// <summary>
         /// The ignore URI change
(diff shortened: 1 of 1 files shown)


amaitland:
Thanks for creating a PR, much easier to discuss. :+1:

Which of the built in .Net controls are you seeing this problem with? 

Are you sure it's wise to capture the mouse on every left click? That doesn't sound sensible to me.

AngryShaDoW:
Hi.
>Which of the built in .Net controls are you seeing this problem with?

We implement sticky windows, that using wndProc, and if at window resize user pass mouse over browser then mouse capture will been reset and we haven't get messages

>Are you sure it's wise to capture the mouse on every left click? That doesn't sound sensible to me.

If you think so, we can make a compromise decision: Set mouse capture on mouse leave, but only if mouseDown have been raised by current control. But first could you to give your arguments that the current decision is not reasonable.

Thanks.

amaitland:
> We implement sticky windows, that using wndProc, and if at window resize user pass mouse over browser then mouse capture will been reset and we haven't get messages

This sounds more like an edge case and maybe better handled by you sub-classing `ChromiumWebBrowser`. Implementing an option to disable the current implementation would be acceptable.

> If you think so, we can make a compromise decision:

Without seeing the change I cannot comment.

> But first could you to give your arguments that the current decision is not reasonable.

I've seen problems with `Mouse.Capture` before and think it should be used sparingly, see #1723 for a past issue. At this point in time the burden is on you to convince me your proposed changes are technically sound and I'm not convinced they are. If you have a `MSDN` link or relevant technical article then I'd be happy to read it when I have time. 

The browser only needs to capture the mouse when it leaves, and the left button is down. Unnecessary calls should be avoided.

amaitland:
Any follow up on this?

amaitland:
Closing due to lack of feedback.

PR #2428, opened 2018-06-24
Fix possible null reference of RenderHandler

DamirAinullin:
Fix possible null reference of RenderHandler in second part of condition.

### DIFF
--- a/CefSharp.Wpf/ChromiumWebBrowser.cs
+++ b/CefSharp.Wpf/ChromiumWebBrowser.cs
@@ -1575,24 +1575,22 @@ namespace CefSharp.Wpf
                         browser.GetHost().NotifyScreenInfoChanged();
                     }
 
-                    //Ignore this for custom bitmap factories
-                    if (RenderHandler != null)
+                    //Ignore this for custom bitmap factories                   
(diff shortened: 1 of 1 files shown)


amaitland:
Please include a description and simplify the title. 

Looking at the code again and could probably be rewritten to use **is** checks. Happy to go with this for now :+1:

DamirAinullin:
@amaitland Done.

CefSharp.Wpf/ChromiumWebBrowser.cs
```diff
@@ -1576,18 +1576,21 @@ private void PresentationSourceChangedHandler(object sender, SourceChangedEventA
                     }
 
                     //Ignore this for custom bitmap factories
-                    if (RenderHandler != null && RenderHandler.GetType() == typeof(WritableBitmapRenderHandler) || RenderHandler.GetType() == typeof(InteropBitmapRenderHandler))
+                    if (RenderHandler != null)
```
amaitland: This is now redundant and can be removed

DamirAinullin: removed

Issue #2440, opened 2018-07-05
Cannot generate dump file with crash_reporter.cfg


Michael-Tajmajer-Emrsn:
What version of the product are you using?

65.0.0-pre02

On what operating system?

Windows 10 (1607)

What steps will reproduce the problem?

Build CefSharp.WinForms.Example, ensure that crash_reporter.cfg is in the CefSharp.WinForms.Example\bin\x86\Release folder.

[Config]
ProductName=CefSharp
ProductVersion=63.0.0
AppName=CefSharp
ExternalHandler=CefSharp.BrowserSubprocess.exe

Run CefSharp.WinForms.Example.exe, naviagate to chrome://crash, or chrome://inducebrowsercrashforrealz and no dmp file is created

Please provide any additional information below.

debug.log:

[0705/110851.221:ERROR:registration_protocol_win.cc(84)] TransactNamedPipe: The pipe has been ended. (0x6D)
[0705/110851.229:INFO:crash_reporting.cc(215)] Crash reporting enabled for process: browser
[0705/110851.337:ERROR:registration_protocol_win.cc(56)] CreateFile: The system cannot find the file specified. (0x2)
[0705/110851.338:INFO:crash_reporting.cc(215)] Crash reporting enabled for process: gpu-process
[0705/110851.601:ERROR:gpu_process_transport_factory.cc(1029)] Lost UI shared context.
[0705/110851.632:ERROR:registration_protocol_win.cc(56)] CreateFile: The system cannot find the file specified. (0x2)
[0705/110851.637:INFO:crash_reporting.cc(215)] Crash reporting enabled for process: renderer
[0705/110851.737:ERROR:command_buffer_proxy_impl.cc(114)] ContextResult::kFatalFailure: Shared memory handle is not valid
[0705/110856.366:ERROR:render_frame_impl.cc(896)] Intentionally crashing (with null pointer dereference) because user navigated to chrome://crash/
[0705/110859.369:ERROR:registration_protocol_win.cc(56)] CreateFile: The system cannot find the file specified. (0x2)
[0705/110859.373:INFO:crash_reporting.cc(215)] Crash reporting enabled for process: renderer
[0705/111225.515:ERROR:registration_protocol_win.cc(84)] TransactNamedPipe: The pipe has been ended. (0x6D)
[0705/111225.523:INFO:crash_reporting.cc(215)] Crash reporting enabled for process: browser
[...]

PR #2449, opened 2018-07-10
Bug fix: handle type=crashpad-handler

perlun:
I was seeing instances of the subprocess being spawned this way when running the CefSharp.Wpf.Example project.

```
Application: CefSharp.BrowserSubprocess.exe
Framework Version: v4.0.30319
Description: The process was terminated due to an unhandled exception.
Exception Info: System.ArgumentNullException
   at System.Number.StringToNumber(System.String, System.Globalization.NumberStyles, NumberBuffer ByRef, System.Globalization.NumberFormatInfo, Boolean)
   at System.Number.ParseInt32(System.String, System.Globalization.NumberStyles, System.Globalization.NumberFormatInfo)
   at System.Int32.Parse(System.String)
   at CefSharp.BrowserSubprocess.Program.Main(System.String[])
```

Debugging showed me that no such argument was being passed to the process in these cases:

image

This PR works around the issue by opting out from getting the parent process ID for such processes.

### DIFF
--- a/CefSharp.BrowserSubprocess/Program.cs
+++ b/CefSharp.BrowserSubprocess/Program.cs
@@ -4,6 +4,7 @@
 
 using System;
 using System.Diagnostics;
+using System.Linq;
 using System.Threading.Tasks;
 using CefSharp.Internals;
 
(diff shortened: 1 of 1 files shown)


perlun:
Thanks @amaitland - I cherry-picked this to master in 2e966d99 now also.

amaitland:
Thanks @perlun removed redundant `using System.Linq` in commit https://github.com/cefsharp/CefSharp/commit/383734efdf798a90c619b64f3c3ea7dd6fa5960e

Strange that on `Windows 10` I wasn't seeing any sort of crash, just the process wasn't starting. This should resolve #2440

perlun:
> Thanks @perlun removed redundant using System.Linq in commit 383734e

Ah, thanks! It was a remnant from an early version of the code where I was scanning through the list of arguments for the process type but then I realized it was kind of pointless since that's already been taken care of by the existing code. 😉 

> Strange that on Windows 10 I wasn't seeing any sort of crash, just the process wasn't starting. 

That's interesting. I was running this on a Windows 7 VM actually.

PR #2474, opened 2018-07-29
Setting screen info from C# code 

alfein:
On the core level, I'd like to be able to have the screen object on the javascript side to properly reflect the size and the available size of the monitor hosting the window where browser control lives.

On the Wpf Side, I added an example of how.  

@amaitland, at a minimum there should be a way to propagate this information from the C# callback of GetScreenInfo(), even if at the core level, you chose to not actually implement it, the behavior should be as is.

### DIFF
--- a/CefSharp.Core/Internals/RenderClientAdapter.h
+++ b/CefSharp.Core/Internals/RenderClientAdapter.h
@@ -47,14 +47,17 @@ namespace CefSharp
 
                 auto screenInfo = _renderWebBrowser->GetScreenInfo();
 
-                if (screenInfo.HasValue == false || screen_info.device_scale_factor == screenInfo.Value.ScaleFactor)
+                if (screenInfo.HasValue == false)
                 {
                     return false;
                 }
 
-                //NOTE: We're relying on a call to GetViewRect to populate the view rectangle
-                //https://bitbucket.org/chromiumembedded/cef/src/47e6d4bf84444eb6cb4d4509231a8c9ee878a584/include/cef_render_handler.h?at=2357#cef_render_handler.h-90
                 screen_info.device_scale_factor = screenInfo.Value.ScaleFactor;
+				screen_info.rect.width = screenInfo.Value.Width;
+				screen_info.rect.height = screenInfo.Value.Height;
+				screen_info.available_rect.width = screenInfo.Value.AvailableWidth;
+				screen_info.available_rect.height = screenInfo.Value.AvailableHeight;
+
                 return true;
             }
 
--- a/CefSharp.Core/Internals/RenderClientAdapter.h
+++ b/CefSharp.Core/Internals/RenderClientAdapter.h
@@ -14,164 +14,167 @@ using namespace CefSharp::Structs;
 
 namespace CefSharp
 {
-    namespace Internals
-    {
-        private class RenderClientAdapter : public ClientAdapter,
-            public CefRenderHandler
-        {
-        private:
-            gcroot<IRenderWebBrowser^> _renderWebBrowser;
+	namespace Internals
+	{
+		private class RenderClientAdapter : public ClientAdapter,
+			public CefRenderHandler
+		{
+		private:
+			gcroot<IRenderWebBrowser^> _renderWebBrowser;
 
-        public:
-            RenderClientAdapter(IWebBrowserInternal^ webBrowserInternal, IBrowserAdapter^ browserAdapter):
-                ClientAdapter(webBrowserInternal, browserAdapter)
-            {
-                _renderWebBrowser = dynamic_cast<IRenderWebBrowser^>(webBrowserInternal);
-            }
+		public:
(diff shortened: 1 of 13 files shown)


amaitland:
Please review your formatting and revert all the unnecessary changes. 

Spaces not tabs also.

alfein:
@amaitland, thanks for checking in.  I fixed formatting of the C++ code and removed all other code from the sample project.    I also added comments to the C++ file.  The changes to the ScreenInfo struct are well-documented.  I don't think any other comments are required.

I think this change is going to be quite useful for anyone willing to add additional C# logic to return monitor information, as currently the 


```javaccript
window.screen
```
object returns wrong values.

amaitland:
There are still a huge number of unnecessary formatting changes, please revert. 

Review using the files changed tab up the top of this page.

amaitland:
I'd like them reverted thanks.Makes the review easier. I think you've changed files that can be fully reverted at a quick glance.

CefSharp.Core/Internals/RenderClientAdapter.h
```diff
@@ -23,7 +23,7 @@ namespace CefSharp
             gcroot<IRenderWebBrowser^> _renderWebBrowser;
 
         public:
-            RenderClientAdapter(IWebBrowserInternal^ webBrowserInternal, IBrowserAdapter^ browserAdapter):
+            RenderClientAdapter(IWebBrowserInternal^ webBrowserInternal, IBrowserAdapter^ browserAdapter) :
```
amaitland: This I don't particularly have s problem with, doesn't make the review harder.

CefSharp.Wpf.Example/MainWindow.xaml.cs
```diff
@@ -4,12 +4,12 @@
 
 using System;
 using System.Collections.ObjectModel;
-using System.Threading.Tasks;
 using System.Windows;
 using System.Windows.Input;
 using CefSharp.Example;
 using CefSharp.Wpf.Example.Controls;
 using CefSharp.Wpf.Example.ViewModels;
+using CefSharp.Wpf.Example.Views;
```
amaitland: This file looks like it can be fully reverted.

CefSharp.Wpf/ChromiumWebBrowser.cs
```diff
@@ -26,7 +26,7 @@
 using CursorType = CefSharp.Enums.CursorType;
 using Rect = CefSharp.Structs.Rect;
 
-namespace CefSharp.Wpf 
+namespace CefSharp.Wpf
```
amaitland: Are code changes actually required in this class? Looks like formatting only, this sort of thing makes reviews harder

CefSharp/Structs/ScreenInfo.cs
```diff
@@ -7,10 +7,40 @@ namespace CefSharp.Structs
...
+
+        /// <summary>
+        /// Width of the monitor nearest to the window hosting the browser
+        /// </summary>
+        public int Width { get; set; }
+
+        /// <summary>
+        /// Available Left location  of the monitor nearest to the window hosting the browser.  iF toolbar is on the left - this would be a non-zero value
+        /// </summary>
+        public int AvailableLeft { get; set; }
```
amaitland: Use Rect to represent the 4 values https://github.com/cefsharp/CefSharp/blob/master/CefSharp/Structs/Rect.cs

Maybe the AvailableRect property nullable?

CefSharp/Structs/ScreenInfo.cs
```diff
@@ -7,10 +7,40 @@ namespace CefSharp.Structs
...
+
         public ScreenInfo(float scaleFactor) : this()
         {
             ScaleFactor = scaleFactor;
         }
+
+        /// <summary>
+        /// Height of the monitor nearest to the window hosting the browser
+        /// </summary>
+        public int Height { get; set; }
```
amaitland: I think for completeness depth and the other properties should be added.

CefSharp.Wpf/ChromiumWebBrowser.cs
amaitland: Can this whole file be reverted? Looks like formatting changes only

CefSharp.Wpf.Example/MainWindow.xaml.cs
```diff
@@ -4,12 +4,12 @@
 
 using System;
 using System.Collections.ObjectModel;
-using System.Threading.Tasks;
```
amaitland: Revert this file if there are only formatting changes

alfein:
@amaitland, thanks for making changes, but I would have done that over some weekend.  Apologies for being not too responsive.  

Again, thanks for finishing it.  Hope it wasn't too burdensome.

amaitland:
> Again, thanks for finishing it. Hope it wasn't too burdensome.

Was very straight forward 😄 

Changes have been merged into `cefsharp/67` branch

PR #2480, opened 2018-08-07
Add OSR AccessiblityHandler.

igor-eremyashev:
I'm working on Accessibility support. I implemented CefAccessibilityHandler and added IAccessibilityHandler interface. Handling accessiblity tree and location changes is still to be implemented, as well as creating custom AutomationPeers, but this might be useful as a starting point.

### DIFF
--- a/CefSharp.Core/Internals/ClientAdapter.h
+++ b/CefSharp.Core/Internals/ClientAdapter.h
@@ -29,8 +29,7 @@ namespace CefSharp
             public CefDialogHandler,
             public CefDragHandler,
             public CefDownloadHandler,
-            public CefFindHandler,
-			public CefAccessibilityHandler
+            public CefFindHandler
         {
         private:
             gcroot<IWebBrowserInternal^> _browserControl;
@@ -193,10 +192,6 @@ namespace CefSharp
 
             //CefFindHandler
             virtual DECL void OnFindResult(CefRefPtr<CefBrowser> browser, int identifier, int count, const CefRect& selectionRect, int activeMatchOrdinal, bool finalUpdate);
-
-			//CefAccessibilityHandler
-			virtual DECL void OnAccessibilityLocationChange(CefRefPtr<CefValue> value);
-			virtual DECL void OnAccessibilityTreeChange(CefRefPtr<CefValue> value);
 
             //sends out an eval script request to the render process
             Task<JavascriptResponse^>^ EvaluateScriptAsync(int browserId, bool isBrowserPopup, int64 frameId, String^ script, String^ scriptUrl, int startLine, Nullable<TimeSpan> timeout);
--- a/CefSharp.Core/Internals/ClientAdapter.h
+++ b/CefSharp.Core/Internals/ClientAdapter.h
@@ -29,7 +29,8 @@ namespace CefSharp
             public CefDialogHandler,
             public CefDragHandler,
             public CefDownloadHandler,
-            public CefFindHandler
+            public CefFindHandler,
+			public CefAccessibilityHandler
         {
         private:
             gcroot<IWebBrowserInternal^> _browserControl;
@@ -192,6 +193,10 @@ namespace CefSharp
 
             //CefFindHandler
             virtual DECL void OnFindResult(CefRefPtr<CefBrowser> browser, int identifier, int count, const CefRect& selectionRect, int activeMatchOrdinal, bool finalUpdate);
+
+			//CefAccessibilityHandler
+			virtual DECL void OnAccessibilityLocationChange(CefRefPtr<CefValue> value);
+			virtual DECL void OnAccessibilityTreeChange(CefRefPtr<CefValue> value);
 
             //sends out an eval script request to the render process
             Task<JavascriptResponse^>^ EvaluateScriptAsync(int browserId, bool isBrowserPopup, int64 frameId, String^ script, String^ scriptUrl, int startLine, Nullable<TimeSpan> timeout);
--- a/CefSharp.Core/Internals/TypeConversion.h
+++ b/CefSharp.Core/Internals/TypeConversion.h
@@ -187,32 +187,32 @@ namespace CefSharp
 
                 auto type = value->GetType();
 
-                if (type == CefValueType::VTYPE_BOOL)
+                if (type == ::CefValueType::VTYPE_BOOL)
                 {
                     return value->GetBool();
                 }
 
-                if (type == CefValueType::VTYPE_DOUBLE)
+                if (type == ::CefValueType::VTYPE_DOUBLE)
                 {
                     return value->GetDouble();
                 }
 
-                if (type == CefValueType::VTYPE_INT)
+                if (type == ::CefValueType::VTYPE_INT)
                 {
                     return value->GetInt();
                 }
 
-                if (type == CefValueType::VTYPE_STRING)
+                if (type == ::CefValueType::VTYPE_STRING)
                 {
                     return StringUtils::ToClr(value->GetString());
                 }
 
-                if (type == CefValueType::VTYPE_DICTIONARY)
+                if (type == ::CefValueType::VTYPE_DICTIONARY)
                 {
                     return FromNative(value->GetDictionary());
                 }
 
-                if (type == CefValueType::VTYPE_LIST)
+                if (type == ::CefValueType::VTYPE_LIST)
                 {
                     return FromNative(value->GetList());
                 }
--- a/CefSharp.Core/Internals/TypeConversion.h
+++ b/CefSharp.Core/Internals/TypeConversion.h
@@ -15,6 +15,7 @@
 #include "Serialization\V8Serialization.h"
 
 using namespace System::Collections::Generic;
+using namespace System::Collections::Specialized;
 using namespace CefSharp::Internals::Serialization;
 
 namespace CefSharp
--- a/CefSharp.Core/Internals/TypeConversion.h
+++ b/CefSharp.Core/Internals/TypeConversion.h
@@ -212,10 +212,10 @@ namespace CefSharp
                     return FromNative(value->GetDictionary());
                 }
 
-				if (type == CefValueType::VTYPE_LIST)
-				{
-					return FromNative(value->GetList());
-				}
+                if (type == CefValueType::VTYPE_LIST)
+                {
+                    return FromNative(value->GetList());
+                }
                 
                 return nullptr;
             }
@@ -243,16 +243,16 @@ namespace CefSharp
                 return dict;
             }
 
-			static List<Object^>^ FromNative(const CefRefPtr<CefListValue>& list)
-			{
-				auto result = gcnew List<Object^>(list->GetSize());
-				for (auto i = 0; i < list->GetSize(); i++)
-				{
-					result->Add(DeserializeObject(list, i, nullptr));
-				}
+            static List<Object^>^ FromNative(const CefRefPtr<CefListValue>& list)
+            {
+                auto result = gcnew List<Object^>(list->GetSize());
+                for (auto i = 0; i < list->GetSize(); i++)
+                {
+                    result->Add(DeserializeObject(list, i, nullptr));
+                }
 
-				return result;
-			}
+                return result;
+            }
         };
     }
 }
--- a/CefSharp.Core/Internals/TypeConversion.h
+++ b/CefSharp.Core/Internals/TypeConversion.h
@@ -211,6 +211,11 @@ namespace CefSharp
                 {
                     return FromNative(value->GetDictionary());
                 }
+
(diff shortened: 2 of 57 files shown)


CefSharp.Core/Internals/ClientAdapter.h
```diff
@@ -193,6 +194,10 @@ namespace CefSharp
             //CefFindHandler
             virtual DECL void OnFindResult(CefRefPtr<CefBrowser> browser, int identifier, int count, const CefRect& selectionRect, int activeMatchOrdinal, bool finalUpdate);
 
+			//CefAccessibilityHandler
```
amaitland: These should be moved into `RenderClientAdapter`. They're only called as part of `CefRenderHandler`

amaitland: Also spaces not tabs

CefSharp.Example/CefExample.cs
```diff
@@ -95,6 +95,9 @@ public static void Init(AbstractCefSettings settings, IBrowserProcessHandler bro
             //Possibly useful when experiencing blury fonts.
             //settings.CefCommandLineArgs.Add("disable-direct-write", "1");
 
+            //settings.CefCommandLineArgs.Add("force-renderer-accessibility", "1");
```
amaitland: Please add a comment/link as to what this command line argument does (or remove it if it's not relevant).

CefSharp.WinForms/ChromiumWebBrowser.cs
```diff
@@ -182,6 +182,13 @@ public IRequestContext RequestContext
         [Browsable(false), EditorBrowsable(EditorBrowsableState.Never), DefaultValue(null)]
         public IFindHandler FindHandler { get; set; }
 
+        /// <summary>
+        /// Implement <see cref="IAccessibilityHandler" /> to handle events related to accessibility.
+        /// </summary>
+        /// <value>The accessibility handler.</value>
+        [Browsable(false), EditorBrowsable(EditorBrowsableState.Never), DefaultValue(null)]
+        public IAccessibilityHandler AccessibilityHandler { get; set; }
```
amaitland: This should be removed as it will never be called.

CefSharp/Handler/IAccessibilityHandler.cs
```diff
@@ -0,0 +1,25 @@
...
+    /// <summary>
+    /// Implement this interface to receive accessibility notification when accessibility events have been registered. 
+    /// The methods of this class will be called on the UI thread.
+    /// </summary>
+    public interface IAccessibilityHandler
+    {
+        /// <summary>
+        /// Called after renderer process sends accessibility location changes to the browser process.
+        /// </summary>
+        /// <param name="value"></param>
```
amaitland: Please add comment

CefSharp/Handler/IAccessibilityHandler.cs
```diff
@@ -0,0 +1,25 @@
...
+        void OnAccessibilityLocationChange(object value);
+
+        /// <summary>
+        /// Called after renderer process sends accessibility tree changes to the browser process.
+        /// </summary>
+        /// <param name="value"></param>
+        void OnAccessibilityTreeChange(object value);
```
amaitland: Please add comment

CefSharp/IWebBrowser.cs
```diff
@@ -187,6 +187,12 @@ public interface IWebBrowser : IDisposable
         /// <value>The find handler.</value>
         IFindHandler FindHandler { get; set; }
 
+        /// <summary>
+        /// Implement <see cref="IAccessibilityHandler" /> to handle events related to accessibility.
+        /// </summary>
+        /// <value>The accessibility handler.</value>
+        IAccessibilityHandler AccessibilityHandler { get; set; }
```
amaitland: Remove this as it's only valid for `WPF` and `OffScreen`

CefSharp.Core/Internals/TypeConversion.h
```diff
@@ -237,6 +242,17 @@ namespace CefSharp
 
                 return dict;
             }
+
+			static List<Object^>^ FromNative(const CefRefPtr<CefListValue>& list)
```
amaitland: Spaces not tabs

CefSharp/IWebBrowser.cs
igor-eremyashev: Should I move AccessibilityHandler to IRenderWebBrowser?

amaitland: Hmm, probably not. It's only implemented explicitly.

Shouldn't make much difference if you don't add it to an interface.

igor-eremyashev: But if I move OnAccessibilityLocationChange and OnAccessibilityTreeChange implementations to RenderClientAdapter.h then IRenderWebBrowser should have reference to AccessibilityHandler, right? RenderClientAdapter references IRenderWebBrowser.

amaitland: It does, my mistake. Just don't implement the property explicitly.

igor-eremyashev: Not sure if I got it correctly, but I moved CefAccessibilityHandler implementation to RenderClientAdapter and added IAccessibilityHandler to IRenderWebBrowser. Looks like it makes sense since we only need IAccessibilityHandler for WPF and OffScreen.

CefSharp/Handler/IAccessibilityHandler.cs
```diff
@@ -0,0 +1,25 @@
...
+    /// Implement this interface to receive accessibility notification when accessibility events have been registered. 
+    /// The methods of this class will be called on the UI thread.
+    /// </summary>
+    public interface IAccessibilityHandler
+    {
+        /// <summary>
+        /// Called after renderer process sends accessibility location changes to the browser process.
+        /// </summary>
+        /// <param name="value">Updated location info.</param>
+        void OnAccessibilityLocationChange(object value);
```
amaitland: Does passing in `object` make sense here? Would it be better to create a wrapper around `CefValue`?

You can convert `CefValue` to `Json`, though not sure if that helps.

http://magpcss.org/ceforum/apidocs3/projects/(default)/(_globals).html#CefWriteJSON(CefRefPtr%3CCefValue%3E,cef_json_writer_options_t)

igor-eremyashev: Creating a wrapper is the next step I'm going to work on. For now I don't fully understand the structure on the objects being passed from CEF. All I have now is this code from CefClient
https://bitbucket.org/chromiumembedded/cef/src/a659b48fd1c041984cfaa0d335d60943bd2c6991/tests/cefclient/browser/osr_accessibility_helper.cc?at=master&fileviewer=file-view-default
but it is hard to tell what is the complete structure of there objects as it might not handle all the cases. Also they might change the structure in the future, so it might be easier to implement the parsing in C#.

amaitland: I know nothing about accessibility, so really cannot say. A wrapper around `CefValue` sounds reasonable.

igor-eremyashev: Sorry, I was thinking about something like `AccessibilityLocationChangeArgs` or `AccessibilityTreeChangeArgs`, this kind of wrapper. `CefValue` looks very much like `object`, it is a base class for various types: `int`, `bool`, `string`, `Dictionary`, `List` and others. So it does what `object` does. `TypeConversion.h` methods accept and return `object`:
`static CefRefPtr<CefValue> ToNative(Object^ value)`
`static Object^ FromNative(const CefRefPtr<CefValue>& value)`
I don't like passing `object` as a method argument, but it seems to make sense here.
What do you think?

CefSharp/Handler/IAccessibilityHandler.cs
```diff
@@ -0,0 +1,25 @@
+﻿// Copyright © 2010-2018 The CefSharp Authors. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style license that can be found in the LICENSE file.
+
+namespace CefSharp
+{
+    /// <summary>
+    /// Implement this interface to receive accessibility notification when accessibility events have been registered. 
+    /// The methods of this class will be called on the UI thread.
```
amaitland: Change to `It's important to note this event is fired on a CEF UI thread, which by default is not the same as your application UI thread`

CefSharp.Core/Internals/TypeConversion.h
```diff
@@ -187,30 +187,35 @@ namespace CefSharp
...
                     return StringUtils::ToClr(value->GetString());
                 }
 
-                if (type == CefValueType::VTYPE_DICTIONARY)
+                if (type == ::CefValueType::VTYPE_DICTIONARY)
                 {
                     return FromNative(value->GetDictionary());
                 }
+
+                if (type == ::CefValueType::VTYPE_LIST)
```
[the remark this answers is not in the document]
igor-eremyashev: Actually I found it useful to have IValue.ToObject() method while working on custom AutomationPeer. This way it is much easier to inspect accessibility events during debugging. So I added IValue.ToObject method and kept this code.

CefSharp/IWebBrowser.cs
```diff
@@ -186,7 +186,7 @@ public interface IWebBrowser : IDisposable
         /// </summary>
         /// <value>The find handler.</value>
         IFindHandler FindHandler { get; set; }
-
```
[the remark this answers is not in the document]
igor-eremyashev: Do you mean to restore AccessibilityHandler in IWebBrowser?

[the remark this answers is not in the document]
amaitland: No, you've made formatting changes that are unnecessary.

CefSharp/IValue.cs
```diff
@@ -0,0 +1,61 @@
...
+        /// Returns the underlying value as type list.
+        /// </summary>
+        /// <returns></returns>
+        IList<IValue> GetList();
+
+        /// <summary>
+        /// Returns the underlying value as a dynamic object.
+        /// </summary>
+        /// <returns></returns>
+        object GetObject();
```
amaitland: Is this comment correct?

CefSharp/IValue.cs
```diff
@@ -0,0 +1,61 @@
...
+    /// </summary>
+    public interface IValue : IDisposable
+    {
+        ValueType Type { get; }
+
+        /// <summary>
+        /// Returns the underlying value as type bool.
+        /// </summary>
+        /// <returns></returns>
+        bool GetBool();
```
amaitland: Add missing return comment,  same for the other methods.

CefSharp.Core/Internals/TypeConversion.h
```diff
@@ -15,6 +15,7 @@
 #include "Serialization\V8Serialization.h"
 
 using namespace System::Collections::Generic;
+using namespace System::Collections::Specialized;
```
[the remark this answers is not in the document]
igor-eremyashev: CefResponse::HeaderMap ToNative(NameValueCollection^ headers) uses NameValueCollection which is located in System::Collections::Specialized. I'm not sure how it was compiling before.

CefSharp/IValue.cs
igor-eremyashev: Probably it would be better to say 'as an object'.

CefSharp/IValue.cs
igor-eremyashev: Looks like in all these methods `returns` comments will be the same as `summary` comments, I don't know what should I add.

CefSharp.Core/Internals/TypeConversion.h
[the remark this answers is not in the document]
amaitland: One of the other headers must include a reference.

Fine to leave.

CefSharp/IValue.cs
amaitland: Probably should also mention that it's converted a managed object.

CefSharp/IValue.cs
amaitland: Duplicate is fine, they'll appear in the published `API Doc` (http://cefsharp.github.io/api/)

amaitland:
Missing `OVERRIDE` statements adding in https://github.com/cefsharp/CefSharp/commit/fe1d542b549a629dd8e958203a180757c3d01a19

amaitland:
Merged into `cefsharp/67`

igor-eremyashev:
@amaitland Is there a way to merge this changes back to v65.1? v67 and v69 seem to have rendering issues (#2547).

amaitland:
@igor-eremyashev Technically I believe it's possible. If you're asking me to back port the changes and release a new version then you would have to pay for my time. Obviously nothing stopping you from performing a build yourself.

Issue #2053, opened 2017-05-18
Add Accessibility Support


RichiCoder1:
The Cef project just recently merged accessibility support of OSR (https://bitbucket.org/chromiumembedded/cef/pull-requests/99/issue-1217-cef3-accessibility-support/diff) and it would be terrific to see that supported in the next version of CefSharp.

amaitland:
https://bitbucket.org/chromiumembedded/cef/commits/406867f293042006e9f5f1a1e5201db012e50adf

RichiCoder1:
I might (maybe) be able to pick this up. Not overly familiar with either Cef or CefSharp though. I looks like it's a matter of forwarding accessibility events? (Which I have no idea how to do, but I can at least look into)

amaitland:
`IBrowserHost.SetAccessibilityState`  added in commit https://github.com/cefsharp/CefSharp/commit/921ea55fff0495c3e1895d7d4ba694e06163fca3

From a wrapper point of view `CefAccessibilityHandler` needs to be implemented and `IAccessibilityHandler` created (same as any of the other handlers)

BenKeyFSI:
What is the status of this? Did commit 921ea55 resolve the issue? If so, why is the CefSharp.WinForms.Test application still inaccesible? Do I need to add a call to the SetAccessibilityState function somewhere?

amaitland:
This issue relates to WPF. 

See https://stackoverflow.com/questions/42163764/how-to-make-jaws-screen-reader-recognize-and-read-content-of-cefsharp-chromiumwe for an example of how to get the WinForms version working.

BenKeyFSI:
I meant the CefSharp.Wpf.Example. The Windforms example currently works as is. Unfortunately, the WPF one does not. I have a key client that wants to use CEF in a WPF application with JAWS.

amaitland:
Pull requests welcome

BenKeyFSI:
I would be glad to work on this issue, if it is still open. However, my question regarding the status of this issue has yet to be answered. I have a key client who wants JAWS to be able to work in an application that uses CEF in a WPF application.

I found this search while doing a Google search. I am a little unclear on the status. The last comment before my first comment is a commit message that seems to be related. An yet the issue is still open. Is that because more work needs to be done?

Specifically, I do not understand the following comment.

"From a wrapper point of view CefAccessibilityHandler needs to be implemented and IAccessibilityHandler created (same as any of the other handlers)."

[...]

PR #2495, opened 2018-08-21
WPF - Add basic UI Accessibility support.

igor-eremyashev:
This pull request adds basic UI automation support. Accessibility tree parsing was implemented using `cefclient` source code as a reference:
https://bitbucket.org/chromiumembedded/cef/src/ea0e213bef0367edfedb7cf0a487e04e8c6cdf00/tests/cefclient/browser/osr_accessibility_node_win.cc?at=master&fileviewer=file-view-default
Chromium sources were also used:
https://cs.chromium.org/chromium/src/ui/accessibility/ax_tree_source.h?dr=C&g=0
Windows Narrator and inspect.exe (part of Windows SDK) were used for testing. 
Three steps are required to enable UI automation support:

1. Pass `force-renderer-accessibility` flag as a command line argument
2. Replace `ChromiumWebBrowser` with `ChromiumWebBrowserWithAccessibilitySupport` 
3. Set AccessibilityHandler for the browser control 

The accessibility support is very basic at this point and doesn't handle cases like text selection, node attribute updates and so on.


timaiv:
Successfully Implemented Accessibility support for me based on this PR.
Some notes:

- root tree now (in v75, v79) some differs from PR.
- GetClickablePointCore in PR calculated wrong (at least no PointToScreen, you can take GetBoundingRectangleCore). You can calculate as half of width\height like in UIElementAutomationPeer, or something from visible intersect. 
- GetLocation caculated wrong (you also need add  -offsetNode.ScrollY, -offsetNode.ScrollX).
- in IsOffscreenCore i used intersected visible rect and window location info. (like mostly in win32, wpf uses IsOffscreenBehavior Enum)
- also i used in GetClassNameCore, GetLocalizedControlTypeCore role. (not empty string).

igor-eremyashev:
@timaiv Good to hear you did it! Could you prepare your own pull request with the changes you did? It would be great to have accessibility support with all the fixes and changes finally merged to master.

smallb2007:
@timaiv 
Could you please share your implementation?

justindarc:
I spent several days investigating taking this PR, rebasing it to `master` and was able to get it "working". However, I believe this is the wrong approach to solving the WPF accessibility problem. If you use [inspect.exe](https://docs.microsoft.com/en-us/windows/win32/winauto/inspect-objects) to examine the UIAutomation/MSAA tree produced by this PR (or any approach similar to this), you will see that the tree appears completely different than what the Chrome browser itself produces.

Specifically, there are several key differences:
- The `FrameworkId` in the UIA nodes are reported as "WPF" whereas the Chrome browser (and other native Win32-based CEF examples) report "Chrome".
- The `ProviderDescription` with this approach reports "Unidentified Provider (managed:MS.Internal.Automation.ElementProxy)" and Chrome (and other native Win32-based CEF examples) report "Microsoft: MSAA Proxy (IAccessible2) (unmanaged:UIAutomationCore.DLL)"
- Chrome (and other native Win32-based CEF examples) report an `AriaRole` attribute that simply does not exist when taking this approach

But, most importantly, there is no possible way to take this approach and make elements "invokable". The screen reader can certainly take the UIAutomation tree produced by this PR and allow the user to traverse and "read" the content. However, there is no way to "invoke" the elements to match the expected behavior. This would require having a direct reference to the corresponding DOM node so that when the UIAutomation element gets invoked, it can forward that event to it.

It occurred to me that CEF itself is already producing a UIAutomation/MSAA accessibility tree and this approach is essentially attempting to reinvent the wheel by re-creating that tree a second time. However, since CEF's accessibility tree is deep within unmanaged code, it is not possible (or at least trivial) to get direct references to the nodes within that tree to make them invokable. I then realized that the problem is actually a bug that the CEF UIAutomation/MSAA accessibility tree is not being exposed properly. This *should* happen automatically when hosting the hwnd for the Chromium browser instance in the WPF control, but it was not.

After spending some more time investigating this problem from this totally different perspective, I came up with a fix in #3173 that is only 2 lines of code. Aside from setting the `force-renderer-accessibility` command line argument for CEF, the actual fix comes inside `CreateOffscreenBrowserWindowInfo` in ChromiumWebBrowser.cs. When we create the `WindowInfo`, it is necessary to use `SetAsChild` to attach the Chromium browser's hwnd to the control in order for the UIAutomation/MSAA accessibility tree to be exposed properly. Previously, `SetAsWindowless` was being used which is why there was no accessibility tree being exposed at all.

I hope the fix in #3173 helps others who have also been struggling with this problem!

amaitland:
> I spent several days investigating taking this PR, rebasing it to `master` and was able to get it "working". However, I believe this is the wrong approach to solving the WPF accessibility problem. I

Looking at https://github.com/justindarc/CefSharp/commit/5a7f767dd66cbee5e00ff4ee7dcba334d45e5538 and this looks like a step forward, I'll pull this into `master` and possibly move this into the `CefSharp.Wpf.Experimental` namespace to allow for easier use/testing. Thanks 👍 

Once I've cheery picked those changes I'll be closing this `PR`.

> But, most importantly, there is no possible way to take this approach and make elements "invokable". The screen reader can certainly take the UIAutomation tree produced by this PR and allow the user to traverse and "read" the content. However, there is no way to "invoke" the elements to match the expected behavior. This would require having a direct reference to the corresponding DOM node so that when the UIAutomation element gets invoked, it can forward that event to it.

With the `CEF API`  as it stands we are limited in the implementation we can provide. For `Accessibility` using the native `Chromium` implementation is always going to provide a better user experience.

> After spending some more time investigating this problem from this totally different perspective, I came up with a fix in #3173 that is only 2 lines of code. Aside from setting the `force-renderer-accessibility` command line argument for CEF, the actual fix comes inside `CreateOffscreenBrowserWindowInfo` in ChromiumWebBrowser.cs. When we create the `WindowInfo`, it is necessary to use `SetAsChild` to attach the Chromium browser's hwnd to the control in order for the UIAutomation/MSAA accessibility tree to be exposed properly. Previously, `SetAsWindowless` was being used which is why there was no accessibility tree being exposed at all.

Unfrtunately this isn't a viable solution as I've detailed in #3173 

I am working on a drop in replacement for `CefSharp.Wpf` that hosts the native `Chromium` browser in a `HwndHost` which is avaliable at https://github.com/cefsharp/CefSharp.Wpf.HwndHost (alternatively you can use the `WinForms` verison in a `WindowsFormsHost).

justindarc:
> Looking at [justindarc@5a7f767](https://github.com/justindarc/CefSharp/commit/5a7f767dd66cbee5e00ff4ee7dcba334d45e5538) and this looks like a step forward, I'll pull this into `master` and possibly move this into the `CefSharp.Wpf.Experimental` namespace to allow for easier use/testing. Thanks 👍
> 
> Once I've cheery picked those changes I'll be closing this `PR`.
> 

Sounds good. Thanks!

> > After spending some more time investigating this problem from this totally different perspective, I came up with a fix in #3173 that is only 2 lines of code. Aside from setting the `force-renderer-accessibility` command line argument for CEF, the actual fix comes inside `CreateOffscreenBrowserWindowInfo` in ChromiumWebBrowser.cs. When we create the `WindowInfo`, it is necessary to use `SetAsChild` to attach the Chromium browser's hwnd to the control in order for the UIAutomation/MSAA accessibility tree to be exposed properly. Previously, `SetAsWindowless` was being used which is why there was no accessibility tree being exposed at all.
> 
> Unfrtunately this isn't a viable solution as I've detailed in #3173
> 
> I am working on a drop in replacement for `CefSharp.Wpf` that hosts the native `Chromium` browser in a `HwndHost` which is avaliable at https://github.com/cefsharp/CefSharp.Wpf.HwndHost (alternatively you can use the `WinForms` verison in a `WindowsFormsHost).

I was able to get your `HwndHost` control up and running and it seems to effectively do exactly what I was aiming for in #3173. Please let me know if you have a list of remaining issues and I'll see if I can help get this closer to being releasable. Thanks again!

amaitland:
> I was able to get your `HwndHost` control up and running and it seems to effectively do exactly what I was aiming for in #3173. Please let me know if you have a list of remaining issues and I'll see if I can help get this closer to being releasable. Thanks again!

@justindarc I've created https://github.com/cefsharp/CefSharp.Wpf.HwndHost/issues/1 to track creating a set of `Nuget` packages. I think it's probably ready for a beta/pre release. I only have a single machine to test on, so the testing I can do is a little limited (I've only tested on Win 10 1903 with 125% DPI). Happy to discuss further on https://github.com/cefsharp/CefSharp.Wpf.HwndHost/issues/1 (or create a new issue for any problems you find) 👍

amaitland:
> Looking at [justindarc@5a7f767](https://github.com/justindarc/CefSharp/commit/5a7f767dd66cbee5e00ff4ee7dcba334d45e5538) and this looks like a step forward, I'll pull this into `master` and possibly move this into the `CefSharp.Wpf.Experimental` namespace to allow for easier use/testing. Thanks 👍

Merged into master, see https://github.com/cefsharp/CefSharp/issues/2053#issuecomment-664301558 for details.

Closing this `PR` as it's been superseded.

PR #2498, opened 2018-08-31
Make CefSharp compatible  with .NET Framework 4.0

MikeDiuga:
Make CefSharp compatible  with .NET Framework 4.0

I have noticed elsewhere many requests for CefSharp to support .NET Framework 4.0 - for solutions that can't be migrated to .NET >= 4.5.2.
This solution implements compatibility by using Microsoft's BCL packages, as well as abstractions for .NET 4.0 and above, as much as is possible (TaskHelpers etc.).

Users of the .NET 4.0 compatible assemblies can prevent cascading the requirement of the Microsoft BCL packages for dependent assemblies (MS have a build target that enforces this) by adding the following to their project files:
```
<PropertyGroup>
  <SkipValidatePackageReferences>true</SkipValidatePackageReferences>
</PropertyGroup>
```


vwmberry95:
Are you no longer working on this?  I'm curious to see how it ends up.

MikeDiuga:
Hi @vwmberry95!
I will be, but ran into these build problems with VS2015 late on Friday - that I don't currently have installed. I was developing this against VS2017 and MSBuild 15, where it builds and functions fine.

The issue should be easy to fix, so hopefully I'll get time in a day or two :)

MikeDiuga:
@vwmberry95 and anyone else who's interested, I will be maintaining a .NET Framework compatible 4.0 branch here, if you're interested :)
https://github.com/emisgroup/CefSharp.Net40

Expect updates to lag behind official releases.

PR #2504, opened 2018-09-08
Maintenance - Fixed formatting, file encoding, and copyright year

merceyz:
- Applied formatting
- Fixed file encoding inconsistencies, was a mix of ANSI, UTF8, and UTF8-BOM. Now it's UTF8
- Made copyright note show the year the file was created


cztomczak:
Just IMO. Copyright year note in a file should state in which year the copyright started. You can find such information e.g. on wikipedia which explains how copyright notes work. For example if a new file is created in 2014 it should state `Copyright (c) 2014 The CefSharp Authors`. There is no need to update copyright year each next year.. because the copyright doesn't expire. Another thing is that when I enter for example the `JavascriptAsyncMethodCallback.cpp` file I can see in it `// Copyright © 2010-2018 The CefSharp Authors. All rights reserved.` which seems wrong. When I click git "History" in that file I can clearly see that this file was created in year 2015, so it is not possible for you to keep copyright on the code from that file in year 2010.

merceyz:
I was just following what has been done previously however, you're right.
I could fix it so that it shows the year the file was created and nothing more, so that we'll never have to update it again.

If @amaitland agrees I'll go ahead and commit that

amaitland:
Honestly I hadn't though twice about the copyright year, it was just copied and pasted from files that predate my time. Having the year the file was created makes sense, thanks for the suggestion @cztomczak  👍 and thanks @merceyz for implementing.

> Fixed file encoding inconsistencies, was a mix of ANSI, UTF8, and UTF8-BOM. Now it's UTF8

@merceyz  I'm wondering if this is fighting battle, for me when I create a `C++` file it defaults to `1252` and when I create a `C#` file it defaults to `65001`. Unless there is a solution wide setting in `VS` then I think we'll end up with a mix in no time and it's really hard to enforce.

amaitland:
> We can use EditorConfig, but that requires VS2017 to work
> https://docs.microsoft.com/en-us/visualstudio/ide/create-portable-custom-editor-options?view=vs-2017

@merceyz Great, sounds promising 👍 

When you are ready then by all means merge this (squash merge please). Thanks for your hard work.

merceyz:
Due to 'continuous-integration/appveyor/branch Expected — Waiting for status to be reported
Required' I can't.

Do you want the EditorConfig changes to be its own PR or a part of this one?

amaitland:
> Do you want the EditorConfig changes to be its own PR or a part of this one?

Separate `PR` is fine.

PR #2533, opened 2018-10-10
Make CefSharp compatible with .NET Framework 4.0

MikeDiuga:
Make CefSharp compatible with .NET Framework 4.0

I have noticed elsewhere many requests for CefSharp to support .NET Framework 4.0 - for solutions that can't be migrated to .NET >= 4.5.2.
This solution implements compatibility by using Microsoft's BCL packages, as well as abstractions for .NET 4.0 and above, as much as is possible (TaskHelpers etc.).

Users of the .NET 4.0 compatible assemblies can prevent cascading the requirement of the Microsoft BCL packages for dependent assemblies (MS have a build target that enforces this) by adding the following to their project files:

```xml
<PropertyGroup>
  <SkipValidatePackageReferences>true</SkipValidatePackageReferences>
</PropertyGroup>
```


amaitland:
Whilst this is an interesting technical achievement and I commend you for the effort 👍 It really is advisable to reach out to project maintainer(s) to see if they're willing to accept drastic before you put in a huge amount of effort. As it stands I have no interest in putting in the time required to review, test and maintain these changes. If you would like to maintain your own branch that supports `.Net 4.0` then you are more than welcome to, it can even be included as project under the main `CefSharp` organisation. You would have to publish your `Nuget` packages separately.

Having a separate branch that was specifically for `.Net 4.0` would require far fewer changes.

amaitland:
> I have noticed elsewhere many requests for CefSharp to support .NET Framework 4.0 - for solutions that can't be migrated to .NET >= 4.5.2.

`CefSharp` isn't the only `CEF` wrapper available, `CefGlue` and `ChromiumFX` both still supports `.Net 4.0` and can do everything that `CefSharp` can do, and more (`CefSharp` wrapper could be entirely written in `C#` using `CefGlue/ChromiumFx`).

PR #2535, opened 2018-10-12
FolderSchemeHandlerFactory - change resource file share mode.

igor-eremyashev:
This small change will allow hot edit and reload for local pages, which is helpful for development purposes.

### DIFF
--- a/CefSharp/SchemeHandler/FolderSchemeHandlerFactory.cs
+++ b/CefSharp/SchemeHandler/FolderSchemeHandlerFactory.cs
@@ -16,10 +16,11 @@ namespace CefSharp.SchemeHandler
     /// </summary>
     public class FolderSchemeHandlerFactory : ISchemeHandlerFactory
     {
-        private string rootFolder;
-        private string defaultPage;
-        private string schemeName;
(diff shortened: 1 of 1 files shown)


amaitland:
I'm not sure `ReadWrite` should be the default, happy for it to be configurable. It should probably be readonly when you deploy an app into production (unless you really know what your doing, in which case the `FolderSchemeHandlerFactory` probably isn't suitable).

igor-eremyashev:
It is always possible to close an application, modify its pages and then start it again. The only difference is `FileShare.Read` won't allow to do that without restart. The only way to prevent page resources from modification I can think of is to store them as embedded resources and sign the assembly.

amaitland:
> It is always possible to close an application, modify its pages and then start it again. The only difference is `FileShare.Read` won't allow to do that without restart.

It's modification whilst attempting a `Read` that I'd be concerned about.

igor-eremyashev:
Would you prefer to have `FileShare` constuctor parameter or a boolean flag to configure this behavior?

amaitland:
Using the FileShare enum makes it clearer I think.

PR #2545, opened 2018-10-21
Fix IDE1006 in FindReplaceResponseFilter.cs

Atvaark:
Error	 IDE1006	Naming rule violation: These words must begin with upper case characters: encoding


amaitland:
Thanks 👍 Will merge when `GitHub` resolves it's outage.

amaitland:
Thanks :+1:

PR #2585, opened 2018-12-03
WPF - Use nameof for DependencyProperty.Register instead of hardcoded strings

merceyz:
- Use `nameof` when registering a `DependencyProperty` instead of a hardcoded string for the property name

This has no effect on the build output, `nameof(x)` is replaced with `"x"` at compile time
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/nameof


amaitland:
Thanks :+1: 

`nameof` wasn't an option when those properties were first added.

PR #2589, opened 2018-12-06
WPF - Change PixelFormat from Bgra32 to Pbgra32 to fix control transparency

### LINKED ISSUE #2127: WPF - Transparency Rendering Too Dark
    - CefSharp.Common / CefSharp.Wpf v57.0.0

   - Either / Both

    - Windows 10 x64

   - WPF

    - Set a white coloured background behind the control
    - Enable OffScreenTransparentBackground
    - Navigate to a page with a 50% white background (body background: rgba(255,255,255,0.5), html background: transparent)

   - Background is gray rather than white.

    - No exceptions or error messages occur.

    - No, but likely doesn't use transparent background.

maloshuk:
Resolves #2127

When making CefSharp.WPF control semi-transparent and show it above another window, the control is darken. That happens because of incorrect interpretation of pixel data format received from cef.  Cef provides pixel data with premultiplied alpha (PixelFormats.Pbgra32 in .NET), while CefSharp.WPF treated them as having straight alpha (PixelFormats.Bgra32). 

The same issue was experienced in CEF delopment. 
https://bitbucket.org/chromiumembedded/cef/issues/584
Wich leaded to Skia and statement that it provides pixel data with premultiplied alpha. https://bugs.chromium.org/p/skia/issues/detail?id=547.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -19,7 +19,7 @@ The upstream [`CEF` forum](http://magpcss.org/ceforum/) is a valuable resource, 
 It maybe helpful to run the `cefclient` application and compare output with `CefSharp`. The `WinForms` and `WPF` versions use two different rendering modes, `WPF` uses Offscreen Rendering (`OSR`). `OffScreen` also uses `OSR` mode.
 
 - Download **Sample Application** from http://opensource.spotify.com/cefbuilds/index.html
-- To compare with WPF run `cefclient --multi-threaded-message-loop --off-screen-rendering-enabled`
+- To compare with WPF run `cefclient --multi-threaded-message-loop --off-screen-rendering-enabled --enable-gpu`
 - To compare with WinForms `cefclient --multi-threaded-message-loop`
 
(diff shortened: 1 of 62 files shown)


merceyz:
Shouldn't the other rendering methods be updated as well?

amaitland:
> That happens because of incorrect interpretation of pixel data format received from cef. Cef provides pixel data with premultiplied alpha (PixelFormats.Pbgra32 in .NET), while CefSharp.WPF treated them as having straight alpha (PixelFormats.Bgra32).

@maloshuk Excellent, thanks 👍 

> Shouldn't the other rendering methods be updated as well?

I believe there are six instances that require updating.

maloshuk:
> > Shouldn't the other rendering methods be updated as well?
> 
> I believe there are six instances that require updating.

Yes,  'Find all references' gives 5 more places where PixelFormats.Bgra32 is in use. But I do not know code that good to edit all of them in blind. I will find time to fix them too. 
Thanks for support. You doing great product.

amaitland:
> But I do not know code that good to edit all of them in blind

All instances should be changed :+1:

maloshuk:
> All instances should be changed 👍

Ok. I changed them.

amaitland:
Thanks

PR #2590, opened 2018-12-06
WinForms - Added missing Browsable attribute on JavascriptObjectRepository + Updated DefaultValue attribute on UseParentFormMessageInterceptor

merceyz:
- Added a missing `Browsable(false)` attribute on the property `JavascriptObjectRepository`
- Updated the `DefaultValue` attribute on the property `UseParentFormMessageInterceptor`


amaitland:
@merceyz The other `Browsable` attributes have a `DefaultValue` set, I guess this should be added for consistency or removed from all if not required. Thoughts?

```c#
[Browsable(false), EditorBrowsable(EditorBrowsableState.Never), DefaultValue(null)]
```

amaitland:
Looks fine, merge whenever you are finished 👍

PR #2592, opened 2018-12-08
Update CONTRIBUTING.md

merceyz:
- Corrected `Readd` to `Read`

Can't make a direct commit so PR it is


amaitland:
Should probably have been `re-add`. Without the line breaks it's much harder to read.

It would appear this text was extracted from https://github.com/cefsharp/CefSharp/commit/9407ba1828b473b7186e33c139bf776701460f1b

PR #2595, opened 2018-12-09
Added SourcePath to CefSharp.Core and CefSharp.BrowserSubprocess.Core

merceyz:
- Added `SourcePath` for CEF files to `CefSharp.Core` and `CefSharp.BrowserSubprocess.Core`

Without this it compiles perfectly fine but `IntelliSense` doesn't suggest anything for CEF related types. They all show up as `error-type`.

### DIFF
--- a/CefSharp.BrowserSubprocess.Core/CefSharp.BrowserSubprocess.Core.vcxproj
+++ b/CefSharp.BrowserSubprocess.Core/CefSharp.BrowserSubprocess.Core.vcxproj
@@ -73,7 +73,7 @@
     <IgnoreImportLibrary>true</IgnoreImportLibrary>
     <OutDir>bin\$(Platform)\$(Configuration)\</OutDir>
     <IntDir>obj\$(Platform)\$(Configuration)\</IntDir>
-    <IncludePath>$(SolutionDir)packages\$(CefSdkVer)\CEF;$(IncludePath)</IncludePath>
+    <SourcePath>$(SolutionDir)packages\$(CefSdkVer)\CEF;$(SourcePath)</SourcePath>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
     <LinkIncremental>true</LinkIncremental>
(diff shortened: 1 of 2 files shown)


amaitland:
`Intellisense` works for me without this change. It can be a little hit and miss. It's usually only a problem when doing a clean checkout, needs a restart after a successful build.

Do you have a reference doc for `IncludePath`?

merceyz:
I just tried: clean checkout, restored nuget packages, build all, restart.
Intellisense still doesn't show anything, with include path set it works instantly.

I don't, no

amaitland:
intellisense1
intellisense2

As it stands this is what I get using `master`, I pretty much get what I would expect. 

When I have a few minutes I'll see what I can find on `IncludePath`. We currently have `AdditionalIncludeDirectories` set which has basically the same value. So I'm curious as to what the difference is.

merceyz:
https://stackoverflow.com/a/6883392/2923845
https://docs.microsoft.com/en-us/previous-versions/visualstudio/visual-studio-2008/t9az1d21(v=vs.90)
https://docs.microsoft.com/en-us/previous-versions/visualstudio/visual-studio-2010/73f9s62w(v=vs.100)

Seems like `AdditionalIncludeDirectories` is for the compiler specifically and that `IncludePath` is for general use

CefSharp.BrowserSubprocess.Core/CefSharp.BrowserSubprocess.Core.vcxproj
```diff
@@ -73,25 +73,29 @@
     <IgnoreImportLibrary>true</IgnoreImportLibrary>
     <OutDir>bin\$(Platform)\$(Configuration)\</OutDir>
     <IntDir>obj\$(Platform)\$(Configuration)\</IntDir>
+    <IncludePath>$(SolutionDir)packages\$(CefSdkVer)\CEF;$(IncludePath)</IncludePath>
```
amaitland: Does using `SourcePath` work for you? Description given in `VS` being `Path to use when searching for source files to use for Intellisense`

```
<SourcePath>$(SolutionDir)packages\$(CefSdkVer)\CEF;$(SourcePath)</SourcePath>
```

merceyz: Yes, same effect.
Should probably switch over to that one then as that's what it's made for

amaitland:
Thanks :+1:

PR #2597, opened 2018-12-09
Refactoring - Move CefCookie conversion to TypeConversion

merceyz:
- Moved conversion of `CefCookie` -> `Cookie` to `CefSharp.Core\Internals\TypeConversion.h`
- Copied `ConvertCefTimeToDateTime` from `CefSharp.BrowserSubprocess.Core\TypeUtils.h` to `CefSharp.Core\Internals\TypeConversion.h` as including the file caused errors

The conversion of cookies was duplicated in
- `CefSharp.Core\Internals\ClientAdapter.cpp`
- `CefSharp.Core\Internals\CookieVisitor.cpp`
- `CefSharp.Core\ResourceHandlerWrapper.cpp`

### DIFF
--- a/CefSharp.Core/ResourceHandlerWrapper.cpp
+++ b/CefSharp.Core/ResourceHandlerWrapper.cpp
@@ -7,6 +7,7 @@
 #include "Internals/CefRequestWrapper.h"
 #include "Internals/CefResponseWrapper.h"
 #include "Internals/CefCallbackWrapper.h"
+#include "Internals/TypeConversion.h"
 #include "ResourceHandlerWrapper.h"
 
 using namespace System::Runtime::InteropServices;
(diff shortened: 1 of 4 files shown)


CefSharp.Core/ResourceHandlerWrapper.cpp
```diff
@@ -64,49 +65,6 @@ namespace CefSharp
...
-                lastAccess.second,
-                lastAccess.millisecond);
-        }
-
-        return cookie;
+        return TypeConversion::FromNative(cefCookie);
     }
-}
\ No newline at end of file
+}
```
amaitland: I though all the line endings would have been normalised as part of #2504?

merceyz: Yeah, it should have.
For some stupid reason Visual Studio doesn't apply the `insert_final_newline` rule unless you open all files and save them manually.
I'll write a quick program to fix this.

amaitland: Only if you really want to. Fine to leave as is.

merceyz: So you're fine with this PR getting merged?

merceyz: > Only if you really want to. Fine to leave as is.

Rather not have these sorts of changes show up in unrelated commits/PR in the future

PR #2605, opened 2018-12-14
Cef.Get* - Added initialize checks to Cef.Get* methods

### LINKED ISSUE #2600: In Cef.Getxxxx throw Exception if not Cef.IsInitialized
There are a number of methods in the `Cef` class that can only be used once `CEF` has been initialized. Add some sanity checks and throw exception if Initialize has not been called.

- [ ] `GetGlobalCookieManager`
- [ ] `GetGlobalRequestContext`
- [ ] `GetBlockingCookieManager`
- [ ] `GetPlugins`

There are probably others.

Hopefully this reduces support requests.

Inspired by https://pastebin.com/G7r1eJEp

merceyz:
Resolves #2600 

- Added initialized check to `GetGlobalCookieManager`, `GetGlobalRequestContext`, `GetBlockingCookieManager`, and `GetPlugins`

The reason a macro was used instead of a function is so that the exception can contain the method name.

Calling `Cef.GetGlobalCookieManager();` before `Cef.Initialize()` causes an exception with this message to be thrown: `CefSharp::Cef::GetGlobalCookieManager requires that CEF has been initialized first.`

### DIFF
--- a/CefSharp.Core/Cef.h
+++ b/CefSharp.Core/Cef.h
@@ -32,6 +32,8 @@ namespace CefSharp
 {
     public ref class Cef sealed
     {
+        #define InitializedCheck() if (!_initialized) throw gcnew Exception(__FUNCTION__ + " requires that CEF has been initialized first.")
+
     private:
         static Object^ _sync;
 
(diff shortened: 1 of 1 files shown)


amaitland:
Thanks 👍

There are likely more methods that can have this check in the `Cef` class and I was going to include details of when you can start accessing `RequestContext/CookieManager/etc`.

amaitland:
Follow up commit https://github.com/cefsharp/CefSharp/commit/65408138493030f64563f4acbfa38a0fe0f5a1cf

PR #2613, opened 2018-12-27
Add Chromium Extension Support

amaitland:
- [x] Add IRequestContext.LoadExtension
- [x] Add IExtensionHandler
- [x] Add IExtension

TODO:
- [x] Invoke `LoadExtension` on `UI` Thread
- [x] Fix `BrowserSettings` not being passed correctly
- [x] Finish xml doc rewrite
- [x] Add example (import from cefclient)


amaitland:
Commit https://github.com/cefsharp/CefSharp/pull/2613/commits/29d5c2dace3f6c6786a864eb4eb089a1bdba28e8 has been reverted, it builds and runs fine using `VS2017`, fails when building with `VS2015`. Will need to investigate further. 

Will merge this and resolve later as is a nice to have, not a deal breaker when using `Chrome Extensions`.

PR #2637, opened 2019-02-01
WinForms avoid throwing exception when setting BrowserSettings to null

mbragg12:


amaitland:
Quick fix for #2638 

Will go with this for the `cefsharp/71` branch, will standardise on using the private field going forward.

PR #2640, opened 2019-02-02
Build.ps1 - Automated more tasks relating to updating CefSharp version and copyright year

merceyz:
build.ps1 changes:
- Updates the version number in `CefSharp.shfbproj`
- Updates the version number in `appveyor.yml`
- Updates copyright year on compiled files to the current year, same as CEF does
- Specifies `utf-8` as the encoding when reading files
- Saves files as UTF-8 without BOM
- build.ps1 is now saved as `utf-8-bom`, otherwise it breaks on windows when containing non-ASCII characters (©)

Editorconfig:
- Specified `utf-8-bom` for `*.ps1` files

Maintenance:
- LICENSE: Removed the copyright year as it's not needed
- NuGet: Removed the copyright year on the `nuspec` files as it's not needed

Was going to make build.ps1 update the year on LICENSE and NuGet as well, but the years for NuGet were unnecessary and LICENSE should only change year when it's updated.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-// Copyright © 2017 The CefSharp Authors. All rights reserved.
+// Copyright © The CefSharp Authors. All rights reserved.
 //
 // Redistribution and use in source and binary forms, with or without
 // modification, are permitted provided that the following conditions are
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-// Copyright © 2010-2018 The CefSharp Authors. All rights reserved.
+// Copyright © 2017 The CefSharp Authors. All rights reserved.
 //
 // Redistribution and use in source and binary forms, with or without
 // modification, are permitted provided that the following conditions are
--- a/.editorconfig
+++ b/.editorconfig
@@ -18,14 +18,18 @@ charset                  = utf-8
 end_of_line              = crlf
 trim_trailing_whitespace = true
 
+# Powershell files (build.ps1)
+[*.ps1]
+charset = utf-8-bom
+
 # Xml config files
 [*.{props,targets,config,nuspec,manifest}]
-indent_size              = 2
+indent_size = 2
 
 # C++ Files
 [*.{cpp,h,in}]
-curly_bracket_next_line  = true
-indent_brace_style       = Allman
+curly_bracket_next_line = true
+indent_brace_style      = Allman
 
 [*.cs]
 # Capitalization styles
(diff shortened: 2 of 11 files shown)


merceyz:
@amaitland Looking at a350f963f4ea744066fbbac3274b33e2c2406f1b, it seems you want to keep the year. I can update the script to do this instead of removing them if you want.

amaitland:
> Looking at [a350f96](https://github.com/cefsharp/CefSharp/commit/a350f963f4ea744066fbbac3274b33e2c2406f1b), it seems you want to keep the year. I can update the script to do this instead of removing them if you want.

@merceyz I did a little additional reading on the subject and I'm now more confused than ever. I see articles claiming that no copyright year is required, and when I look at assemblies from say `Microsoft` they don't have a `Year`. When I look at most open source projects they appear to have a `year` range. I don't know of having a corporation/company makes a difference and it probably depends on the country. `CEF` only a single year in the binaries themselves. `Google` themselves just appear to have a single year, so I guess going with a single year is acceptable.

amaitland:
I'm happy to take on the task of updating this. I'll probably also add the option to generate the `Sandcastle API` doc as part of the `build.ps1` file to take advantage of the auto updating 👍

merceyz:
> I'm happy to take on the task of updating this.

What exactly do you want changed in this PR?
I'll update it, no problem.

> I'll probably also add the option to generate the `Sandcastle API` doc as part of the `build.ps1` file to take advantage of the auto updating

That should probably be it's own thing

LICENSE
```diff
@@ -1,4 +1,4 @@
-// Copyright © 2010-2018 The CefSharp Authors. All rights reserved.
+// Copyright © 2017 The CefSharp Authors. All rights reserved.
```
amaitland: Should this be 2019?

amaitland:
Just resolving the merge conflicts which it appears you've already done.

LICENSE
merceyz: c12b2392184a49309c4bfcb8b46db8cf8c07e9fb

I was a little bit uncertain on this one so I set it to the last time it was actually modified.
However, it seems like this year can be removed as well

merceyz: The year has now been removed

merceyz:
@amaitland is there anything else to do on this PR or is it fine if I merge it?

amaitland:
I haven't had a chance to run it locally yet. 

I'm ok if you merge it now as we're early in the development cycle,  tweaks can be made later if required :+1:

PR #2651, opened 2019-02-15
Improved dispose pattern in ChromiumWebBrowser

### LINKED ISSUE #1954: Code Cleanup - Improve ChromiumWebBrowser Dispose/Finalise
Revise the `Dispose(bool)` method in the three `ChromiumWebBrowser` classes. As time has gone one the code has been refactored and doesn't quite follow the correct pattern (only dispose of `unmanaged` resources when `disposing = false`).

- Implement a thread safe disposed pattern in all three instances see https://blogs.msdn.microsoft.com/blambert/2009/07/24/a-simple-and-totally-thread-safe-implementation-of-idisposable/ (the `WPF` version differs slightly, better to make them all the same)
- Only dispose of `managed` resources when `disposing = true`. 
- Ordering of disposal is important so keep the ordering the same

Reference: https://msdn.microsoft.com/en-us/library/b1yfkh5e(v=vs.110).aspx

merceyz:
- Added an `IsDisposed` property to `IWebBrowser`
- Implemented a thread safe dispose pattern
- Now only disposes of managed objects when `disposing` is set to `true`
- Simplified the handling of design mode and disposing

Resolves #1954 

EDIT:
As this is getting merged in parts, here are the relevant commits
OffScreen: https://github.com/cefsharp/CefSharp/commit/58fac3865b46c0b55807f670298d0eb070a964ce https://github.com/cefsharp/CefSharp/commit/b1ef9e8bc2c55a5e0e9a0b66988f670695836177

### DIFF
--- a/CefSharp.OffScreen/ChromiumWebBrowser.cs
+++ b/CefSharp.OffScreen/ChromiumWebBrowser.cs
@@ -53,7 +53,7 @@ namespace CefSharp.OffScreen
         /// <summary>
         /// Gets a value indicating whether this instance is disposed.
         /// </summary>
-        /// <value><c>true</c> if this instance is disposed; otherwise, <c>false</c>.</value>
+        /// <value><see langword="true" /> if this instance is disposed; otherwise, <see langword="false" />.</value>
         public bool IsDisposed
         {
             get
--- a/CefSharp.OffScreen/ChromiumWebBrowser.cs
+++ b/CefSharp.OffScreen/ChromiumWebBrowser.cs
@@ -5,6 +5,7 @@
 using System;
 using System.Drawing;
 using System.Drawing.Imaging;
+using System.Threading;
 using System.Threading.Tasks;
 using CefSharp.Enums;
 using CefSharp.Internals;
@@ -42,6 +43,24 @@ namespace CefSharp.OffScreen
         /// Flag to guard the creation of the underlying offscreen browser - only one instance can be created
         /// </summary>
         private bool browserCreated;
+
+        /// <summary>
+        /// The value for disposal, if it's 1 (one) then this instance is either disposed
+        /// or in the process of getting disposed
+        /// </summary>
+        private int disposeSignaled;
+
+        /// <summary>
+        /// Gets a value indicating whether this instance is disposed.
+        /// </summary>
+        /// <value><c>true</c> if this instance is disposed; otherwise, <c>false</c>.</value>
+        public bool IsDisposed
+        {
+            get
+            {
+                return Interlocked.CompareExchange(ref disposeSignaled, 1, 1) == 1;
+            }
+        }
 
         /// <summary>
         /// A flag that indicates whether the WebBrowser is initialized (true) or not (false).
@@ -325,6 +344,23 @@ namespace CefSharp.OffScreen
         /// <param name="disposing"><c>true</c> to release both managed and unmanaged resources; <c>false</c> to release only unmanaged resources.</param>
         protected virtual void Dispose(bool disposing)
         {
+            if (Interlocked.CompareExchange(ref disposeSignaled, 1, 0) != 0)
+            {
+                return;
+            }
+
+            if (disposing)
+            {
+                browser = null;
+                IsBrowserInitialized = false;
+
+                if (managedCefBrowserAdapter != null)
+                {
+                    managedCefBrowserAdapter.Dispose();
+                    managedCefBrowserAdapter = null;
+                }
+            }
+
             // Don't reference event listeners any longer:
             AddressChanged = null;
             BrowserInitialized = null;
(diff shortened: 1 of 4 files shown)


CefSharp.WinForms/ChromiumWebBrowser.cs
```diff
@@ -431,79 +451,92 @@ private void InitializeFieldsAndCefIfRequired()
...
-        protected override void Dispose(bool disposing)
+        ~ChromiumWebBrowser()
+        {
+            Dispose(false);
+        }
+
+        /// <summary>
+        /// Releases all resources used by the <see cref="ChromiumWebBrowser"/> object
+        /// </summary>
+        public new void Dispose()
```
amaitland: What exactly is the reason for newing over the existing implementation? 

https://referencesource.microsoft.com/#System/compmod/system/componentmodel/Component.cs,118

merceyz: I did it to have the `summary` be the same for all 3 versions (WinForms, OffScreen, and WPF)

amaitland: Any objection to removing it? Whilst it's somewhat harmless I'd rather avoid using **new** unless there's a really compelling reason, I just look at it and think code smell, and puzzle why it's there. Maybe it's just me....

merceyz: Not at all, removed both the `Dispose` and `Finalizer` since their function is already taken care of by the base class

amaitland: Thanks 👍

CefSharp.OffScreen/ChromiumWebBrowser.cs
```diff
@@ -320,24 +339,15 @@ public void Dispose()
...
-            FrameLoadEnd = null;
-            FrameLoadStart = null;
-            LoadError = null;
-            LoadingStateChanged = null;
-            Paint = null;
-            StatusMessage = null;
-            TitleChanged = null;
-
-            Cef.RemoveDisposable(this);
+            if (Interlocked.CompareExchange(ref disposeSignaled, 1, 0) != 0)
```
amaitland: I'm thinking we need a comment for this line in the three instances, as it does more than just check if the count is 0 or 1.

Thoughts?

merceyz: I feel it explains itself, the method is called CompareExchange and if the user hovers over it they can see exactly what it does (compare and if equal, exchange + return original value).

merceyz:
> Any objection if I merge this one class at a time?

I would like to keep these changes together and not split them up, however it's up to you.

> Rather than trying to review and test everything all at once, and potentially having this go stale, I think it makes sense to work through them one by one. `OffScreen`, `WPF` and `WinForms` is probably the order of lest to most complex.

The code in all three is basically the same, only difference is `WPF` and `WinForms` moves the actual disposing out of the `protected virtual void Dispose(bool disposing)` so that the designer stays happy

CefSharp.OffScreen/ChromiumWebBrowser.cs
amaitland: For anyone unfamiliar with `CompareExchange` it's kinda cryptic. It maybe super clear now, if you come back in 6-12 months it might not be so clear

At a cursory glance even I though, can this be changed to use the `IsDisposed` property.

I'll add a comment later.

merceyz: Yeah, you're right.

Sure

amaitland:
> I would like to keep these changes together and not split them up, however it's up to you.

It'll be some time before I have a chance to review them all at once. What advantage do you see in trying to merge this all at once? They are unrelated changes from a code point of view.

> The code in all three is basically the same, only difference is `WPF` and `WinForms` moves the actual disposing out of the `protected virtual void Dispose(bool disposing)` so that the designer stays happy

It's the sort of thing that's easy to introduce a subtle change, the structure looks fine.

merceyz:
> It'll be some time before I have a chance to review them all at once. What advantage do you see in trying to merge this all at once? They are unrelated changes from a code point of view.

Less noise in the history, but mainly because i'd like to make more changes to all three at the same time for other things.

amaitland:
> changes to all three at the same time for other things

What sort of things exactly? Related to `Dispose` improvements?

amaitland:
`OffScreen` changes merged in https://github.com/cefsharp/CefSharp/commit/58fac3865b46c0b55807f670298d0eb070a964ce

Addition comment added in https://github.com/cefsharp/CefSharp/commit/b1ef9e8bc2c55a5e0e9a0b66988f670695836177

merceyz:
> What sort of things exactly? Related to `Dispose` improvements?

First thing that comes to mind that is sort of related is a disposed check in all public functions.

amaitland:
Please keep the scope of this `PR` limited to resolving #1954

merceyz:
Of course

CefSharp.WinForms/ChromiumWebBrowser.cs
```diff
@@ -431,79 +451,77 @@ private void InitializeFieldsAndCefIfRequired()
...
-                managedCefBrowserAdapter.Dispose();
-                managedCefBrowserAdapter = null;
+                if (managedCefBrowserAdapter != null)
+                {
+                    managedCefBrowserAdapter.Dispose();
+                    managedCefBrowserAdapter = null;
+                }
+
+                // Don't maintain a reference to event listeners anylonger:
+                AddressChanged = null;
```
amaitland: Was moving this block to lower down in the order a deliberate change? Just checking in case we need to add an additional comment.

merceyz: Setting event listeners to null? Yes, I did it so that they're all in the same place in all implementations.

amaitland:
Will hopefully review/merge the `WPF` changes in the next few days.

amaitland:
@merceyz Thoughts on a `Breaking Changes` entry for this one? I'm struggling to think of what changes users would need to make, just a slight change in behaviour. Maybe you can think of something.

merceyz:
@amaitland The `IsDisposed` property added to `IWebBrowser` would break builds for anyone that implements that interface on their own

PR #2654, opened 2019-02-17
Fixed message interceptor not working after a render process switch

### LINKED ISSUE #2579: Message interceptor no longer intercepts WinForms messages
    - What version are you using? Nuget 69.0.0

- **What architecture x86 or x64?** x64 (possibly happens with x86, but I haven't tested it)

    - Win10

- **Are you using `WinForms`, `WPF` or `OffScreen`?** `WinForms`

Using `CefSharp.WinForms.Example`:  Build and run the example project
Navigate to "google.com" (or pretty much any other address)
After the page has finished loading open the "Zoom Level" menu and then click in the browser area.

Expected: Menu closes since you clicked elsewhere
Result: Menu stays open

    - Does the cef log provide any relevant information? Doesn't look like it.

    - Any other background information that's relevant? Are you doing something out of the ordinary? 3rd party controls? Nope

Nope (cef_binary_3.3538.1852.gcb937fc_windows64_client.tar.bz2)

Note: I have noticed that the following logic does not get hit in the broken state:
https://github.com/cefsharp/CefSharp/blob/485297bb7f149a708480165130b3319382f7846e/CefSharp.WinForms.Example/BrowserTabUserControl.cs#L250

merceyz:
- Made the `ChromeWidgetMessageInterceptor` set itself up again after a render process switch happens

Fixes #2579

### DIFF
--- a/CefSharp.WinForms.Example/BrowserTabUserControl.cs
+++ b/CefSharp.WinForms.Example/BrowserTabUserControl.cs
@@ -6,7 +6,6 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Runtime.InteropServices;
-using System.Threading;
 using System.Threading.Tasks;
 using System.Windows.Forms;
 using CefSharp.Example;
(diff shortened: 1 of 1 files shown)


amaitland:
I haven't had a chance to run it yet, just looking at the code, looks like a very neat and tidy solution 👍 

It's probably worth adding some background comments as to what's going on, that shouldn't necessarily stop this from being merged.

CefSharp.WinForms.Example/BrowserTabUserControl.cs
```diff
@@ -232,71 +231,90 @@ private void OnIsBrowserInitializedChanged(object sender, IsBrowserInitializedCh
...
+                            messageInterceptor = new ChromeWidgetMessageInterceptor((Control)Browser, chromeWidgetHostHandle, message =>
                             {
-                                IntPtr chromeWidgetHostHandle;
-                                if (ChromeWidgetHandleFinder.TryFindHandle(browserHandle, out chromeWidgetHostHandle))
+                                const int WM_MOUSEACTIVATE = 0x0021;
+                                const int WM_NCLBUTTONDOWN = 0x00A1;
+                                const int WM_LBUTTONDOWN = 0x0201;
+                                const int WM_DESTROY = 0x0002;
+
+                                // Render process switch happened, need to find the new handle
```
merceyz: Anything more than this one?
You mean a `///<summary>` or?

amaitland: I think a summary of what's going on would be helpful. I can add that later if you'd like

merceyz: Yeah, a followup commit would probably be the best

amaitland:
Added an extra comment in https://github.com/cefsharp/CefSharp/commit/574bc16db23e3bb2df1f8cf94abaf6d3fb47603c

PR #2659, opened 2019-02-22
Readme.md - Add SharpBrowser

robinrodricks:
I've added SharpBrowser to the readme. The last time you mentioned that it was not using nuget, so I fixed that. Its now taking the latest version from nuget. I just included the `bin` directory in the repo because sometimes CefSharp has issues downloading all the needed assets. I've also updated to CefSharp 71 and .NET 4.6.

Could you consider it please?


amaitland:
Sure 👍

robinrodricks:
Thank you!!

PR #2669, opened 2019-03-02
Core - BrowserSettings: Fixed a memory leak

merceyz:
When a `BrowserSettings` object is created, it creates an instance of `CefBrowserSettings`, then when the `BrowserSettings` object is disposed/collected it never deletes the `CefBrowserSettings` instance meaning it leaks. This PR fixes that by deleting it when `BrowserSettings` is disposed/collected.

Before this change, creating and disposing `1 000 000` `BrowserSettings` objects caused the memory usage to climb to `337,71 MB`, after this change it goes to `99,32 MB`

### DIFF
--- a/CefSharp.Core/BrowserSettings.h
+++ b/CefSharp.Core/BrowserSettings.h
@@ -41,7 +41,12 @@ namespace CefSharp
 
         !BrowserSettings()
         {
-            _browserSettings = NULL;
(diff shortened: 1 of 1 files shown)


amaitland:
Saw your initial commit, before I had a chance to comment that you were releasing the `CEF` allocated `pointer` you'd fixed it 😄 

Thanks!

PR #2670, opened 2019-03-02
Core - WindowInfo: Fixed a memory leak

merceyz:
When a `WindowInfo` object is created, it creates an instance of `CefWindowInfo`, then when the `WindowInfo` object is disposed/collected it never deletes the `CefWindowInfo` instance meaning it leaks. This PR fixes that by deleting it when `WindowInfo` is disposed/collected.

Before this change, creating and disposing `1 000 000` `WindowInfo` objects caused the memory usage to climb to `130,58 MB`, after this change it goes to `99,32 MB`

### DIFF
--- a/CefSharp.Core/WindowInfo.h
+++ b/CefSharp.Core/WindowInfo.h
@@ -14,10 +14,12 @@ namespace CefSharp
     {
     private:
         CefWindowInfo* _windowInfo;
+        bool _ownsPointer = false;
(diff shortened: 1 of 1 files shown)


amaitland:
At this point in time I'm planning on merging this and #2669 into `cefsharp/71` and releasing as part of `71.0.1`. Will update if that plan changes.

PR #2683, opened 2019-03-09
Refactoring - Remove reference to ChromiumWebBrowser in DefaultRenderHandler

merceyz:
- Added a `ViewportSize` property to `IRenderHandler` and `DefaultRenderHandler`
- Removed reference to `ChromiumWebBrowser` in the `DefaultRenderHandler`
- Removed `size` field in `ChromiumWebBrowser`
- The `Size` property in `ChromiumWebBrowser` now uses the size from the `RenderHandler`

### DIFF
--- a/CefSharp.OffScreen/IRenderHandler.cs
+++ b/CefSharp.OffScreen/IRenderHandler.cs
@@ -15,6 +15,12 @@ namespace CefSharp.OffScreen
     public interface IRenderHandler : IDisposable
     {
         /// <summary>
+        /// Size of the Chromium viewport.
+        /// This must be set to something other than 0x0 otherwise Chromium will not render.
+        /// </summary>
+        System.Drawing.Size ViewportSize { get; set; }
+
+        /// <summary>
         /// Called to allow the client to return a ScreenInfo object with appropriate values.
         /// If null is returned then the rectangle from GetViewRect will be used.
         /// If the rectangle is still empty or invalid popups may not be drawn correctly. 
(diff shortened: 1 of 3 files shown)


CefSharp.OffScreen/IRenderHandler.cs
```diff
@@ -14,6 +14,12 @@ namespace CefSharp.OffScreen
     /// </summary>
     public interface IRenderHandler : IDisposable
     {
+        /// <summary>
+        /// Size of the Chromium viewport.
+        /// This must be set to something other than 0x0 otherwise Chromium will not render.
+        /// </summary>
+        System.Drawing.Size ViewportSize { get; set; }
```
amaitland: I'd prefer to keep the implementation detail from being added to the interface, the `CefRenderHandlerWrapper` won't use this property, so really shouldn't be added here. I know this complicates things slightly.

Also the dream is that one day I'll move the `IRenderHandler` implementation into the `CefSharp` project, and share it between `OffScreen` and `WPF` (the `WPF` implementation needs a major rethink to make this happen).

amaitland: > the `CefRenderHandlerWrapper` won't use this property



Correction, `RenderClientAdapter.h`.

alfein: Sorry to bother you, @amaitland, can you, please let me know what's required to sponsor a release?

What's the contribution size are you looking for?

Thx

amaitland: @alfein You can contact using https://gitter.im/amaitland to discuss your requirements.

Anyone with a `GitHub` account can signup in a few clicks if you don't have an account.

If that doesn't work then let me know.

PR #2691, opened 2019-03-15
WPF - Implement UpdateDragCursor

joaompneves:
Implemented the UpdateDragCursor method of the WPF webview.

Tested with: https://jsfiddle.net/tw5f801j/

The drag&drop cursors are retrieved from Windows lib (ole32) and displayed when appropriate.

The implementation still has a minor problem: the cursor flickers, which is more noticeable when the cursor changes to forbidden. I don't know why this happens. If anyone can shed light, that would be much appreciated.


CefSharp.Wpf/ChromiumWebBrowser.cs
```diff
@@ -787,7 +787,8 @@ void IRenderWebBrowser.UpdateDragCursor(DragOperationsMask operation)
         /// <param name="operation">describes the allowed operation (none, move, copy, link). </param>
         protected virtual void UpdateDragCursor(DragOperationsMask operation)
         {
-            //TODO: Someone should implement this
+            var dragCursor = DragCursorProvider.GetCursor(operation);
+            UiThreadRunSync(() => Mouse.SetCursor(dragCursor));
```
amaitland: Is there a reason this is run `Sync`? If so please add a comment as to the reason.

CefSharp.Wpf/DragCursorProvider.cs
```diff
@@ -0,0 +1,56 @@
+﻿using System;
```
amaitland: Needs a license added to the header.

CefSharp.Wpf/DragCursorProvider.cs
```diff
@@ -0,0 +1,56 @@
...
+        private static extern IntPtr LoadCursor(IntPtr hInstance, ushort lpCursorName);
+        
+        private static readonly Dictionary<DragOperationsMask, Cursor> DragCursors;
+
+        static DragCursorProvider() 
+        {
+            var library = LoadLibrary("ole32.dll");
+            DragCursors = new Dictionary<DragOperationsMask, Cursor>()
+            {
+                { DragOperationsMask.None, GetCursorFromLib(library, 1) },
```
amaitland: What's the difference between this `None` cursor and `Cursors.None`? Be better to use the built in one where possible

https://docs.microsoft.com/en-us/dotnet/api/system.windows.input.cursors.none?view=netframework-4.7.2

CefSharp.Wpf/DragCursorProvider.cs
```diff
@@ -0,0 +1,56 @@
...
+                { DragOperationsMask.Move, GetCursorFromLib(library, 2) },
```
amaitland: Is `Move` substantially different from `SizeAll`? Be better to use the built in cursor where possible.

https://docs.microsoft.com/en-us/dotnet/api/system.windows.input.cursors.sizeall?view=netframework-4.7.2#System_Windows_Input_Cursors_SizeAll

amaitland:
> The implementation still has a minor problem: the cursor flickers

How does it behave if you just map the different types to a built in cursor? Is it something to do with the custom cursor?

joaompneves:
> Please provide a list of steps used for testing the different cursor types.

Changed the link with a more self-explanatory example. Just drag the Drag me box over the others.

CefSharp.Wpf/ChromiumWebBrowser.cs
joaompneves: This is an attempt to reduce the flicker, but it still exists.

CefSharp.Wpf/DragCursorProvider.cs
joaompneves: Cursors.None is an invisible cursor as the None in drag&drop is a forbidden one.

CefSharp.Wpf/DragCursorProvider.cs
joaompneves: Drag&drop cursors don't exist in the built-in collection.

These are the ones we get from ole32:
https://i.stack.imgur.com/TffLL.png

joaompneves:
> > The implementation still has a minor problem: the cursor flickers
> 
> How does it behave if you just map the different types to a built in cursor? Is it something to do with the custom cursor?

Behaves the same way.

joaompneves:
Solved the flicker problem. The feature is now OK.

joaompneves:
Turns out the solution is more simple. Just had to store the drag operation result on the update drag cursor and use that in the DragOver event. I have tried this before without success, because the key is to handle the drag over event, otherwise the cursor gets overridden by wpf (default behavior).

amaitland:
Much cleaner thanks 👍 If all goes well testing wise this will be in the `73` release.

amaitland:
Formatting fixes for the html file in https://github.com/cefsharp/CefSharp/commit/ea843f31c172dcd7da4b11d9330b9f9cb2fd8f28

`&` needs to be encoded as `&amp;` for xml doc also

Issue #2662, opened 2019-02-25
IJavascriptCallback.ExecuteAsync with DateTime as param produces incorrect JavaScript Date


kurtzky:
- **What version of the product are you using?**
    - 69.0.0

- **What architecture x86 or x64?**
    - x86

- **On what operating system?**
    - Win7

- **Are you using `WinForms`, `WPF` or `OffScreen`?**
    - WPF

- **What steps will reproduce the problem?**
These are just code snippets that show the overall design of the code.

**Class1.cs**
```
private ChromiumWebBrowser webBrowser;
private WebInterface webInterface;

private void CreateCef()
{
	webBrowser = new ChromiumWebBrowser();
	webInterface = new WebInterface(webBrowser);
	webBrowser.RegisterAsyncJsObject("webInterface", webInterface);	
}

//This will be called whenever we receive data from server
private void OnNotificationReceived(DateTime value)
{
	webInterface.InvokeNotification(value);
}

```

**WebInterface.cs**
```
private readonly ChromiumWebBrowser WebBrowser;
private IJavascriptCallback Callback;

public WebInterface(ChromiumWebBrowser browser) 
{
	WebBrowser = browser;
}

public void SubscribeEvent(IJavascriptCallback onCallback)
{
	Callback = onCallback;
}

public void InvokeNotification(object data)
{
	Callback?.ExecuteAsync(data); //data sent is 01/01/2018 12:00:00
}
```

**JsInterface.ts**
```
export class JsInterface {
	public OnNotify: (data: any) => void;
	
	constructor() {
	}

	public Initialize(): void {
		webInterface.SubscribeEvent((data) => {
            		if (this.OnNotify != null) {
                		this.OnNotify(data);
            		}
        	});
	}
}
```

**MainUi.ts**

```
//this is the Javascript side that receives data from server
export class MainUi {
	private _userInterfaceInstance: any;

	constructor() {
    	}

	public Initialize()
	{
		this._userInterfaceInstance = new JsInterface();
	        this._userInterfaceInstance.Initialize();
		this._userInterfaceInstance.OnNotify = OnCallback;
	}

	function OnCallback(data: any): void {
		 //data received is 01/01/2018 12:00:25
	}
}

```
- **What is the expected output? What do you see instead?**
Expected: 01/01/2018 12:00:00
Output: 01/01/2018 12:00:25

[...]

PR #2695, opened 2019-03-19
Added QUnit tests for DateTime serialization

kurtzky:


amaitland:
A detailed description is required, including the issue that this `PR` relates to.

kurtzky:
Created QUnit tests to cover other DateTime values. Related to Issue #2662.

amaitland:
When I ran the tests locally they passed, which is kinda what I was expecting.

I've reopened #2662 though without failing test cases there isn't a lot I can do.

Issue #2688, opened 2019-03-15
Feature Request - Add IRenderProcessHandler.OnContextCreated/Released


amaitland:
Allow additional interfaces and wrappers to allow for execution of `JavaScript` directly in the render process by adding

- `IRenderProcessHandler.OnContextCreated`
- `IRenderProcessHandler.OnContextReleased`

https://magpcss.org/ceforum/apidocs3/projects/(default)/CefRenderProcessHandler.html

Additional wrappers/interfaces will be required for:
- `CefV8Context`
  - [ ] `Enter`
  - [ ]  `Eval`
  - [x]  `Execute` (Calls `Eval` without returning a result)
  - [ ]  `Exit`
  - [ ]  `GetGlobal`
  - [ ]  `IsValid`
- `CefV8Value`
  - [ ] AdjustExternallyAllocatedMemory
  - [ ] ClearException
  - [ ] CreateArray
  - [ ] CreateArrayBuffer
  - [ ] CreateBool
  - [ ] CreateDate
  - [ ] CreateDouble
  - [ ] CreateInt
  - [ ] CreateNull() 
  - [ ] CreateObject( won't have accessor/interceptor constructor arguments initially) 
  - [ ] CreateString
  - [ ] CreateUInt
  - [ ] CreateUndefined
  - [ ] DeleteValue
  - [ ] DeleteValue
  - [ ] ExecuteFunction
  - [ ] GetArrayBufferReleaseCallback
  - [ ] GetArrayLength
  - [ ] GetBoolValue
  - [ ] GetDateValue
  - [ ] GetDoubleValue
  - [ ] GetException
  - [ ] GetExternallyAllocatedMemory
  - [ ] GetFunctionName
  - [ ] GetIntValue
  - [ ] GetKeys
  - [ ] GetStringValue
  - [ ] GetUIntValue
  - [ ] GetValue
  - [ ] GetValue
  - [ ] HasException
  - [ ] HasValue
  - [ ] HasValue
  - [ ] IsArray
  - [ ] IsArrayBuffer
  - [ ] IsBool
  - [ ] IsDate
  - [ ] IsDouble
  - [ ] IsFunction
  - [ ] IsInt
  - [ ] IsNull
  - [ ] IsObject
  - [ ] IsSame
  - [ ] IsString
  - [ ] IsUInt
  - [ ] IsUndefined
  - [ ] IsUserCreated
  - [ ] IsValid
  - [ ] NeuterArrayBuffer
[...]

PR #2697, opened 2019-03-19
Add Initial IRenderProcessHandler.OnContextCreated/Released implementation

amaitland:
Currently only `IV8Context.Execute` is implemented, you can execute javascript without returning a result.

Can **only** be called in the render process.

Issue #2688


amaitland:
Demo of using your own application exe as the `BrowserSubProcess` is at https://github.com/cefsharp/CefSharp.MinimalExample/commit/05dc8ad9278584c0d5f8cf7d6d500fd1358a173d

CefSharp.BrowserSubprocess.Core/Wrapper/Browser.cpp
```diff
@@ -0,0 +1,221 @@
+// Copyright © 2015 The CefSharp Authors. All rights reserved.
```
merceyz: Should probably be 2019

CefSharp.BrowserSubprocess.Core/Wrapper/Browser.h
```diff
@@ -0,0 +1,188 @@
+// Copyright © 2015 The CefSharp Authors. All rights reserved.
```
merceyz: Should probably be 2019

CefSharp.BrowserSubprocess.Core/Wrapper/Frame.cpp
```diff
@@ -0,0 +1,296 @@
+// Copyright © 2015 The CefSharp Authors. All rights reserved.
```
merceyz: Should probably be 2019

CefSharp.BrowserSubprocess.Core/Wrapper/Frame.h
```diff
@@ -0,0 +1,253 @@
+// Copyright © 2015 The CefSharp Authors. All rights reserved.
```
merceyz: Should probably be 2019

CefSharp.BrowserSubprocess.Core/Wrapper/Browser.cpp
amaitland: Thanks 👍 

Fixed in https://github.com/cefsharp/CefSharp/commit/3731dfa9311c78375581daf435aef4ea5e9eb14b

PR #2701, opened 2019-03-22
Reorder dispose pattern in ChromiumWebBrowser

amaitland:
WinForms/Wpf/OffScreen - Reorder Dispose methods
- Set browser initialized to false
- Release event handlers
- Dispose managedCefBrowserAdapter
- Set handlers to null

Follow up to #2651

### DIFF
--- a/CefSharp.OffScreen/ChromiumWebBrowser.cs
+++ b/CefSharp.OffScreen/ChromiumWebBrowser.cs
@@ -367,6 +367,10 @@ namespace CefSharp.OffScreen
                 StatusMessage = null;
                 TitleChanged = null;
 
+                // Release reference to handlers, except LifeSpanHandler which is done after Disposing
+                // ManagedCefBrowserAdapter otherwise the ILifeSpanHandler.DoClose will not be invoked.
+                this.SetHandlersToNullExceptLifeSpan();
+
                 browser = null;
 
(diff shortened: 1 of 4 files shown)


amaitland:
All three controls should be roughly the same in order now. Releasing the event handlers before `Disposing` of `ManagedCefBrowserAdapter` should keep the events from being fired mid `Dispose`.

I'm also considering freeing the handlers earlier and just setting `ILifeSpanHandler = null` after  `ManagedCefBrowserAdapter.Dispose`

merceyz:
> I'm also considering freeing the handlers earlier and just setting ILifeSpanHandler = null after ManagedCefBrowserAdapter.Dispose

Is it an option to set all handlers to null before that call, then have it handle the logic of closing the browser further down the stack?

amaitland:
> Is it an option to set all handlers to null before that call

All handlers except `LifeSpanHandler`. It has to be freed after otherwise `LifeSpanHandler.DoClose` won't be called. See https://github.com/cefsharp/CefSharp/blob/cefsharp/71/CefSharp.Wpf/ChromiumWebBrowser.cs#L642

amaitland:
`LifeSpanHandler` is now set to `null` independently of the other handlers which are set to `null` after the event handlers.

PR #2710, opened 2019-03-30
Refactoring - Remove IsBrowserInitializedChangedEventArgs

merceyz:
- Removed `IsBrowserInitializedChangedEventArgs`

This is a breaking change, anyone listening to the `OnIsBrowserInitializedChanged` event needs to change it's signature from
`void OnIsBrowserInitializedChanged(object sender, IsBrowserInitializedChangedEventArgs args)`
to
`void OnIsBrowserInitializedChanged(object sender, EventArgs e)`


amaitland:
LGTM, thanks 👍 I'm just going to bump this to the `75` milestone as I expect this will create quite a few support requests and I'd like to defer that until the next version, there are quite a few breaking changes included in this release as is.

amaitland:
The` MinimalExample` was updated, see https://github.com/cefsharp/CefSharp.MinimalExample/commit/f0c5d77bfa9363c1d76445e937c07575903641f9 for another example.

PR #2720, opened 2019-04-09
Builds with custom suffix

apla:
Allows to build package with customised version of CefSharp. If `ceflib.dll` also customised, this changeset also allows to write redist version to `packages.config` and then restore required nuget packages

 See also https://github.com/cefsharp/cef-binary/pull/77

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -5,7 +5,10 @@
     [Parameter(Position = 1)]
     [string] $Version = "73.1.12.0",
     [Parameter(Position = 2)]
-    [string] $AssemblyVersion = "73.1.12.0"
+    [string] $AssemblyVersion = "73.1.12.0",
+    [string] $RedistVersion,
+    [string] $Suffix
+
 )
 
 $WorkingDir = split-path -parent $MyInvocation.MyCommand.Definition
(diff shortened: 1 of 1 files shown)


amaitland:
What is the requirement for a suffix here? You can pretty much use any version you like as a param currently. 

Changing the packages.config file is really outside the scope of this build script, should be done in another script,  it's also incomplete as you only update CefSharp.Core. The script is already a bit bloated.

build.ps1
```diff
@@ -5,16 +5,26 @@
...
 $WorkingDir = split-path -parent $MyInvocation.MyCommand.Definition
 $CefSln = Join-Path $WorkingDir 'CefSharp3.sln'
 
 # Extract the current CEF Redist version from the CefSharp.Core\packages.config file
 # Save having to update this file manually Example 3.2704.1418
-$CefSharpCorePackagesXml = [xml](Get-Content (Join-Path $WorkingDir 'CefSharp.Core\Packages.config'))
-$RedistVersion = $CefSharpCorePackagesXml.SelectSingleNode("//packages/package[@id='cef.sdk']/@version").value
+$PackagesXmlFilename = Join-Path $WorkingDir 'CefSharp.Core\Packages.config'
+$CefSharpCorePackagesXml = [xml](Get-Content ($PackagesXmlFilename))
+$RedistVersionNode = $CefSharpCorePackagesXml.SelectSingleNode("//packages/package[@id='cef.sdk']/@version")
```
amaitland: Changing package versions shouldn't happen as part of the build, it should be done prior to executing the script. This also only updates the `CefSharp.Core` package, doesn't take into account `CefSharp.BrowserSubprocess.Core`, so this could easily lead to a broken build.

build.ps1
```diff
@@ -281,21 +291,26 @@ function Nupkg
     if(-not (Test-Path $nuget)) {
         Die "Please install nuget. More information available at: http://docs.nuget.org/docs/start-here/installing-nuget"
     }
+    
+    $NupkgVersion = $Version
+    if($Suffix) {
+        $NupkgVersion = $Version + '-' + $Suffix
```
amaitland: Effectively you are creating a `pre-release` version by adding a `-` (See https://semver.org/#spec-item-9 if your not familiar with `SemVer`).

This is already supported by the build script, just pass in `$Version` that contains the appended suffix.

amaitland:
Closing as I have no plans of merging this. Using a `suffix` is already supported and used for every `-pre` release of `CefSharp`.

Issue #2714, opened 2019-04-01
Feature Request - Add IAudioHandler


amaitland:
`CEF` has added the ability to capture the audio output to a buffer. Relevant commit.

https://bitbucket.org/chromiumembedded/cef/commits/9f41a27e586a6b557fc171d2f03b332c1338e8d2?at=3683

Requires addition of a new `IAudioHandler` interface and ChannelLayout` enum.

amaitland:
Initial PR at #2727 (very much a work in progress)

amaitland:
`IAudioHandler` has been added, there is currently no working example as I ran out of time when writing one. The only area in question for the wrapper implementation is `const float** data` being mapped to `IntPtr` correctly. I believe it's correct though haven't been able to test.

It's important to note that the data received is `RAW PCM` data, you won't be able to write it directly to disk and play it in just any media player as it contains no header information.

amaitland:
The Chromium implementation has changed so dramatically that CEF has removed support see https://bitbucket.org/chromiumembedded/cef/commits/1f893c76a97eaa88e3ff782b8e2e47dd0f5d27b2

Track https://bitbucket.org/chromiumembedded/cef/issues/2755/reimplement-audio-mirroring-for for details on when it will be reimplemented

Was removed from `CefSharp` in https://github.com/cefsharp/CefSharp/commit/802bb7f43c33626c28975ea958777daff0f2e9e1

amaitland:
#3152 will track the progress of re-adding the audio handler.

PR #2727, opened 2019-04-17
Add IAudioHandler

amaitland:
Initial implementation for #2714

Still some work required, this compiles, nothing past that so far. Still need to validate how to map `float**` to a `C#` type. One option would be to copy the data to a `byte[]` in native code, this implies the user is always going to want access to the data, and will have a perf implication if they don't.

The default pattern of `return this` in `ClientAdapter` has been used.

```
virtual DECL CefRefPtr<CefAudioHandler> GetAudioHandler() OVERRIDE { return this; }
```

Unfortunately this will always trigger the `Audio Mirroring` feature, which we don't want unless the user has explicitly provided an `IAudioHandler` implementation. So this will also need to be rewritten with a separate wrapper class.

https://bitbucket.org/chromiumembedded/cef/commits/9f41a27e586a6b557fc171d2f03b332c1338e8d2?at=3683#Llibcef/browser/browser_host_impl.ccT3046

### DIFF
--- a/CefSharp.Core/Internals/ClientAdapter.cpp
+++ b/CefSharp.Core/Internals/ClientAdapter.cpp
@@ -1156,6 +1156,42 @@ namespace CefSharp
             }
         }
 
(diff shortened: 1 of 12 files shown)


amaitland:
AudioMirroring is now only enabled when a handler is assigned

Issue #2722, opened 2019-04-12
NullReferenceException in WPF.Example "Javascript Binding Stress Test" 


michalkania:
Master branch
![image](https://user-images.githubusercontent.com/8491194/56049707-5cf47b00-5d4a-11e9-9443-2c5efcae1ef9.png)

- Win10
- WPF
- Debug x86 (VS2015)

Steps:
1. Build WPF.Example app.
2. Once it started, hit "Javascript Binding Stress Test" at the right.
3. It will start showing off the results. CLOSE THE TAB.
4. Just wait (around 10 seconds). NullReferenceException will occur.

In: **WebBrowserExtensions.cs** file
Method: **public static void Reload(this IWebBrowser browser, bool ignoreCache)**
Line: **var cefBrowser = browser.GetBrowser();**

System.NullReferenceException was unhandled by user code
  HResult=-2147467261
  Message=Object reference not set to an instance of an object.
  Source=CefSharp
  StackTrace:
       at CefSharp.WebBrowserExtensions.Reload(IWebBrowser browser, Boolean ignoreCache) in C:\Users\username\Source\Repos\CefSharp\CefSharp\WebBrowserExtensions.cs:line 596
       at CefSharp.WebBrowserExtensions.Reload(IWebBrowser browser) in C:\Users\username\Source\Repos\CefSharp\CefSharp\WebBrowserExtensions.cs:line 584
       at CefSharp.Wpf.Example.ViewModels.BrowserTabViewModel.<.ctor>b__72_8(Task t) in C:\Users\username\Source\Repos\CefSharp\CefSharp.Wpf.Example\ViewModels\BrowserTabViewModel.cs:line 123
       at System.Threading.Tasks.ContinuationTaskFromTask.InnerInvoke()
       at System.Threading.Tasks.Task.Execute()
  InnerException:

PR #2731, opened 2019-04-18
[WPF.Example] Null check

michalkania:
Simple null check to prevent App crash if we close a tab during a test.

Closes: https://github.com/cefsharp/CefSharp/issues/2722


amaitland:
Thanks,  will merge during the week.

PR #2735, opened 2019-04-21
Cefsharp/71

crazyfroggy:
1 - Method LoadUntilEnd(string url) :
this would be useful, avoiding to re-invent the wheel + the hot water every time i need to take an hot shower when cycling.

2 - Method LoadWithCredentials (string url, string user, string password)
this would be useful, avoiding to re-invent wheel + hot water every time i need to take an hot shower when cycling, the bicycle being padlocked on a street lamp.
Eventually with some option  ......, bool UntilEnd), to mix with the method 1.

Btw, I don't want to merge 40 commits into master.
To change the Cefsharp71 Title (? why this title that i didn't wrote ?) -> something else, would be nice.


amaitland:
Please don't open a pull request unless you have code to contribute to the project.

amaitland:
If you would like to create a feature request then you can do so by creating an issue, if you are new to GitHub then see https://help.github.com/en/articles/creating-an-issue for details. 

Please include the use case,  a basic functional spec and any example/pseudo code you have.  References to hot water, showering and cycling are not relevant, please only include technical details. If you have implemented the features you require already then include the code as a sample.

PR #2749, opened 2019-04-25
IME implementation in WPF

a-marmer:
IME implementation in WPF which supports Japanese, Chinese, Korean languages.
Instruction how to test install it on Windows 10 https://www.coscom.co.jp/learnjapanese801/msime_win10_en1.html


amaitland:
Please remove unnessicary unrelated code changes. There are quite a few at first glance. 

What versions of Windows have you tested on?

a-marmer:
Which are the unrelated code changes?
It is tested on Windows 7, Windows 10.

CefSharp.Example/Filters/FindReplaceResponseFilter.cs
```diff
@@ -36,11 +36,6 @@ public class FindReplaceResponseFilter : IResponseFilter
         /// </summary>
         private List<byte> overflow = new List<byte>();
 
-        /// <summary>
```
amaitland: Revert this

CefSharp.WinForms.Example/BrowserTabUserControl.cs
```diff
@@ -259,7 +259,6 @@ private void SetupMessageInterceptor()
                             {
                                 const int WM_MOUSEACTIVATE = 0x0021;
                                 const int WM_NCLBUTTONDOWN = 0x00A1;
-                                const int WM_LBUTTONDOWN = 0x0201;
```
amaitland: Revert this

CefSharp.Wpf/ChromiumWebBrowser.cs
```diff
@@ -703,10 +707,10 @@ private void InternalDispose(bool disposing)
             Cef.RemoveDisposable(this);
         }
 
-        /// <summary>
-        /// Gets the ScreenInfo - currently used to get the DPI scale factor.
-        /// </summary>
-        /// <returns>ScreenInfo containing the current DPI scale factor</returns>
+/// <summary>
```
amaitland: Revert this

CefSharp.Wpf/ChromiumWebBrowser.cs
```diff
@@ -960,7 +964,47 @@ void IRenderWebBrowser.OnCursorChange(IntPtr handle, CursorType type, CursorInfo
...
+                return current as Window;
+            }
+
+            var imeKeyboardHandler = WpfKeyboardHandler as WpfKeyboardHandler;
+            if (imeKeyboardHandler.IsActive)
+            {
+                var screenInfo = GetScreenInfo();
+                var scaleFactor = screenInfo.HasValue ? screenInfo.Value.DeviceScaleFactor : 1.0f;
+
+                UiThreadRunSync(() =>
```
amaitland: Is there a reason this is run sync? Add comment if there is.

CefSharp.Wpf/ChromiumWebBrowser.cs
```diff
@@ -960,7 +964,47 @@ void IRenderWebBrowser.OnCursorChange(IntPtr handle, CursorType type, CursorInfo
...
+                {
+                    var parentWindow = GetParentWindow();
+                    Point pnt = parentWindow.PointToScreen(new Point(0, 0));
```
amaitland: What is the purpose of this  code?  Please add comment.

amaitland:
@a-marmer View your proposed changes using the Files Changed tab above, aka https://github.com/cefsharp/CefSharp/pull/2749/files

amaitland:
Has this been tested on HighDPI displays? 

Initially will need to split this out into it's own keyboard handler so it's optional.  If that's sorted in the next day or two it can be inclu7in the 73 release for people to test. 

Personally IME is not something I'm familiar with,  nor do I have a requirement for using it.

CefSharp.Wpf/ChromiumWebBrowser.cs
amaitland: We are manually tracking screen location already using https://github.com/cefsharp/CefSharp/blob/master/CefSharp.Wpf/ChromiumWebBrowser.cs#L1741
Should be able to reuse that and do away with the sync call.

a-marmer: GetBrowserScreenLocation() returns absolute position, here is used relative position.

a-marmer:
It was tested with HighDPI and it works.. Though if DPI is changed when application is running, `PresentationSourceChangedHandler` is not always called and then IME window position is not correct. 
I tested `SystemEvents.DisplaySettingsChanged += OnDisplaySettingsChanged;`  and `OnDisplaySettingsChanged` is always called. Maybe to use `OnDisplaySettingsChanged`?

CefSharp3.sln builds on my machine, it is not clear why build fails with appveyor, do you know why?

amaitland:
Inline functions are not supported by the default compiler used with `VS2015`

CefSharp.Wpf/ChromiumWebBrowser.cs
```diff
@@ -559,7 +562,7 @@ private void NoInliningConstructor()
             browserSettings = new BrowserSettings(frameworkCreated: true);
             RenderHandler = new InteropBitmapRenderHandler();
 
-            WpfKeyboardHandler = new WpfKeyboardHandler(this);
```
amaitland: As I said we won't be enabling this by default so revert this

CefSharp.Wpf/ChromiumWebBrowser.cs
amaitland: Please add a detailed description of what is being calculated and the reason for it, comment you added above is pretty meaningless. 

Relative position to what? At a minimum this will need to be run asynchronous, running sync is too problematic,

CefSharp.Wpf/ChromiumWebBrowser.cs
```diff
@@ -958,9 +961,48 @@ void IRenderWebBrowser.OnCursorChange(IntPtr handle, CursorType type, CursorInfo
...
+            {
+                var screenInfo = GetScreenInfo();
+                var scaleFactor = screenInfo.HasValue ? screenInfo.Value.DeviceScaleFactor : 1.0f;
+
+                // Needs to be executed in 'source' window thread (imeKeyboardHandler subclassed 'source' in 'setup' function).
+                UiThreadRunSync(() =>
+                {
+                    var parentWindow = GetParentWindow();
+                    if (parentWindow != null)
+                    {
```
amaitland: This is the code that requires a clear description

CefSharp.Wpf/ChromiumWebBrowser.cs
```diff
@@ -958,9 +961,48 @@ void IRenderWebBrowser.OnCursorChange(IntPtr handle, CursorType type, CursorInfo
             }
         }
 
+        private Visual GetParentWindow()
```
amaitland: Seems like you are reinventing the wheel here, what was the problem using the build in function? 

https://docs.microsoft.com/en-us/dotnet/api/system.windows.window.getwindow?view=netframework-4.8

CefSharp.Wpf/ChromiumWebBrowser.cs
```diff
@@ -958,9 +961,48 @@ void IRenderWebBrowser.OnCursorChange(IntPtr handle, CursorType type, CursorInfo
...
+            {
+                current = VisualTreeHelper.GetParent(current);
+            }
+
+            return current as Window;
+        }
+
         void IRenderWebBrowser.OnImeCompositionRangeChanged(Range selectedRange, Rect[] characterBounds)
         {
-            OnImeCompositionRangeChanged(selectedRange, characterBounds);
```
amaitland: You've effectively broken OnImeCompositionRangeChanged for anyone overriding it.

I think this whole block can likely be moved into the handler for now.

CefSharp.Wpf/Internals/WpfIMEKeyboardHandler.cs
```diff
@@ -0,0 +1,299 @@
...
+
+        internal bool IsActive { get; set; }
+
+        public WpfIMEKeyboardHandler(ChromiumWebBrowser owner): base(owner) {}
+
+        private void OwnerLostFocus(object sender, System.Windows.RoutedEventArgs e)
+        {
+            IsActive = false;
+
+            // These calls are needed in order for IME to function correctly.
```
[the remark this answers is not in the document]
a-marmer: Cannot explain, but only these calls combination works correctly. Also there were another IME implementation attempt   https://github.com/cefsharp/CefSharp/pull/2103/commits/8b9ae4fc9600fc8db431b5c06b9366876e4d7dec in which these calls are used :-)

amaitland:
I've taken your `PR` and squashed it into https://github.com/cefsharp/CefSharp/commit/840af401e73d15f1c9ca2c5ded6c755bcfe4e1cf

I've restructured things in https://github.com/cefsharp/CefSharp/commit/6ed722ad6c867f0448e712940a4c594ce1abb3d5 you can review and comment here. I haven't had time to test anything. Code is now self contained, it's in the `Experimental` namespace as I'm not really comfortable with the code as a whole, you are unable to explain key sections which obviously isn't your fault, as I don't know what they do it's hard to give the stamp of approval. Everything is fairly consistent at least now.

amaitland:
This was effectively merged in #2757

There has been some follow up in the form of #2782

If further changes are required then please submit a followup `PR`, thanks 👍

PR #2767, opened 2019-05-10
Refactor to avoid loading System.ServiceModel assembly

slozier:
When WCF is not used the System.ServiceModel assembly is not required. With this refactoring the runtime will avoid loading the assembly if WCF if not used.

Edit: My main motivation behind this change it to be able to use CefSharp with .NET Core 3.0. While in the long term it may be better to generate .NET Core specific assemblies, this change allows .NET Core which does not have System.ServiceModel to load the .NET Framework assemblies. As a side effect it also results in a slightly smaller memory footprint for .NET Framework applications that don't use WCF.

I have tested this change by using the NuGet packages (CefSharp.Common/CefSharp.WinForms generated using `build.ps1 vs2017`) and checking the loaded assemblies when running a program in Release mode. I've also tested it with a .NET Core 3.0 application and it loads up properly (instead of just crashing).

### DIFF
--- a/CefSharp.Core/ManagedCefBrowserAdapter.cpp
+++ b/CefSharp.Core/ManagedCefBrowserAdapter.cpp
@@ -49,7 +49,8 @@ void ManagedCefBrowserAdapter::CreateBrowser(IWindowInfo^ windowInfo, BrowserSet
     delete windowInfo;
 }
 
-void ManagedCefBrowserAdapter::InitializeBrowserProcessServiceHost(IBrowser^ browser)
+// NOTE: This was moved out of OnAfterBrowserCreated to prevent the System.ServiceModel assembly from being loaded when WCF is not enabled.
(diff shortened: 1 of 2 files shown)


amaitland:
Please update the description to include details on the reason for this change and how you've tested it. Overall looks like a simple solution :+1: 

Code wise please add comments as to the reason for the changes, make sure the methods are removed later as part of any code cleanup effort, they're only called once so they'd be a candidate to be removed. Did you test with a release build? The c# compiler would typically inline such a change as part of an optimisation effort, quick search and https://stackoverflow.com/questions/3329214/is-it-possible-to-force-a-function-not-to-be-inlined suggests it might be possible to specify noinline on a function in vc++, I haven't looked into this in any detail.

slozier:
@amaitland Thanks for the feedback. Updated as requested.

amaitland:
Great, thanks 👍

PR #2771, opened 2019-05-12
Async JavascriptBinding - Add support for returning Task

### LINKED ISSUE #2758: Feature Request - Async Javascript Binding Add support for returning Task
73.1.130

System architecture - x64, problem exists on both program architectures

Win10

`WPF`

Bind a simple object, which has an `async` method. Run the method from Javascript.
https://gist.github.com/perf2711/e96e69abad24c254d0192e88eb705709
I hosted the `index.html` file through `python3 -m http.server 8000`.

Synchronous methods work as they should, as well waiting for result from an asynchronous method in synchronous context. When running an `async` method, nothing is returned, promise never resolves and the object "locks up", preventing any additional calls.

This code was run inside the developer console.
```
>await TestBoundObject.syncMethod()
"SyncMethod"
>await TestBoundObject.syncMethod1()
"AsyncMethod1"
>await TestBoundObject.syncMethod2()
"AsyncMethod2"
>await TestBoundObject.syncMethod3()
"AsyncMethod3"
>await TestBoundObject.asyncMethod1()
>await TestBoundObject.asyncMethod2()
>await TestBoundObject.asyncMethod3()
>await TestBoundObject.syncMethod()
```

No exceptions are thrown. Breakpoints in each function are triggered.
I don't see any information regarding not using `async` methods in bound objects in `General usage`, however, there are no examples of such case.

No, at least not in the `Javascript Binding` test. Anyway I think this issue cannot be reproduced in the sample app.

amaitland:
Remove Concurrent Task execution code from MethodRunnerQueue
Add ConcurrentMethodRunnerQueue
  - For the new code path the Func is executed, if the Result is a generic Task we wait for it's completion
  - If the task is not generic then we return immediately

This is only enabled when CefSharpSettings.ConcurrentTaskExecution = true;

Additional tests will be required before this is production ready

Resolves #2758

### DIFF
--- a/CefSharp.Core/ManagedCefBrowserAdapter.cpp
+++ b/CefSharp.Core/ManagedCefBrowserAdapter.cpp
@@ -117,7 +117,7 @@ JavascriptObjectRepository^ ManagedCefBrowserAdapter::JavascriptObjectRepository
     return _javaScriptObjectRepository;
 }
 
-MethodRunnerQueue^ ManagedCefBrowserAdapter::MethodRunnerQueue::get()
+IMethodRunnerQueue^ ManagedCefBrowserAdapter::MethodRunnerQueue::get()
 {
     return _methodRunnerQueue;
 }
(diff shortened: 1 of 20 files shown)


amaitland:
Basics are complete, need to add a few more tests, something a little more real world.

At some point this should be the default behaviour. Short term we should return a error message if a Task is returned and direct the user to enable this feature.

amaitland:
TODO:

- [ ] Review Exception handling
- [ ] Add more test cases, they'll need to be run conditionally only if the `Parallel`  version is enabled.
- [ ] Return error message if using `Task` with default methodrunnerqueue

PR #2804, opened 2019-06-09
WinForms - Add example of how to dock devtools into the same window.

Geevo:
The pull request shows an example of how to dock devtools into a panel on the same window as the browser.

The only caveat is we require the LifeSpanHandler to be uncommented.

Using a SplitContainer we can mimic similar behaviour to that of Google Chrome.

A new devtools window can be opened for each tab.

Resizing/moving is actually all handled by the PopupAsChildHelper.

To close the devtools the parent control needs to be disposed of which will call the OnBeforeClose as normal and remove the item from the list of childhelpers.
(Unless the is another way to release the handle gracefully which i couldn't find)


amaitland:
Thanks! 👍 Will review after I've finished upgrading to the latest `CEF` version (fairly major rewrite)

amaitland:
I'll add a slight variation of this PR in the not too distant future, rather than replacing the existing devtools menu item we'll add a new one that specifically opens Devtools docked.

Geevo:
No problem. I’m more than happy to make any required changes if you like, let me know.

amaitland:
> To close the devtools the parent control needs to be disposed of which will call the OnBeforeClose as normal and remove the item from the list of childhelpers.
> (Unless the is another way to release the handle gracefully which i couldn't find)

@Geevo Can you clarify which `handle` you are referring? You make reference a few times in the comments though I'm not sure which handle you are referring.

> I’m more than happy to make any required changes if you like, let me know.

If you have time then by all means, I was just thinking we'd add a second menu option that was like `DevTools (Show Docked)` or something similar, and assign the `LifeSpanHandler` when that item was chosen. I haven't checked to see exactly how feasible this is.

Geevo:
> @Geevo Can you clarify which `handle` you are referring? You make reference a few times in the comments though I'm not sure which handle you are referring.

Apologises, I should of been more explicit. I'm referring to the DevTools window `handle` created by CEF. 

IIRC I couldn't find a way of removing/disposing the child  (DevTools) from its parent (Panel2) when hiding the `Panel` from the `SplitContainer `

> If you have time then by all means, I was just thinking we'd add a second menu option that was like `DevTools (Show Docked)` or something similar, and assign the `LifeSpanHandler` when that item was chosen. I haven't checked to see exactly how feasible this is.

Sure, i'll have a look at this, this week and see what i can accomplish. 
Perhaps with the `LifeSpanHandler` could set some `boolean` properties of what type of `popups` should be handled providing we can satisfy all possible potential conditions.

amaitland:
> Sure, i'll have a look at this, this week and see what i can accomplish.

Great, thanks 👍 

> Perhaps with the `LifeSpanHandler` could set some `boolean` properties of what type of `popups` should be handled providing we can satisfy all possible potential conditions

Historically there hasn't been a way to identify `DevTools` in `OnBeforePopup`, I haven't checked in the last version. By all means check for yourself, I may have missed something or the new `NetworkService` implementation behaves different (which would be ideal if it provided the URL).

CefSharp.WinForms.Example/Handlers/LifeSpanHandler.cs
```diff
@@ -69,17 +69,30 @@ void ILifeSpanHandler.OnAfterCreated(IWebBrowser browserControl, IBrowser browse
...
             if (browser.MainFrame.Url.Equals("chrome-devtools://devtools/devtools_app.html"))
             {
+                //If the IBrowserHost.CloseDevTools() is called whilst the brower is docked
+                //we need to handle it appropriately.
+                //The only way to release the handle is by disposing of its parent.
+                if (parentHandle != null)
+                {
+                    chromiumWebBrowser.Invoke(new Action(() =>
+                    {
+                        parentHandle.Dispose();
```
amaitland: Are you actually calling `Dispose` on the `Panel` a second time here?

amaitland:
> IIRC I couldn't find a way of removing/disposing the child (DevTools) from its parent (Panel2) when hiding the `SplitContainer.Panel2`.

I was somewhat expecting `CloseDevTools` to be called and the cleanup to be done in `DoClose`, is the problem removing the `Panel` from it's parent?

CefSharp.WinForms.Example/BrowserTabUserControl.cs
```diff
@@ -434,5 +434,42 @@ private void FindCloseButtonClick(object sender, EventArgs e)
...
+            }
+        }
+
+        public void CloseDevToolsDocked()
+        {
+            if (!browserSplitContainer.Panel2Collapsed)
+            {
+               browserSplitContainer.Panel2Collapsed = true;
+
+                if (devToolsPanel != null || !devToolsPanel.IsDisposed)
```
campersau:
```suggestion
                if (devToolsPanel != null && !devToolsPanel.IsDisposed)
```

CefSharp.WinForms.Example/BrowserTabUserControl.cs
```diff
@@ -434,5 +434,42 @@ private void FindCloseButtonClick(object sender, EventArgs e)
...
+
+        public void ShowDevToolsDocked()
+        {
+            if (browserSplitContainer.Panel2Collapsed)
+            {
+                browserSplitContainer.Panel2Collapsed = false;
+
+                if (devToolsPanel == null || devToolsPanel.IsDisposed)
+                {
+                    devToolsPanel = new Panel()
```
amaitland: I don't think the `Panel` should be initially created in the `Designer`, if we're creating a new instance of it then that doesn't really fit. Also makes the example harder to copy and paste.

CefSharp.WinForms.Example/BrowserTabUserControl.cs
```diff
@@ -434,5 +434,42 @@ private void FindCloseButtonClick(object sender, EventArgs e)
...
+                    {
+                        Dock = DockStyle.Fill
+                    };
+                    browserSplitContainer.Panel2.Controls.Add(devToolsPanel);
```
amaitland: Running this multiple times are we leaving `Disposed` panels in the `Controls` collection?

CefSharp.WinForms.Example/Handlers/LifeSpanHandler.cs
Geevo: I added this as a precaution in the event that if the `IBrowserHost.CloseDevTools()` was called directly without the use of `CloseDevToolsDocked()`. 
if `CloseDevToolsDocked()` gets called accordingly the `parentHandle` should already be null and the `Dispose` shouldn't get called for a second time. 
However in this instance the `SplitContainer.Panel2` will remain visible but all children  should be disposed.

I'm happy to change this so that all the clean up is done via the `ILifeSpanHandler.DoClose`

CefSharp.WinForms.Example/BrowserTabUserControl.cs
Geevo: Sure, are you happy for me to create a field variable outside of the designer instead?

CefSharp.WinForms.Example/BrowserTabUserControl.cs
Geevo: Good point! This is highly likely every disposed `Panel` will exist the collection.
Will check this out!

CefSharp.WinForms.Example/Handlers/LifeSpanHandler.cs
Geevo: Maybe in hindsight I should of also checked if the parent is `IsDisposed` or `Disposing`

Geevo: Ah I remember now, when calling to `CloseDevToolsDocked()` which will `Dispose` the `Panel`.
The `ILifeSpanHandler.DoClose()` does not get called at all, it goes straight to `ILifeSpanHandler.OnBeforeClose()`

CefSharp.WinForms.Example/BrowserTabUserControl.cs
amaitland: I'd probably avoid it, just retrieve the instance from the Controls collection when required.

Geevo:
Apologies for the delay, I've had less time to look at this than I anticipated.

Sadly, there is still no real way to fully identify between a `browser window popup` or `devtools` popup.
However, a `browser window popup` will call always call the `OnBeforeBrowse()` first, the `devtools` will only call the `OnAfterCreated()`.

We can set the `LifeSpanHandler` constructor parameter `handleBrowserPopups` to `false` , this will ignore handling popups and will use CEF's default behavior and as result they will not have a `parent` to which we can safely assume this is the `devtools` popup and add it to the `popupasChildHelpers` collection.

Here is currently a list of changes I've completed currently.
- Removed devToolsPanel from designer
- Removed CloseDevToolsDocked() method
- Made ShowDevToolsDocked() method easier to copy and paste
- Closing Devtools should now be handled in LifeSpanHandler.DoClose()
- Added a bool to the LifeSpanHandler constructor whether to handle Browser popups.
- Added Toolscript menu item to show DevTools Docked or Popup
- Added async Method to check it DevTools is open

amaitland:
Thanks :+1: Hopefully I'll have time to take this for a spin in the next few days.

Overall it'll need a more comments describing the custom behaviour,  I'll add some as I go.

Ultimately it'll be squashed into a single commit,  so splitting them out isn't necessary.

amaitland:
I made some minor changes and added/revised some of the comments see https://github.com/cefsharp/CefSharp/commit/7980c304784cb9806e1a40f4be0719d5a078debd

Any further additions/corrections by all means submit another `PR` 👍

PR #2812, opened 2019-06-17
OffScreen - ChromiumWebBrowser constructor change RequestContext to IRequestContext

medoni:
Changed typeof requestContext to IRequestContext in ChromiumWebBrowser.ctor

My requestContext was returned from RequestContext.CreateContext()


amaitland:
Thanks 👍

amaitland:
> My requestContext was returned from RequestContext.CreateContext()

Short term you should be able to simply cast `IRequestContext` to `RequestContext`

medoni:
> Short term you should be able to simply cast `IRequestContext` to `RequestContext`
yes this works. thx

PR #2745, opened 2019-04-25
WPF - Implement Touch and Stylus Input

GSonofNun:
Resolves #228 

Feedback welcome. In general seems to work the same way chrome does with touch and stylus input.

PR #2817, opened 2019-06-26
WPF - Fix and improve touch and stylus

kkwpsv:
**Summary:** 
   - With #2745, touch and stylus can work.
   - There're still a bug. Some operations may send duplicate events (like mousemove and touchmove when touch) , which cause bugs in some pages.
   - Stylus pen can be improved. When use stylus pen, there will be many points in single event, and we can use all of them.
   - With the help of my friend @walterlv, we fixed this bug and made this improvement.

**Changes:**
   - Check if only mouse in all OnMouseXXX methods.
   - Check if only stylus pen in all OnStylusXXX methods.
   - Send all points to host in OnStylus.
   - Fix comments. 
      
## How Has This Been Tested?
We have run the wpf example at Surface with mouse, touch and stylus pen. It works fine.

## Types of changes
- [ ] Updated documentation

**Checklist:**
- [x] Commented my code

### DIFF
--- a/CefSharp.Wpf/Experimental/ChromiumWebBrowserWithTouchSupport.cs
+++ b/CefSharp.Wpf/Experimental/ChromiumWebBrowserWithTouchSupport.cs
@@ -20,9 +20,7 @@ namespace CefSharp.Wpf.Experimental
     /// </summary>
     public class ChromiumWebBrowserWithTouchSupport : ChromiumWebBrowser
     {
-        #region Mouse
-        //Mouse, touch and sylus pen will Run OnMouseXXX.
-        //Use e.StylusDevice == null to ensure only mouse.
+        //Mouse, touch and stylus pen will Run OnMouseXXX.
 
         /// <summary>
         /// Handles the <see cref="E:MouseButton" /> event.
@@ -45,6 +43,7 @@ namespace CefSharp.Wpf.Experimental
         /// <param name="e">The <see cref="T:System.Windows.Input.MouseEventArgs" /> that contains the event data.</param>
         protected override void OnMouseMove(MouseEventArgs e)
         {
+            //Use e.StylusDevice == null to ensure only mouse.
             if (e.StylusDevice == null)
             {
                 base.OnMouseMove(e);
@@ -57,6 +56,7 @@ namespace CefSharp.Wpf.Experimental
         /// <param name="e">The <see cref="T:System.Windows.Input.MouseButtonEventArgs" /> that contains the event data. The event data reports that the mouse button was released.</param>
         protected override void OnMouseUp(MouseButtonEventArgs e)
         {
+            //Use e.StylusDevice == null to ensure only mouse.
             if (e.StylusDevice == null)
             {
                 base.OnMouseUp(e);
@@ -71,6 +71,7 @@ namespace CefSharp.Wpf.Experimental
         /// This event data reports details about the mouse button that was pressed and the handled state.</param>
         protected override void OnMouseDown(MouseButtonEventArgs e)
         {
+            //Use e.StylusDevice == null to ensure only mouse.
             if (e.StylusDevice == null)
             {
                 base.OnMouseDown(e);
@@ -83,15 +84,13 @@ namespace CefSharp.Wpf.Experimental
         /// <param name="e">The <see cref="T:System.Windows.Input.MouseEventArgs" /> that contains the event data.</param>
         protected override void OnMouseLeave(MouseEventArgs e)
         {
+            //Use e.StylusDevice == null to ensure only mouse.
             if (e.StylusDevice == null)
             {
                 base.OnMouseLeave(e);
             }
         }
 
-        #endregion
-
-        #region Touch
         //Only Touch will Run OnTouchXXX.
 
         /// <summary>
@@ -182,12 +181,8 @@ namespace CefSharp.Wpf.Experimental
             }
         }
 
-        #endregion Touch
-
-        #region Stylus Pen
         //Both touch and stylus pen will Run OnStylusXXX.
-        //We use OnTouchXXX to handle touch, which is faster.
-        //Use e.StylusDevice.TabletDevice.Type == TabletDeviceType.Stylus to ensure only stylus pen.
+        //We use the OnTouchXXX methods which contain more intermediate points to handle the user's touch so that we can track the user's fingers faster.
 
         /// <summary>
         /// Invoked when an unhandled <see cref="E:System.Windows.Input.StylusDown" /> attached event reaches an element in its route that is derived from this class. Implement this method to add class handling for this event.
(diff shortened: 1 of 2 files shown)


CefSharp.Wpf/Experimental/ChromiumWebBrowserWithTouchSupport.cs
```diff
@@ -20,20 +20,84 @@ namespace CefSharp.Wpf.Experimental
     /// </summary>
     public class ChromiumWebBrowserWithTouchSupport : ChromiumWebBrowser
     {
+        #region Mouse
+        //Mouse, touch and sylus pen will Run OnMouseXXX.
```
walterlv: Fix spelling error from `sylus` to `stylus`.

CefSharp.Wpf/Experimental/ChromiumWebBrowserWithTouchSupport.cs
```diff
@@ -20,20 +20,84 @@ namespace CefSharp.Wpf.Experimental
     /// </summary>
     public class ChromiumWebBrowserWithTouchSupport : ChromiumWebBrowser
     {
+        #region Mouse
```
amaitland: Please remove the regions

CefSharp.Wpf/Experimental/ChromiumWebBrowserWithTouchSupport.cs
```diff
@@ -118,38 +182,54 @@ private void OnTouch(TouchEventArgs e)
             }
         }
 
+        #endregion Touch
+
+        #region Stylus Pen
+        //Both touch and stylus pen will Run OnStylusXXX.
+        //We use OnTouchXXX to handle touch, which is faster.
```
walterlv: We use the OnTouchXXX methods which contain more intermediate points to handle the user's touch so that we can track the user's fingers faster.

CefSharp.Wpf/Experimental/ChromiumWebBrowserWithTouchSupport.cs
```diff
@@ -20,20 +20,84 @@ namespace CefSharp.Wpf.Experimental
...
             }
         }
+
+        /// <summary>
+        /// Invoked when an unhandled <see cref="E:System.Windows.Input.Mouse.MouseMove" /> attached event reaches an element in its route that is derived from this class. Implement this method to add class handling for this event.
+        /// </summary>
+        /// <param name="e">The <see cref="T:System.Windows.Input.MouseEventArgs" /> that contains the event data.</param>
+        protected override void OnMouseMove(MouseEventArgs e)
+        {
+            if (e.StylusDevice == null)
```
amaitland: Please add comments

amaitland: I think this could be more descriptive,  explain in what case mouse and stylus events are present in the one event please.

CefSharp.Wpf/Experimental/ChromiumWebBrowserWithTouchSupport.cs
```diff
@@ -118,38 +181,53 @@ private void OnTouch(TouchEventArgs e)
             }
         }
 
+        //Both touch and stylus pen will Run OnStylusXXX.
```
amaitland: Please remove the random comments and add a remarks section to the class xml doc, provide a detailed summary of the different cases please. The details are spread out and that's not ideal

amaitland:
Thanks for the the contribution,  will look in more detail hopefully in the next few days.

Please list the devices you've tested on, including operating system version and revision e.g. Surface Pro 1 running Windows 10 1809, thanks

CefSharp.Wpf/Experimental/ChromiumWebBrowserWithTouchSupport.cs
amaitland: Using comment from OnMouseButton above I'd fine

kkwpsv:
The logic of onMouseXXX in `ChromiumWebBrowser` is a bit mess.
I handle all mouse event in `ChromiumWebBrowserWithTouchSupport` and only pass mouse event from 
real mouse to `ChromiumWebBrowser`.

CefSharp.Wpf/Experimental/ChromiumWebBrowserWithTouchSupport.cs
kkwpsv: Mouse and stylus events won't present in the one event please.
But one stylus operation will raise both mouse event and stylus event.

walterlv:
@amaitland I've tested this pull request in my Surface Pro device and these are the device information:

* Surface Pro 5
* Windows 10 (10.0.17763.n) (1809)
* Input Devices: 2 Stylus Pen device + 10 points Precise Touch + Touch Pad

kkwpsv:
I also tested it on my pc with a touchscreen.

- Compatible PC
- Windows 10 (10.0.18362.175) (1903)
- Input Devices: PHL 231C5T Screen + Mouse

amaitland:
As a general comment a `Git Commit Message` that simply says `Fix` really isn't particularly meaningful. The git commit should have a description of the changes. Please provide detailed commit messages in future.

> The logic of onMouseXXX in `ChromiumWebBrowser` is a bit mess.

How exactly is it a mess? The whole idea is that `Touch` support is added without making any changes to `ChromiumWebBrowser`.

> I handle all mouse event in `ChromiumWebBrowserWithTouchSupport` and only pass mouse event from
> real mouse to `ChromiumWebBrowser` now.

That is perfectly acceptable as `ChromiumWebBrowserWithTouchSupport` is **experimental**, improvements will be made to `ChromiumWebBrowser` when code `Touch` support is eventually added and `ChromiumWebBrowserWithTouchSupport` is removed.

> I also send all points from stylus event instead of single, which can make handwriting smoother.

Great improvement, thanks 👍 This sort of comment should be included in your `Git Commit` message. If I was to review your commit without reading this `PR` then I would have no idea what code improvements were made.

> In addition, it seem to be wrong behavior in `OnMouseLeave` in `ChromiumWebBrowser`.
> When we pressed mouse button and leave , it shouldn't send `MouseClickEvent`. At least in chrome, it won't click.

That's issue https://github.com/cefsharp/CefSharp/issues/2258 I'm well are of the behaviour being different.

kkwpsv:
> How exactly is it a mess? The whole idea is that Touch support is added without making any changes to ChromiumWebBrowser.

`ChromiumWebBrowserWithTouchSupport` overrides `OnMouseButton` before, but `OnMouseButton` is called by other mouse events.  It's not a directy event handler. I think `OnMouseButton` should be private, and not be overrided by `ChromiumWebBrowserWithTouchSupport`. Now, I only override event handlers, and not make any change to `ChromiumWebBrowser`. Changes made to `ChromiumWebBrowser` in #2745 has been restored.

amaitland:
> `ChromiumWebBrowserWithTouchSupport` overrides `OnMouseButton` before, but `OnMouseButton` is called by other mouse events. It's not a direct event handler. I think `OnMouseButton` should be private, and not be overrided by `ChromiumWebBrowserWithTouchSupport`

Much better explanation, instead of `is a bit mess` next time provide detail like this, thanks 👍 

> Changes made to `ChromiumWebBrowser` in #2745 has been restored.

Just for clarity, #2745 was never merged, what you are actually referring to is the changes I made in commit https://github.com/cefsharp/CefSharp/commit/3d14f198674dba6516276897d384c1cc0c996f92

Issue #2821, opened 2019-06-27
PdfPrintSettings margin values are in points, not millimeters


matthew-a-thomas:
- **What version of the product are you using?**
Build of https://github.com/cefsharp/CefSharp/commit/f77c6d2cb404694a2598e905bad6af110e4fdb5f (latest as of this writing)
- **Architecture, OS, etc**
All
- **What steps will reproduce the problem?**
The comments on the `MarginX` (top, bottom, left, right) properties in `CefSharp.PdfPrintSettings` indicate that they are measured in millimeters.
- **What is the expected output?**
Those properties are actually measured in points (1"/72).
- **Please provide any additional information below.**
As recognized in https://github.com/cefsharp/CefSharp/issues/1808 this is a CEF issue first. So I created this issue:
https://bitbucket.org/chromiumembedded/cef/issues/2701/cefpdfprintsettings-margins-are-in-points
But perhaps CefSharp doesn't need to wait for CEF to correct their documentation. Or if CefSharp does need to wait on CEF (perhaps out of principle) then perhaps this GitHub issue here can be used to track status in CefSharp. But I won't tell you what to do :)

Issue #1808, opened 2016-09-23
Incorrect margin values


arvigeus:
```cs
var settings = new PdfPrintSettings()
{
    BackgroundsEnabled = true,
    MarginType = CefPdfPrintMarginType.Custom,
    MarginTop = 20,
    MarginRight = 20,
    MarginBottom = 20,
    MarginLeft = 20
};
```

Result: [PDF opened with GIMP](http://s12.postimg.org/4ayr3x3lp/Capture.png) - 6.9mm

amaitland:
### How do I investigate my issue further?

It's important to remember that `CefSharp` is just a wrapper around the `Chromium Embedded Project (CEF)`, the upstream [`CEF` forum](http://magpcss.org/ceforum/) is a valuable resource. Also `CEF` has it's own [Issue Tracker](https://bitbucket.org/chromiumembedded/cef/issues). Please take the time to research your own problem, including `Googling`, searching on `ceforum`, etc.

--
#### Finding more info
- Have you checked the `Debug.log` file? (Should be located in your apps executing folder by default)
- Have you googled the error message your seeing? (Don't just google the whole thing, try googling just the keywords and possibly add `chromium` as a keyword).
- Are you able to reproduce the problem using `cefclient`? (`cefclient` is the reference implementation provided by `CEF`)
- Need more help debugging? Check the [Troubleshooting Guide](https://github.com/cefsharp/CefSharp/wiki/Trouble-Shooting)
- Are you using the latest version? Check the [Release Branches](https://github.com/cefsharp/CefSharp#release-branches) table to determine if the version your using is supported. 

--
#### How do I download `cefclient`?

- Firstly identity which version of `CEF` your actually using, easiest way is to check your `packages.config` file, look for the `cef.redist.x86` package, it's version will be something like `3.2526.1362`, that's the exact version of `CEF` you'll need to download for comparison. You can also open `chrome://version` in the browser to determine the version.
- Open http://opensource.spotify.com/cefbuilds/index.html (http://cefbuilds.com for older version) in a web browser
- Locate the exact `Branch` you require, for the example above the branch is `2526`, all the versions follow a similar pattern
- Locate the `Windows 32bit` or `Windows 64bit` item, if you don't see the exact version, then click `more revisions`
- Download **Test App**
- Are you using `WPF/Offscreen` or `WinForms`? 

**This next part is key**
[...]

PR #2822, opened 2019-06-27
PdfPrintSettings margin measured in points, not mm

matthew-a-thomas:
Fixes https://github.com/cefsharp/CefSharp/issues/2821 and https://github.com/cefsharp/CefSharp/issues/1808

**Summary:**
The comments on the `MarginX` (top, bottom, left, right) properties in `CefSharp.PdfPrintSettings` now properly reflect that they are measured in points, not mm.

**Changes:**
Changed comments on the `MarginX` properties in `CefSharp.PdfPrintSettings`
      
## How Has This Been Tested?
Created PDF from simple HTML using top margin of `0.5 * factor` and left margin of `1 * factor`
  * When `factor` is 25.4 (millimeters per inch), the resulting PDF has margins that are visibly too small
  * When `factor` is 72 (points per inch), the resulting PDF looks okay. Additionally, Adobe Acrobat Reader includes a measuring tool which snaps to corners and edges of objects. I used it to precisely measure the left and top margins:
measured margins
(there is a boundary within the `H1` DOM element's content box even though I styled the `H1` to have no margin; in spite of this, Adobe Acrobat Reader is able to detect the content box's boundaries and its measuring tool is able to snap to the boundaries of the content box, hence the gap between the measurements and the text. If there remains any doubt due to this gap I can produce another example using a colored `div` or something that will have no whitespace in its content box... this is just what I happened to come up with first)

This demonstrates that the margins are measured in points, not millimeters.

## Types of changes
- [x] Updated documentation

**Checklist:**
- [x] Commented my code

### DIFF
--- a/CefSharp/PdfPrintSettings.cs
+++ b/CefSharp/PdfPrintSettings.cs
@@ -34,22 +34,22 @@ namespace CefSharp
         public int PageHeight { get; set; }
 
         /// <summary>
-        /// Margin in millimeters. Only used if MarginType is set to Custom.
+        /// Margin in points (1"/72). Only used if MarginType is set to Custom.
         /// </summary>
         public double MarginLeft { get; set; }
 
         /// <summary>
-        /// Margin in millimeters. Only used if MarginType is set to Custom.
+        /// Margin in points (1"/72). Only used if MarginType is set to Custom.
         /// </summary>
         public double MarginTop { get; set; }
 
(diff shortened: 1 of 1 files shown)


amaitland:
Thanks,  quick look at the source and it appears to be in points, only comment would be is the margins appear to be scaled based on the scale factor which might be nice for users to know.

PR #2829, opened 2019-07-08
Javascript Integration - Passing empty array in JS now maps to empty list/array instead of null

angshuman-agarwal:
**Summary:** [When Empty Array is passed from JS, do not set Null]
   - Always set an empty list instead of Null

   -  I have modified the SerializeV8Object Implementation
      - In case of arrays, moved the list creation out of the if statement & then always setting the list to the set index.
      

## Types of changes

- [ ] Updated documentation


amaitland:
Please add some `QUnit` tests.

https://github.com/cefsharp/CefSharp/blob/master/CefSharp.Example/JavascriptBinding/AsyncBoundObject.cs
https://github.com/cefsharp/CefSharp/blob/master/CefSharp.Example/Resources/BindingTest.html

A more meaningful `PR` title is required, `Set empty list instead of null` doesn't mean much to anyone reading the change log or release notes.

angshuman-agarwal:
> Please add some `QUnit` tests.

Done

> A more meaningful PR title is required

Done

Issue #2060, opened 2017-06-05
ScrollBar scrolling when mouse move without left button pressing


AppxStudio:
- **What version of the product are you using?**
    -57.0.0

- **What architecture x86 or x64?**
    -x64
- **On what operating system?**
    -Win10

- **Are you using `WinForms`, `WPF` or `OffScreen`?**
  -WPF
- **What steps will reproduce the problem?**
    1)Run the official project 'CefSharp.Wpf.Example'
    2)Draging the scrollbar with left button pressing
    3)move the cursor out of the browser control(move to right side)with left button pressing
    4)release left button
    5)move in browser control,the scrollbar will scrolling with mouse moving

martinburrows:
I'm also seeing this with the latest version 62 prerelease.

The scrolling behavior is also inconsistent with the behavior of other standard windows. When the mouse button is held down on the scrollbar and the mouse is moved out of the window slightly to the right, scrolling should continue (try it with your browser on this page) - CefSharp allows slight movement to the left when scrolling but not to the right. This makes it very difficult to use unless the window is maximized.

martinburrows:
Hi @amaitland - the fix in 49463b2 still does not allow the mouse to be moved even slightly to the right of the scrolllbar, this affects usablilty. Standard windows do allow the mouse to be moved either to the left or the right of the scrollbar.

I had better results by capturing the mouse instead - this works fine with the example app in Windows 10:

```
    protected override void OnMouseLeave(MouseEventArgs e)
    {
            if (!e.Handled && browser != null)
            {
                var modifiers = e.GetModifiers();
            
                if (modifiers == CefEventFlags.LeftMouseButton)
                {
                    Mouse.Capture(this, CaptureMode.SubTree);
                }
                else
                {
                    browser.GetHost().SendMouseMoveEvent(-1, -1, true, modifiers);
                }
                
                ((IWebBrowserInternal)this).SetTooltipText(null);
            }
    }
```

amaitland:
@martinburrows Thanks for the feedback, as you can see that change has not been merged as limited testing has been done.

Comments on your code:

- Should be sending the correct coordinates for `SendMouseMoveEvent`, sending `-1` gives weird results.
- For the `Mouse.Capture` you don't appear to be calling `Mouse.Capture(null)` to release the mouse, is there extra code? Or you haven't quite got that far yet?

[...]

PR #2844, opened 2019-07-24
WPF - Use Canvas.Top="0" for PART_image absolute position.

chenhuang444:
Fixes #2843

**Summary:**
   - CefSharp Image Jumps Vertically on Height Change

**Changes:**
   - Change canvas offset to Canvas.Top for PART_image in Generic.xaml
      
## How Has This Been Tested?
   - I do not have a build environment for CefSharp set up, so I tested this by overriding the default template with the changes in this commit. So further testing may be needed

## Screenshots (if appropriate):
CefSharpBug

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Updated documentation

**Checklist:**
- [ ] Tested the code(if applicable)
- [ ] Commented my code
- [ ] Changed the documentation(if applicable)
- [ ] New files have a license disclaimer
- [ ] The formatting is consistent with the project (project supports .editorconfig)

Issue #2820, opened 2019-06-27
Html dropdown does not position correctly when it is close to the bottom of the screen


Dumpen:
- **What version of the product are you using?**
Myget, CefSharp.Wpf, 75.0.110-CI3174

- **What architecture x86 or x64?**
x64

- **On what operating system?**
Win10

- **Are you using `WinForms`, `WPF` or `OffScreen`?**
WPF

- **What steps will reproduce the problem?**
1. Clone https://github.com/Dumpen/CefSharp.MinimalExample
2. Set CefSharp.MinimalExample.Wpf and WebApplication1 as start up projects
3. Start it

- **What is the expected output? What do you see instead?**
The dropdown should be above the select input, but it is below

- **Please provide any additional information below.**
See image [here](https://i.ibb.co/pQZQ5th/ss.jpg). In a normal browser the dropdown is either above the select input or above the whole window if there is no space.

- **Does this problem also occur in the `CEF` Sample Application from http://opensource.spotify.com/cefbuilds/index.html?**
No, see https://i.ibb.co/nwC1nsN/ss2.jpg. I tested with cef_binary_75.0.11+gf50b3c2+chromium-75.0.3770.100_windows64_client.tar

amaitland:
Additional checks to make sure the popup is within the bounds of the control will be required.

Currently the image that represents the drop down is just positioned relative to the canvas
https://github.com/cefsharp/CefSharp/blob/cefsharp/73/CefSharp.Wpf/ChromiumWebBrowser.cs#L1962

A very quick look at the `CEF` source code, and it's doing some checks about the size
https://github.com/chromiumembedded/cef/blob/8a055593d694b386f46530f81f53ed8b7c5ea284/tests/cefclient/browser/osr_renderer.cc#L247

We should do similar, pull requests welcome.

Dumpen:
I changed the code to match the one in CEF which moves the popup but the underlying popup is still where it was before. Is there a method to "paint/refresh" the UI?

   ```
     /// <summary>
        /// Sets the popup size and position implementation.
        /// </summary>
        /// <param name="rect">The popup rectangle (size and position).</param>
        private void SetPopupSizeAndPositionImpl(Rect rect)
        {
            popupImage.Width = rect.Width;
            popupImage.Height = rect.Height;

            int x = rect.X, y = rect.Y;

            // if x or y are negative, move them to 0.
            if (x < 0)
            {
                x = 0;
            }

            if (y < 0)
            {
[...]

PR #2871, opened 2019-08-22
WPF - Capture/Release mouse to allow scrolling outside bounds of browser control

### LINKED ISSUE #2870: WPF - Selecting text continuing below the visible area does not scroll
76.1.90.0, but it's also tested to occur in 74.1.130.0

x64

Windows 10

WPF

Hold the left mouse button to start selecting text on the default "Getting started" page of the WPF example and move the cursor to the bottom of the screen, expecting scrolling to follow automatically, but it only scrolls a tiny bit and then stops.

The browser should scroll automatically to keep being able to select text without having to scroll manually.

Setting `settings.CefCommandLineArgs.Add("disable-threaded-scrolling", "1");` seems to initally fix the issue, but after selecting text a few more times, triggering the auto scroll, the auto scroll suddenly stops working - and any subsequent attempts will not scroll.
The debug.log file does not appear to have any relevant information.

No, the sample application works properly.

mol:
Fixes #2870 and #2060 

**Summary:**
I re-added and improved the previously suggested fix to the no scrolling outside browser issue, which it turns out also fixes the selection issue when dragging past the viewable area.

I know you're not super keen on Mouse.Capture, which I totally understand, but I think it's the only way to accomplish it after having read up on how it works and its implementation in the reference source. `PreviewMouseDownOutsideCapturedElement` I don't believe would accomplish anything related to this.

**Changes:**
I re-added the `Mouse.Capture(this)` and `Mouse.Capture(null)` calls in `OnMouseDown` and `OnMouseUp` respectively, but in a way so the mouse is only captured on left mouse down (I can't think of a reason to capture the other buttons).

I also removed the code in `OnMouseLeave` that was sending a mouse up event when the left mouse is pressed, as it didn't have any effect when testing with these other changes (so I assume is no longer required). But yeah it doesn't seem to do anything at all in relation with the changes though so it could also be kept.
      
## How Has This Been Tested?
Tested in Windows 10 x64. I believe I've looked through all of the issues, commits and pull requests in relation to these issues and I've tested that combo boxes still work and doesn't cause any crashes, as well as every imaginable scenario of button presses to make sure nothing has been broken as a result. I'm happy to run this out to our users and report back in a while if you'd prefer to err on the side of caution though.

## Types of changes

- [ ] Updated documentation

**Checklist:**
- [x] Commented my code

### DIFF
--- a/CefSharp.Wpf/ChromiumWebBrowser.cs
+++ b/CefSharp.Wpf/ChromiumWebBrowser.cs
@@ -2190,12 +2190,11 @@ namespace CefSharp.Wpf
             Focus();
             OnMouseButton(e);
 
-            // We should only need to capture the left button exiting the browser
+            //We should only need to capture the left button exiting the browser
             if (e.ChangedButton == MouseButton.Left && e.LeftButton == MouseButtonState.Pressed)
             {
-                // Capture mouse so events are routed to the browser when it's outside of the control.
-                // This fixes selection issue https://github.com/cefsharp/CefSharp/issues/2870 and the issue with 
-                // no scrolling when on the right of a scroll bar: https://github.com/cefsharp/CefSharp/issues/2060
+                //Capture/Release Mouse to allow for scrolling outside bounds of browser control (#2870, #2060).
+                //This doesn't capture the mouse outside of the WPF Window, see #2258  for further discussion on that topic.
                 Mouse.Capture(this);
             }
 
@@ -2212,10 +2211,10 @@ namespace CefSharp.Wpf
 
             if (e.ChangedButton == MouseButton.Left && e.LeftButton == MouseButtonState.Released)
             {
-                // Release the mouse capture that we grabbed on mouse down.
-                // We won't get here if e.g. the right mouse button is pressed and released
-                // while the left is still held, but in that case the left mouse capture seems
-                // to be released implicitly (even without the left mouse SendMouseClickEvent in leave below)
+                //Release the mouse capture that we grabbed on mouse down.
+                //We won't get here if e.g. the right mouse button is pressed and released
+                //while the left is still held, but in that case the left mouse capture seems
+                //to be released implicitly (even without the left mouse SendMouseClickEvent in leave below)
                 Mouse.Capture(null);
             }
 
--- a/CefSharp.Wpf/ChromiumWebBrowser.cs
+++ b/CefSharp.Wpf/ChromiumWebBrowser.cs
@@ -2190,6 +2190,15 @@ namespace CefSharp.Wpf
             Focus();
             OnMouseButton(e);
 
+            // We should only need to capture the left button exiting the browser
+            if (e.ChangedButton == MouseButton.Left && e.LeftButton == MouseButtonState.Pressed)
+            {
+                // Capture mouse so events are routed to the browser when it's outside of the control.
+                // This fixes selection issue https://github.com/cefsharp/CefSharp/issues/2870 and the issue with 
+                // no scrolling when on the right of a scroll bar: https://github.com/cefsharp/CefSharp/issues/2060
+                Mouse.Capture(this);
+            }
+
             base.OnMouseDown(e);
         }
 
@@ -2200,6 +2209,15 @@ namespace CefSharp.Wpf
         protected override void OnMouseUp(MouseButtonEventArgs e)
         {
             OnMouseButton(e);
+
+            if (e.ChangedButton == MouseButton.Left && e.LeftButton == MouseButtonState.Released)
+            {
+                // Release the mouse capture that we grabbed on mouse down.
+                // We won't get here if e.g. the right mouse button is pressed and released
+                // while the left is still held, but in that case the left mouse capture seems
+                // to be released implicitly (even without the left mouse SendMouseClickEvent in leave below)
+                Mouse.Capture(null);
+            }
 
             base.OnMouseUp(e);
         }
@@ -2215,16 +2233,7 @@ namespace CefSharp.Wpf
                 var modifiers = e.GetModifiers();
                 var point = e.GetPosition(this);
 
-                //If the LeftMouse button is pressed when leaving the control we send a mouse click with mouseUp: true
-                //to let the browser know the mouse has been released
-                if (e.LeftButton == MouseButtonState.Pressed)
-                {
-                    browser.GetHost().SendMouseClickEvent((int)point.X, (int)point.Y, MouseButtonType.Left, mouseUp: true, clickCount: 1, modifiers: modifiers);
-                }
-                else
-                {
-                    browser.GetHost().SendMouseMoveEvent((int)point.X, (int)point.Y, true, modifiers);
-                }
+                browser.GetHost().SendMouseMoveEvent((int)point.X, (int)point.Y, true, modifiers);
 
                 ((IWebBrowserInternal)this).SetTooltipText(null);
             }


amaitland:
> I know you're not super keen on Mouse.Capture, which I totally understand, but I think it's the only way to accomplish it after having read up on how it works and its implementation in the reference source.

I'll admit a capture is likely the only way to get the usability that everyone expects. Adding a link to #1723 for future reference, if this does crash then the stack trace should be compared.

> I re-added the `Mouse.Capture(this)` and `Mouse.Capture(null)` calls in `OnMouseDown` and `OnMouseUp` respectively, but in a way so the mouse is only captured on left mouse down (I can't think of a reason to capture the other buttons).

I was trying to avoid performing a capture on every down/up combination. As part of #2258 my initial attempt was https://github.com/cefsharp/CefSharp/commit/c843317e338e572fb485b31b6ec3c9c7a517ca53 That was a long time ago, the only major problem I rememeber it had was leaving would always perform a capture even if a parent control had a capture at the time, an additional not null check on [Mouse.Captured](https://docs.microsoft.com/en-us/dotnet/api/system.windows.input.mouse.captured?view=netframework-4.8) should probably be enough to resolve that. The change was released as part version `65`, so it was in use for a short period of time and I don't remember any reports of crashing.

> But yeah it doesn't seem to do anything at all in relation with the changes though so it could also be kept.

With a capture in place the code should be removed.

> I've tested that combo boxes still work and doesn't cause any crashes,

The mouse capture was removed and replaced with directly forwarding the events from the `WPF Popup`, subsequently the `Popup` was removed and replaced with drawing the image on a canvas directly over the rendered image. So there shouldn't be any conflicts. The downside of rendering the `Popup` on a `Canvas` is it's constrained within the `Control`. We may need to revert to using a `WPF Popup` as part of #2820

> I'm happy to run this out to our users and report back in a while if you'd prefer to err on the side of caution though.

Sounds good to me, let me know how it goes 👍

CefSharp.Wpf/ChromiumWebBrowser.cs
```diff
@@ -2190,6 +2190,15 @@ protected override void OnMouseDown(MouseButtonEventArgs e)
             Focus();
             OnMouseButton(e);
 
+            // We should only need to capture the left button exiting the browser
+            if (e.ChangedButton == MouseButton.Left && e.LeftButton == MouseButtonState.Pressed)
+            {
+                // Capture mouse so events are routed to the browser when it's outside of the control.
+                // This fixes selection issue https://github.com/cefsharp/CefSharp/issues/2870 and the issue with 
+                // no scrolling when on the right of a scroll bar: https://github.com/cefsharp/CefSharp/issues/2060
```
amaitland: > the issue with no scrolling when on the right of a scroll bar: https://github.com/cefsharp/CefSharp/issues/2060

This should probably reference be replaced with a reference to issue #2258 and make clear the limitation of the capture only working within the `WPF Window`.

Strictly speaking issue #2060 was resolved. Leaving the browser with the mouse down, releasing the mouse outside the browser and returning would still register as having the mouse down and scroll  the browser on mouse move. The fix for #2060 was limited and admittidly not particularl user friendly, hence #2258 was created.

amaitland: Revise that, I'm probably being too pedantic here. Having an extra link doesn't matter, the description not matching the actual issue was the part that was bugging me 😄 

Let's just go with something like below, focus on the intent of the code.

```c#
//Capture/Release Mouse to allow for scrollowing outside bounds of browser control (#2870, #2060).
//This doesn't capture the mouse outside of the WPF Window, see #2258  for further discussion on that topic.
```

amaitland:
Took me a while to remember the reason I actually abandoned the initial set of changes in #2258 (https://github.com/cefsharp/CefSharp/commit/c843317e338e572fb485b31b6ec3c9c7a517ca53). So this will actually reintroduce #2060 when the mouse exits the `WPF Window` (same problem occurs with my attempted fix).

Slightly more detailed steps to reproduce than in #2060 would be:
- Open `WPF Example`
- Open new Tab (google.com should be displayed, this is important as we need to leave the window whilst scrolling)
- Press the `Restore Down` menu bar shortcut (top right) or alternatively press windows key + down
- Drag scroll bar down using mouse, leaving the window
- Release mouse button outside of window
- Move mouse up and down over scroll bar without pressing any buttons

CefSharp.Wpf/ChromiumWebBrowser.cs
mol: Sure, I've updated the comments :)

mol:
> Took me a while to remember the reason I actually abandoned the initial set of changes in #2258 ([c843317](https://github.com/cefsharp/CefSharp/commit/c843317e338e572fb485b31b6ec3c9c7a517ca53)). So this will actually reintroduce #2060 when the mouse exits the `WPF Window` (same problem occurs with my attempted fix).
> 
> Slightly more detailed steps to reproduce than in #2060 would be:
> 
> * Open `WPF Example`
> * Open new Tab (google.com should be displayed, this is important as we need to leave the window whilst scrolling)
> * Press the `Restore Down` menu bar shortcut (top right) or alternatively press windows key + down
> * Drag scroll bar down using mouse, leaving the window
> * Release mouse button outside of window
> * Move mouse up and down over scroll bar without pressing any buttons

You just added this for reference right? I followed the steps and it didn't reproduce the issue for me. Everything looks to be working fine. I also checked that drag/drop to and from works (in Mailbird).

amaitland:
> You just added this for reference right?

Unfortunately no. I just did a clean checkout of https://github.com/cefsharp/CefSharp/pull/2871/commits/356699576f64e5973df5c790e89753681f904c62 and can reliably reproduce the issue.

amaitland:
Screen capture that demos the problem, you should be able to visually see when I have the mouse pressed, as it has a circle. After I've released the mouse the browser still thinks the mouse down as it didn't get the correct message, moving the mouse anywhere on the scroll bar triggers a scroll when no mouse is actually pressed. I tried upgrading `CefSharp.Wpf.Example` to `.Net 4.7.2` and that made no difference.

cefsharpmousecapture

mol:
> Screen capture that demos the problem, you should be able to visually see when I have the mouse pressed, as it has a circle. After I've released the mouse the browser still thinks the mouse down as it didn't get the correct message, moving the mouse anywhere on the scroll bar triggers a scroll when no mouse is actually pressed. I tried upgrading `CefSharp.Wpf.Example` to `.Net 4.7.2` and that made no difference.
> 
> cefsharpmousecapture

That's really weird. I'm running the same branch/commit and with no custom changes and it's not happening for me...
CdxVUUr44P
When the mouse is captured all mouse events should route to the browser right, so the only thing I can think of is something else is capturing the mouse for you but not me. Any ideas? Any test code that might be running for you but not me?

mol:
What version of Windows are you running? I'm on Windows 10 (1903).

amaitland:
> When the mouse is captured all mouse events should route to the browser right, so the only thing I can think of is something else is capturing the mouse for you but not me.

I cannot really tell from the screen capture when you released the mouse, so it's really hard to say if you are doing anything different.

> Any test code that might be running for you but not me?

No, I did a clean checkout to confirm that it is still a problem. It's the exact reason that I reverted the previous mouse capture code last year. 

> What version of Windows are you running? I'm on Windows 10 (1903).

I recently upgraded to `1903` (`Microsoft` finally fixed the bug that was preventing me from upgrading).

I'll test again on a different machine when I can, probably in a week or so.

mol:
> > When the mouse is captured all mouse events should route to the browser right, so the only thing I can think of is something else is capturing the mouse for you but not me.
> 
> I cannot really tell from the screen capture when you released the mouse, so it's really hard to say if you are doing anything different.
> 
> > Any test code that might be running for you but not me?
> 
> No, I did a clean checkout to confirm that it is still a problem. It's the exact reason that I reverted the previous mouse capture code last year.
> 
> > What version of Windows are you running? I'm on Windows 10 (1903).
> 
> I recently upgraded to `1903` (`Microsoft` finally fixed the bug that was preventing me from upgrading).
> 
> I'll test again on a different machine when I can, probably in a week or so.

I'm releasing it outside the window so definitely the same as you.

Well we've released it to all of our users and so far no issues. I have a task for myself to update this pull request in a couple of weeks (earlier if I hear of any issues) so we'll see how it goes :)

mol:
I just had Dimas from the team test it and he can reproduce it. He's using a button on a track pad, not a mouse - so if you're doing the same then I think that's why it happens to him and you and not me.

mol:
Turns out this is related to this issue https://github.com/dotnet/wpf/issues/1323

Adding the workaround mentioned in https://github.com/dotnet/wpf/issues/1323#issuecomment-513571323 to the WPF example fixes the issue. It disables tablet support though and might not be something you want to do in CefSharp. We're already doing it in Mailbird only for the two affected versions.

amaitland:
> He's using a button on a track pad, not a mouse

I am using a Mouse, though this machine does have a touch screen.

> Turns out this is related to this issue [dotnet/wpf#1323](https://github.com/dotnet/wpf/issues/1323)

The stated versions of `1809` and `1903` don't match with my comments in #2258 I don't remember which machine I would have tested on, perhaps it was this one, could have been a desktop. Either way, well before `1809` was released. If time permits I'll do some more testing. 

> Adding the workaround mentioned in [dotnet/wpf#1323 (comment)](https://github.com/dotnet/wpf/issues/1323#issuecomment-513571323) to the WPF example fixes the issue.

Just as an `FYI` you should be able to set `Switch.System.Windows.Input.Stylus.DisableStylusAndTouchSupport` switch, via https://docs.microsoft.com/en-us/dotnet/api/system.appcontext.setswitch?view=netframework-4.8 to disable touch and stylus support. (also configurable via registry for what it's worth).

> It disables tablet support though and might not be something you want to do in CefSharp

I don't think it's the place of a third party framework like `CefSharp` to disable Stylus and Touch support.

timaiv:
I fixed this issue with capturing half year ago for my project like you (but more custom code existed).
More, you need add browser internal call, when capture lost. (mouse button up)
All works fine.

mol:
> I fixed this issue with capturing half year ago for my project like you (but more custom code existed).
> More, you need add browser internal call, when capture lost (LostMouseCapture Event). (mouse button up)
> All works fine.

Thanks. What exactly did you add for the LostMouseCapture event? Do you have some code to share? How does it differ from https://github.com/cefsharp/CefSharp/commit/356699576f64e5973df5c790e89753681f904c62?

Also, have you tested this on a laptop with a touch screen? The issue with the mouse button not being released is only an issue on laptops with a touch screen as far as I know, and as far as I know is due to the referenced bug in WPF.

timaiv:
So. Firstly, cefsharp had bug with content jumping while resizing. And i fixed it by replacing canvas to grid. I extend grid with my custom Grid class.

`internal class eCEFGrid : Grid
{
	private bool isMouseCaptured;

	public event Action<MouseEvent> OnCaptureLostAfterMouseUp;

	public eCEFGrid()
	{
		Mouse.AddLostMouseCaptureHandler(this, LostMouseCapture);
	}

	private void LostMouseCapture(object sender, MouseEventArgs e)
	{
		isMouseCaptured = false;
	}

	protected override void OnMouseDown(MouseButtonEventArgs e)
	{
		CaptureMouse();
		isMouseCaptured = true;
		base.OnMouseDown(e);
	}

	protected override void OnMouseUp(MouseButtonEventArgs e)
	{
		base.OnMouseUp(e);
		Point position = e.GetPosition(this);
		if (isMouseCaptured)
		{
			this.OnCaptureLostAfterMouseUp?.Invoke(new MouseEvent((int)position.X, (int)position.Y, CefEventFlags.None));
		}
		ReleaseMouseCapture();
		isMouseCaptured = false;
	}
}
`
Now, in subclass to ChromiumWebBrowser i do following:
`internal class eCEFGrid : Grid
{
	private bool isMouseCaptured;

	public event Action<MouseEvent> OnCaptureLostAfterMouseUp;

	public eCEFGrid()
	{
		Mouse.AddLostMouseCaptureHandler(this, LostMouseCapture);
	}

	private void LostMouseCapture(object sender, MouseEventArgs e)
	{
		isMouseCaptured = false;
	}

	protected override void OnMouseDown(MouseButtonEventArgs e)
	{
		CaptureMouse();
		isMouseCaptured = true;
		base.OnMouseDown(e);
	}

	protected override void OnMouseUp(MouseButtonEventArgs e)
	{
		base.OnMouseUp(e);
		Point position = e.GetPosition(this);
		if (isMouseCaptured)
		{
			this.OnCaptureLostAfterMouseUp?.Invoke(new MouseEvent((int)position.X, (int)position.Y, CefEventFlags.None));
		}
		ReleaseMouseCapture();
		isMouseCaptured = false;
	}
}
`

Tested on tablet with touch. All works fine everywhere.

mol:
Thanks a lot @timaiv I'll take a look at that when I have some time. About the "jumping" issue - I think I've had users experience that. Can you tell me how to reproduce it?

amaitland:
> About the "jumping" issue - I think I've had users experience that. Can you tell me how to reproduce it?

There was an issue resolved in version 75 (#2844)

@timaiv If you can still reproduce this issue you were seeing then please open a new issue (it's unrelated to mouse capture).

> For ex. CaptureMouse() can return false.

@mol This is something we should handle, store a local variable and only release the mouse capture if it was successfully obtained.

CefSharp.Wpf/ChromiumWebBrowser.cs
```diff
@@ -2201,6 +2209,15 @@ protected override void OnMouseUp(MouseButtonEventArgs e)
         {
             OnMouseButton(e);
 
+            if (e.ChangedButton == MouseButton.Left && e.LeftButton == MouseButtonState.Released)
+            {
+                //Release the mouse capture that we grabbed on mouse down.
+                //We won't get here if e.g. the right mouse button is pressed and released
+                //while the left is still held, but in that case the left mouse capture seems
+                //to be released implicitly (even without the left mouse SendMouseClickEvent in leave below)
+                Mouse.Capture(null);
```
amaitland: If we switch to `UIElement.ReleaseMouseCapture();` it has an additional `Mouse.Captured == this` check. 

https://referencesource.microsoft.com/#PresentationCore/Core/CSharp/System/Windows/UIElement.cs,2514

May as well switch to `UIElementCaptureMouse()` as well for consistency.

timaiv:
Also, you need really subscribe LostMouseCapture event, because you **can lose capture when your button is pressed** to notify browser.
Alt+tab with pressed left button or any other window dialog.
[http://www.shujaat.net/2010/11/wpf-windows-forceful-release-of-mouse.html]()
______________
From my experience, wpf automatically forces to lose capture in some cases if you didnt subscribe any special capture event, i got it. But, now i cant reproduce it on win7, maybe it lie.
______________
I wrote that cefsharp **had** jumping issue. I wrote reason why i used grid.

timaiv:
Finally!
I tried to remove grid, and fix it normally, as i reproduced that mouse still has pressed button on win10 before grid using.
SendMouseClickEvent need be sent before ReleaseMouseCapture(), that why i had problems without my second SendMouseClickEvent.
I did this order implicit it in grid and why it worked. 
Experimental ChromiumWebBrowserWithMouseCapture works normal. But you need more logic, that i wrote (LostMouseCapture event).

amaitland:
> Experimental ChromiumWebBrowserWithMouseCapture works normal. But you need more logic, that i wrote (LostMouseCapture event).

@timaiv You are welcome to submit a `PR` with improvements, please make sure you include a detailed list of steps to reproduce the problem your changes fix. Thanks

CefSharp.Wpf/ChromiumWebBrowser.cs
mol: You already made these changes right? It's a little unclear from the commit with the ChromiumWebBrowserWithMouseCapture but it looks like it.

How are you envisioning this drag fix will make it into the next release now that it's experimental? Should I do something? :)

timaiv:
I added next:
```
protected override void OnLostMouseCapture(MouseEventArgs e)
{
    GetBrowser().GetHost().SendCaptureLostEvent();
    base.OnLostMouseCapture(e);
}
```

and saw that some bug still exists with **alt+tab** (my custom scrollbar stopped jumping after mouse returning on host, that is normal, but style was like pressed). I checked that chrome, firefox scrollbars have same behavior (but it also continue jumps on hovers), other apps like VS don't do it. SendMouseClickEvent also in this case doesn't help.

CefSharp.Wpf/ChromiumWebBrowser.cs
amaitland: > You already made these changes right? It's a little unclear from the commit with the ChromiumWebBrowserWithMouseCapture but it looks like it.

I did. Those changes are only in the `cefsharp/75` branch at the moment. `ChromiumWebBrowserWithMouseCapture` is already marked as `Obsolete`. 

> How are you envisioning this drag fix will make it into the next release now that it's experimental?

I'm reluctant to assign a `milestone` until I get a better understanding of which versions of `Windows 10` the fix will be applied to. Leave this for now, will revisit once there's some clarity. Thanks 👍

amaitland:
Manually merged with additional lost capture call in commit https://github.com/cefsharp/CefSharp/commit/21c89b130be418b3fa6f74c8351037e55d6affc9

Issue #2779, opened 2019-05-17
Browser goes blank if used in combination with WPF TabControl


dkrainyk:
- **What version of the product are you using?**
Nuget 73.1.130 and 74.1.130-CI3117 (not reproducible on 71.0.2)

- **What architecture x86 or x64?**
x64

- **On what operating system?**
Win10

- **Are you using `WinForms`, `WPF` or `OffScreen`?**
WPF

- **What steps will reproduce the problem?**
Use TabControl (or similar approach) to launch ~10 tabs of ChromiumWebBrowser pointing to google.com
Switch between tabs and eventually (after 10-20 switches) browser will go blank (according to debug tools browser still works and you can execute JS, change DOM, etc.).

Adapted minimal example: https://github.com/dkrainyk/CefSharp.MinimalExample/commit/8fa4bd572240695f485d4323a5ca247b1eacbb0b

- **What is the expected output? What do you see instead?**
Browser window goes blank. Page should be visible.
- **Please provide any additional information below.**
debug.log is empty

- **Does this problem also occur in the `CEF` Sample Application from http://opensource.spotify.com/cefbuilds/index.html?**
N/A

SolomoN-ua:
+1

amaitland:
Looks like a bug in CEF,  will add more details later. 

I suspect that testing with with cefclient and minimizing/maximizing the window will probably reproduce the same problem.

For now rollback to version  71.

hucongquan:
version 71  ,I had a similar problem;
environment: winform use cefsharp(nuget);
Sometimes when I refresh or load a page with a large amount of data, it appears white screen, is it too big  data for chrome to process?
![White Screen Issue Report Snapshot](https://user-images.githubusercontent.com/10986908/58083013-bf168a80-7bea-11e9-9205-70b39696361a.png)

amaitland:
> environment: winform use cefsharp(nuget);

@hucongquan If you are using `WinForm` then your issue is completely unrelated to this one.

> is it too big data for chrome to process?

Probably the render process is crashing with an out of memory exception if you are loading huge amounts of data. Upgrading to a newer version of `CEF` is likely the only fix (remember that `CefSharp` is just a wrapper).

---

[...]

PR #2879, opened 2019-08-30
WPF - Updated #2779 hack to run asynchronous without crashing.

mol:
Fixes crash issue in #2779 

**Summary:**
Fixed the crashes that could occur when running on the CEF UI thread async in the updated 2779 hack.

**Changes:**
CefUiThreadRunAsync is now awaited on and I also moved some more code to run on the CEF UI thread. All necessary to avoid the crashing. I made sure to test after each change so nothing is running on that thread that doesn't need to.
      
## How Has This Been Tested?
I ran the crash test code found in the WPF example MainWindow.xaml.cs on this branch: https://github.com/mol/CefSharp/tree/reworkedhack2779 again and again and again after each modification to see if the crashes would stop or not. I also went back and forth with my changes to ensure only the necessary changes were made to stop the crashing.

## Types of changes

- [ ] Updated documentation

**Checklist:**
- [x] Commented my code

### DIFF
--- a/CefSharp.Wpf/ChromiumWebBrowser.cs
+++ b/CefSharp.Wpf/ChromiumWebBrowser.cs
@@ -1769,12 +1769,24 @@ namespace CefSharp.Wpf
                     {
                         await CefUiThreadRunAsync(async () =>
                         {
-                            if (browser != null)
+                            var host = browser?.GetHost();
+                            if (host != null && !host.IsDisposed)
                             {
-                                browser.GetHost().WasHidden(false);
-                            }
+                                try
+                                {
+                                    host.WasHidden(false);
 
-                            await ResizeHackFor2779();
+                                    await ResizeHackFor2779();
+                                }
+                                catch (ObjectDisposedException)
+                                {
+                                    // Because Dispose runs in another thread there's a race condition between
+                                    // that and this code running on the CEF UI thread, so the host could be disposed
+                                    // between the check and using it. We can either synchronize access using locking
+                                    // (potentially blocking the UI thread in Dispose) or catch the extremely rare
+                                    // exception, which is what we do here
+                                }
+                            }
                         });
                     }
 
@@ -1784,14 +1796,22 @@ namespace CefSharp.Wpf
                 {
                     await CefUiThreadRunAsync(() =>
                     {
-                        if (EnableResizeHackForIssue2779)
+                        var host = browser?.GetHost();
+                        if (host != null && !host.IsDisposed)
                         {
-                            resizeHackForIssue2779Enabled = true;
-                        }
+                            if (EnableResizeHackForIssue2779)
+                            {
+                                resizeHackForIssue2779Enabled = true;
+                            }
 
-                        if (browser != null)
-                        {
-                            browser.GetHost().WasHidden(true);
+                            try
+                            {
+                                host.WasHidden(true);
+                            }
+                            catch (ObjectDisposedException)
+                            {
+                                // See comment in catch in OnWindowStateChanged
+                            }
                         }
                     });
 
@@ -1954,9 +1974,17 @@ namespace CefSharp.Wpf
                     SetViewRect(e);
                 }
 
-                if (browser != null)
+                var host = browser?.GetHost();
+                if (host != null && !host.IsDisposed)
                 {
-                    browser.GetHost().WasResized();
+                    try
+                    {
+                        host.WasResized();
+                    }
+                    catch (ObjectDisposedException)
+                    {
+                        // See comment in catch in OnWindowStateChanged
+                    }
                 }
             });
         }
@@ -1982,18 +2010,26 @@ namespace CefSharp.Wpf
             {
                 await CefUiThreadRunAsync(async () =>
                 {
-                    if (browser != null)
+                    var host = browser?.GetHost();
+                    if (host != null && !host.IsDisposed)
                     {
-                        browser.GetHost().WasHidden(!isVisible);
-                    }
+                        try
+                        {
+                            host.WasHidden(!isVisible);
 
-                    if (isVisible)
-                    {
-                        await ResizeHackFor2779();
-                    }
-                    else if (EnableResizeHackForIssue2779)
-                    {
-                        resizeHackForIssue2779Enabled = true;
+                            if (isVisible)
+                            {
+                                await ResizeHackFor2779();
+                            }
+                            else if (EnableResizeHackForIssue2779)
+                            {
+                                resizeHackForIssue2779Enabled = true;
+                            }
+                        }
+                        catch (ObjectDisposedException)
+                        {
+                            // See comment in catch in OnWindowStateChanged
+                        }
                     }
                 });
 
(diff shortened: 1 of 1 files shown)


mol:
@amaitland because ChromiumWebBrowser.Dispose is running on another thread than the CEF UI thread there's now a race condition between both the `browser` variable and more importantly the disposal of the `managedCefBrowserAdapter` - that can cause an `ObjectDisposedException`.

I have it reproducing in a stress test if you want to try it for yourself. Since Dispose can run on any thread and at any time, I think that even in a real world situation it could happen if the developer is disposing a bit "too fast".

I'm thinking there are three ways to go about fixing it.

1. Synchronize access to the `managedCefBrowserAdapter` so it's not disposed while being accessed (this would be tricky because when run on the WPF UI thread disposal waits for the CEF UI tasks to end, which in turn could be invoking on the WPF UI thread - this is the issue I saw when I tried it and it caused a deadlock, but just for two seconds since there's a timeout in CEF). Making dispose async and awaiting inside might work to avoid blocking the WPF UI thread, but that doesn't seem quite right.
2. By catching the exception where it can happen.
3. By disposing the adapter on the CEF UI thread.

I opted to catch the exception since it should be a rare occurance and it can safely be ignored presumably, since we're disposing. I don't know how you feel about that though as it is a bit ugly in my opinion. I also added checking of the host not being disposed instead of the thread unsafe `browser` variable. I also rearranged some code so nothing runs unnecessarily after the browser is disposed anyway.

Option #3 also seems to work well so that's also a way to go. Here's how that would look in the dispose method:
```
if (managedCefBrowserAdapter != null)
{
	// Dispose the browser adapter on the CEF UI thread to avoid race condition with code using it
	Cef.UIThreadTaskFactory.StartNew(() =>
	{
		managedCefBrowserAdapter.Dispose();
	}).ContinueWith(t =>
	{
		managedCefBrowserAdapter = null;
	},
	TaskScheduler.FromCurrentSynchronizationContext());
}
```

Let me know how you feel about all this and if you think running adapter dispose on the CEF UI thread is safe. If it is then it's probably the better option actually.

amaitland:
@mol Thanks for the update, if you can provide some test code for the dispose issue that would be helpful for evaluating the possible solutions.

mol:
> @mol Thanks for the update, if you can provide some test code for the dispose issue that would be helpful for evaluating the possible solutions.

This should do it:

```
private void MainWindowLoaded(object sender, RoutedEventArgs e)
{
	CreateNewTab(CefExample.DefaultUrl, true);

	// Reproduce the crash caused by WasResized and RenderWidgetHostImpl::SynchronizeVisualProperties.
	// See comments: https://github.com/cefsharp/CefSharp/issues/2779
	CreateNewTab();
	CreateNewTab();
	ReproduceWasResizedCrashAsync();
}

private void ReproduceWasResizedCrashAsync()
{
	WindowState = WindowState.Normal;

	Task.Run(() =>
	{
		try
		{
			var random = new Random();

			for (int i = 0; i < 20; i++)
			{
				for (int j = 0; j < 150; j++)
				{
					Dispatcher.Invoke(new Action(() =>
					{
						//if (random.Next(0, 5) == 0)
						//    WindowState = WindowState.Minimized;
						//else
						//    WindowState = WindowState.Normal;

						var newWidth = Width + (i % 2 == 0 ? -5 : 5);
						var newHeight = Height + (i % 2 == 0 ? -5 : 5);
						if (newWidth < 500 || newWidth > 1500)
						{
							newWidth = 1000;
						}
						if (newHeight < 500 || newHeight > 1500)
						{
							newHeight = 1000;
						}
						Width = newWidth;
						Height = newHeight;

						// Get all indexes but the selected one
						var indexes = new List<int>();
						for (int k = 0; k < TabControl.Items.Count; k++)
						{
							if (TabControl.SelectedIndex != k)
							{
								indexes.Add(k);
							}
						}

						// Select a random unselected tab
						TabControl.SelectedIndex = indexes[random.Next(0, indexes.Count)];

						// Close a tab and create a tab once in a while
						//if (random.Next(0, 5) == 0)
						{
							CloseTab(BrowserTabs[Math.Max(1, TabControl.SelectedIndex)]); // Don't close the first tab
							CreateNewTab();
						}
					}));

					// Sleep random amount of time
					Thread.Sleep(random.Next(1, 11));
				}
			}
		}
		catch (TaskCanceledException) { } // So it doesn't break on VS stop
	});
}
```

amaitland:
I rewrote the `hack` removing the threading complications of using the `CEF UI Thread` and using `SemaphoreSlim.WaitAsync` to sync the calls to `WasHidden/WasResized` unfortunately the browser flickered occasionally, so I'm just going with this for now, it appears to do the job and I'm really hoping we'll be able to abandon this hack in the not too distant future. 

I made a few minor changes in https://github.com/cefsharp/CefSharp/commit/5fa61c58ce60b9c60e59413413ba97acf5faec93 to cleanup the code a little, and I added a test menu item in https://github.com/cefsharp/CefSharp/commit/3c708dfc527c010c450e8f0929b7d41d27aededc to make life a little easier.

I'll hopefully have time to run the test a couple more times and then release `75.1.142` in a few days.

@mol Thanks for all the hard work 👍

amaitland:
> f I comment out the code that only closes a tab once in a while, so it does it every time, then I get `AccessViolationException` in the host wrapper often during the stress test, for instance when it's running `Invalidate`, but I've also seen it happen when running `WasHidden`. I've only seen it happen in the code inside `ResizeHackFor2779()` running after the await delay, and checking `host.IsDisposed` when the exception is caught shows it as `true`, so it was disposed **after** the check.

Are you seeing the issue in real world usage or just in testing?

> It might be the only way to solve it since you cannot catch that particular exception...

Taking a local reference and checking the pointer is valid before calling is also an option.

> but according to the MS analytics tool that we have enabled in Mailbird it was also happening in the two year old version of CefSharp so it's probably always been there

Far as I know they only added it in the back end of `2018`. As as you've probably seen it's been removed. Haven't checked which branch the change will be included in yet.

mol:
> Are you seeing the issue in real world usage or just in testing?

I don't have this branch out in the real world, I have the one where I'm running the hack with Wait() to have it sync with the UI thread, albeit with a small timeout since there was a case where we did something in Mailbird and it would deadlock like you were afraid of. I guess if we don't do anything we'll see if it'll happen in the real world when I do a full switch to the 75 release :)

> Taking a local reference and checking the pointer is valid before calling is also an option.

What would you take a local reference of? Not `host` because that isn't becoming null, it's that it disposes "inside" the host after the check but before the code inside runs.

> Far as I know they only added it in the back end of 2018. As as you've probably seen it's been removed. Haven't checked which branch the change will be included in yet.

Yeah I can't say that it's at the same exact place, only that it's a similar breakpoint crash of Chromium.

amaitland:
> What would you take a local reference of?

The `CefBrowserHost` pointer. The code would look something like below

```c++
void CefBrowserHostWrapper::WasResized()
{
    ThrowIfDisposed();

    CefRefPtr<CefBrowserHost> host = _browserHost.get();

    if (host.get())
    {
        host->WasResized();
    }
}
```

CefSharp.Wpf/ChromiumWebBrowser.cs
```diff
@@ -2442,33 +2501,30 @@ protected bool InternalIsBrowserInitialized()
...
-                CefUiThreadRunAsync(async () =>
+                var host = browser?.GetHost();
+                if (host != null && !host.IsDisposed)
                 {
-                    if (browser != null)
-                    {
-                        resizeHackForIssue2779Size = new Structs.Size(viewRect.Width - 1, viewRect.Height - 1);
-                        browser.GetHost().WasResized();
-                    }
+                    resizeHackForIssue2779Size = new Structs.Size(viewRect.Width - 1, viewRect.Height - 1);
```
joaompneves: I believe that this causes crashes on cef due when returning negative values. I was able to reproduce a crash because GetViewRect returns negetive width and height. Probably it would be safer to add 1px instead.

PR #2880, opened 2019-08-30
added note about '1' is not the opposite of '0'

### LINKED ISSUE #2858: Example code/documentation for CefCommandLineArgs usage is misleading
Throughout the CefSharp documentation and example code there are examples of enabling Chromium command-line switches like this:

`settings.CefCommandLineArgs.Add("disable-gpu", "1");`

While this does work, it also would logically lead you to think that if you wanted to programatically enable/disable a flag you could use a zero instead:

`settings.CefCommandLineArgs.Add("disable-gpu", "0");`

However, this is incorrect. As `disable-gpu` (and indeed, almost all of the CEF command-line arguments) are switchless, the "0" is ignored and the `--disable-gpu` flag takes effect. This can easily be reproduced by navigating to chrome://gpu and noting that hardware acceleration is disabled with both of the above lines of code, but if the argument is not added at all, hardware acceleration is enabled.

I would suggest that examples are re-written to use `string.Empty` as the second parameter to make it clear that in the case of switchless parameters it is the presence of the dictionary key alone that toggles the flag, not any particular value of the dictionary value.

`settings.CefCommandLineArgs.Add("disable-gpu", string.Empty);`

Some examples of documentation/code that has this issue:

https://github.com/cefsharp/CefSharp/wiki/General-Usage#cefsettings-and-browsersettings

https://github.com/cefsharp/CefSharp/blob/cdca99c548a641a5d3b7294e6fcbc5840d134508/CefSharp.Example/CefExample.cs
https://github.com/cefsharp/CefSharp/blob/b2acd4ad760e870556b67f51271ce4251f0847ac/CefSharp.OffScreen/CefSettings.cs
https://github.com/cefsharp/CefSharp/blob/c9203c93101e722c30a8c6151d4737eb943ba4d1/CefSharp.Wpf/CefSettings.cs
https://github.com/cefsharp/CefSharp/blob/23b905dbdaba194f3e4261272adb20e42ab0cbf9/CefSharp.Example/Resources/Home.html

- [ ] Update `Wiki` to use new `Add(string)` method.

miller45:
Fixes #2858
**Summary:** 
Added a note in the examples, that using "0" as value in a commandline switch does not switch that feature off

**Changes:** 
  Added some comments lines pointing out that setting are CefCommandLineArgs.Add("some-key", "0") is not switching off that given feature but the "opposite"
## How Has This Been Tested?
Testing in Windows 7 x64. I used the CefSharp.Wpf.Example project and tried the following settings:
settings.CefCommandLineArgs.Add("renderer-startup-dialog", "0");
settings.CefCommandLineArgs.Add("enable-media-stream", "0"); 
settings.CefCommandLineArgs.Add("no-proxy-server", "0"); 
settings.CefCommandLineArgs.Add("disable-gpu", "0"); 
And after starting the program I observed that using "0" or "" does not switch off any of the mentioned features. 

## Types of changes

- [x] Updated documentation

**Checklist:**
- [ ] Commented my code

### DIFF
--- a/CefSharp.Example/CefExample.cs
+++ b/CefSharp.Example/CefExample.cs
@@ -62,7 +62,13 @@ namespace CefSharp.Example
             //HTML5 databases such as localStorage will only persist across sessions if a cache path is specified. 
             settings.CachePath = "cache";
             //settings.UserAgent = "CefSharp Browser" + Cef.CefSharpVersion; // Example User Agent
+						
             //settings.CefCommandLineArgs.Add("renderer-process-limit", "1");
+			
+			// BEWARE: some of the examples like Add("renderer-startup-dialog", "1") suggest that using "0" instead of would switch off the feature			
+			// But this in not the case: For most of these on/off flags chrome just switches it on as long as it sees the flag like --renderer-startup-dialog 
+			// Using "0","1" or "hell-no" will make no difference			
+			
             //settings.CefCommandLineArgs.Add("renderer-startup-dialog", "1");
             //settings.CefCommandLineArgs.Add("enable-media-stream", "1"); //Enable WebRTC
             //settings.CefCommandLineArgs.Add("no-proxy-server", "1"); //Don't use a proxy server, always make direct connections. Overrides any other proxy server flags that are passed.


amaitland:
Thanks, taking a different approach, see https://github.com/cefsharp/CefSharp/issues/2858#issuecomment-528690143 for details.

PR #2881, opened 2019-08-30
Core - Add IURLRequest #1721

### LINKED ISSUE #1721: Feature Request - Add IUrlRequest
Add `Cef.CreateUrlRequest` and the appropriate interfaces/wrappers.

http://magpcss.org/ceforum/apidocs3/projects/(default)/CefURLRequest.html

davidmi:
Add IURLRequest, IURLRequestClient, and GetURLRequest to IFrame

Fixes #1721 

**Summary:** 
   I was interested in #1721 because I wanted to make a request using the offscreen browser and read the response body, so I added this functionality as proposed in #1721

**Changes:** 
   -- Added wrappers for CefURLRequest and CefURLRequestClient
   -- Added corresponding interfaces
   -- Added LoadURLRequest to CefFrameWrapper and IFrame
   -- Added and ran basic unit test for the new 
      
**Feedback requests**
   -- Is there somewhere better to put the test?
   -- The test suite "fails" even though all the tests pass for me because Cef.Shutdown() ends up being called on a different thread, I'm not sure if there is already an issue for this.
   -- Should I provide an implementation of IURLRequestClient? It is non-trivial to implement, I think, though I did put an example in the test.


CefSharp/IFrame.cs
```diff
@@ -182,5 +182,25 @@ public interface IFrame : IDisposable
...
+        /// of type PDE_TYPE_FILE or PDE_TYPE_BYTES.
+        /// For requests originating from the render process: - POST data may only contain a single element of type PDE_TYPE_BYTES.
+        /// - If the response contains Content-Disposition or Mime-Type header values that would not normally be rendered then
+        /// the response may receive special handling inside the browser
+        /// for example, via the file download code path instead of the URL request code path).
+        ///
+        /// The request object will be marked as read-only after calling this method. 
+        /// </summary>
+        //TODO: Implement IURLRequestClient.
+        IURLRequest CreateURLRequest(IRequest request, IURLRequestClient client);
```
davidmi: Since the request object becomes read-only here due to cef internals, maybe I should add a new ImmutableRequest class with a copy constructor from IRequest and no setters? This would protect the user from making a mistake and crashing cef, but I noticed cefsharp generally doesn't have this level of abstraction/internals protection.

amaitland: > maybe I should add a new ImmutableRequest class with a copy constructor from IRequest and no setters

I'd rather notify the user they are attempting to modify a readonly object rather than silently accepting the change.  I'll add a simple `ThrowIfReadOnly` guard shortly.

CefSharp/IFrame.cs
```diff
@@ -182,5 +182,25 @@ public interface IFrame : IDisposable
...
+        /// For requests originating from the browser process: - POST data may only contain a single element
+        /// of type PDE_TYPE_FILE or PDE_TYPE_BYTES.
+        /// For requests originating from the render process: - POST data may only contain a single element of type PDE_TYPE_BYTES.
+        /// - If the response contains Content-Disposition or Mime-Type header values that would not normally be rendered then
+        /// the response may receive special handling inside the browser
+        /// for example, via the file download code path instead of the URL request code path).
+        ///
+        /// The request object will be marked as read-only after calling this method. 
+        /// </summary>
+        //TODO: Implement IURLRequestClient.
```
amaitland: Remove this if no longer relevant.

CefSharp.Test/OffScreen/OffScreenBrowserBasicFacts.cs
```diff
@@ -127,5 +130,83 @@ public async Task CanEvaluateScriptAsyncWithEncodedStringArguments()
             }
         }
 
+        class URLRequestClient : IURLRequestClient
```
amaitland: One class per file please, can move this to `CefSharp.Example` project for use in the examples.

CefSharp.Test/Framework/RequestFacts.cs
```diff
@@ -0,0 +1,18 @@
+// Copyright © 2019 The CefSharp Authors. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style license that can be found in the LICENSE file.
+
+using Moq;
+using Xunit;
+
+namespace CefSharp.Test.Framework
+{
+    public class RequestFacts
```
amaitland: What's the plan for this class?

CefSharp.Test/CefSharpFixture.cs
```diff
@@ -26,7 +26,10 @@ public CefSharpFixture()
                 //HTML5 databases such as localStorage will only persist across sessions if a cache path is specified. 
                 settings.CachePath = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData), "CefSharp\\Tests\\Cache");
 
-                Cef.Initialize(settings, performDependencyCheck: false, browserProcessHandler: null);
+                var t = new System.Threading.Thread(() => {
```
amaitland: Revert this change please.

CefSharp.Core/Internals/CefURLRequestWrapper.h
```diff
@@ -0,0 +1,64 @@
+#pragma once
```
amaitland: Add license

CefSharp.Core/Internals/CefURLRequestWrapper.h
```diff
@@ -0,0 +1,64 @@
...
+
+#include "Stdafx.h"
+
+#include "include\cef_urlrequest.h"
+#include "include\cef_v8.h"
```
amaitland: Remove if not used.

CefSharp.Core/Internals/CefURLRequestClientAdapter.h
```diff
@@ -0,0 +1,91 @@
+#pragma once
```
amaitland: Add license.

CefSharp.Core/Internals/CefURLRequestClientAdapter.cpp
```diff
@@ -0,0 +1,55 @@
...
+}
+
+void CefURLRequestClientAdapter::OnDownloadProgress(CefRefPtr<CefURLRequest> request, int64 current, int64 total)
+{
+    _client->OnDownloadProgress(gcnew CefURLRequestWrapper(request), current, total);
+}
+
+void CefURLRequestClientAdapter::OnDownloadData(CefRefPtr<CefURLRequest> request, const void* data, size_t data_length)
+{
+    auto data_array = gcnew array<Byte>(data_length);
```
amaitland: Use `UnmanagedMemoryStream` to wrap the data in a `Stream.

CefSharp.Core/Internals/CefURLRequestClientAdapter.cpp
```diff
@@ -0,0 +1,55 @@
+// Copyright © 2019 The CefSharp Authors. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style license that can be found in the LICENSE file.
+
+#include "Stdafx.h"
+#include "CefURLRequestClientAdapter.h"
+#include "CefURLRequestWrapper.h"
+#include "CefAuthCallbackWrapper.h"
+
+#include "Cef.h"
```
amaitland: Is `Cef.h` used? Remove if not.

CefSharp.Core/Internals/CefURLRequestWrapper.cpp
```diff
@@ -0,0 +1,31 @@
+// Copyright © 2019 The CefSharp Authors. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style license that can be found in the LICENSE file.
+
+#include "Stdafx.h"
+#include "CefURLRequestWrapper.h"
+#include "CefResponseWrapper.h"
+
+#include "Cef.h"
```
amaitland: Remove if not used.

CefSharp/IFrame.cs
amaitland: Also rename to `CreateUrlRequest`

amaitland:
> The test suite "fails" even though all the tests pass for me because Cef.Shutdown() ends up being called on a different thread, I'm not sure if there is already an issue for this.

Known issue, at some point I'll get around to fixing this. Ignore for now.

> Should I provide an implementation of IURLRequestClient?

I think the example is ok for now.

amaitland:
> This is my first substantial pull request to any public project, please let me know if I've made any faux pas!

Excellent first attempt 👍  It's tough getting started as every project has it's own style and every maintainer their own quirks 😄

CefSharp/IFrame.cs
amaitland: `ThrowIfReadonly` guard added in https://github.com/cefsharp/CefSharp/commit/b35d5151fee5113bc2f833665394f1c9c2aa1ff8

Also a readonly copy of the `NameValueCollection` added. Let me know if you think I've missed anything.

amaitland:
Additionally remove https://github.com/davidmi/CefSharp/blob/1721-pullreq/CefSharp.Core/Internals/CefURLRequestClientWrapper.cpp if it's not used, thanks 👍 

I see you've made some changes, let me know when it's ready for review 😄

davidmi:
> 
> 
> Additionally remove https://github.com/davidmi/CefSharp/blob/1721-pullreq/CefSharp.Core/Internals/CefURLRequestClientWrapper.cpp if it's not used, thanks 👍

Sorry, I ran into some git silliness making a case-only filename change on NTFS, I'll clean stuff up on Linux and it should be ready then.

CefSharp.Core/Internals/CefURLRequestClientAdapter.cpp
davidmi: Done -- could you please make sure my memory management here is correct? I expect it should get deallocated once we leave the stack frame, but I'm not super familiar with how C++\CLI works.

davidmi:
> 
> 
> Additionally remove https://github.com/davidmi/CefSharp/blob/1721-pullreq/CefSharp.Core/Internals/CefURLRequestClientWrapper.cpp if it's not used, thanks 👍
> 
> I see you've made some changes, let me know when it's ready for review 😄

It's ready for review. Thanks!

amaitland:
> It's ready for review. Thanks!

Great, thanks 👍 Very quick glance and it looks like mostly just formatting and improvements to the xml doc that are required.

I'll review in detail when I have some free time.

amaitland:
See https://github.com/cefsharp/CefSharp/commit/4a79b757266c4e0646e71e430f486a853b210324 for follow up commit, mostly just formatting and xml doc improvements.

Thanks for all the hard work 👍

Issue #2796, opened 2019-05-31
Feature Request  - Add .Net Core 3.1 Support


amaitland:
With .Net Core 3.0 adding support for WPF and WinForms I'm seeing more and more queries about running CefSharp using .Net Core 3.0. This issue will track the official progress, there is no timeframe yet, please don't ask when this will be implemented,  .Net Core isn't even officially out yet. As per https://github.com/dotnet/core/blob/master/roadmap.md#upcoming-ship-dates the release is still some months away. 

Microsoft has started implementing C++/CLI as part of .Net Core you can track the issue https://github.com/dotnet/coreclr/issues/18013 They are only adding support for Windows initially. See also https://github.com/dotnet/coreclr/issues/659

`.Net Core` will support mixed mode assemblies on **Windows Only**.

https://devblogs.microsoft.com/dotnet/net-core-3-and-support-for-windows-desktop-applications/

The sync JavaScript Binding implementation will not work as Microsoft is not supporting WCF in .Net Core 3.0.

- [x] Refactor BrowserSubProcess #2891
- [x] Compile project targeting .Net Core (#3181)
- [x] Release Net Core/Net 5.0 packages

amaitland:
With https://github.com/cefsharp/CefSharp/pull/2767 it may be possible to load the next CefSharp release (75 at time of writing) with .Net Core 3.0 referencing the current .Net 4.5.2 framework assemblies. I have no personally tested this personally yet. You will have to avoid using the sync JavaScript Binding as it relies on WCF.

michalczerwinski:
Any news on this one?

kpreisser:
Hi,
first, many thanks for this great project!

I was able to successfully use [`CefSharp.WinForms 75.0.110-CI3174`](https://www.myget.org/feed/cefsharp/package/nuget/CefSharp.WinForms/75.0.110-CI3174) on a **.NET Core 3.0 (Preview 5)** WinForms application (form title in the screenshot contains process filename and `RuntimeInformation.FrameworkDescription` when running with `dotnet.exe`):
![cefsharp-windows](https://user-images.githubusercontent.com/13289184/60210656-a63e6c00-985d-11e9-9fbc-4a6ce0d1fc6e.png)

I also tested that the `IDownloadHandler` was working successfully.

However, directly adding the NuGet package to the .NET Core project did not work as it didn't resolve the dependencies (probably because they are only declared for .NET Framework 4.5.2).

[...]

PR #2885, opened 2019-09-04
WPF - Modify CopyMemory EntryPoint to work with .Net Core

kpreisser:
Contributes to #2796

**Summary:**
   - Fixed the `CopyMemory` entrypoint declaration, which is not working on .NET Core 3.0 as `kernel32.dll` doesn't have an `CopyMemory` export.

**Changes:**
- Changed `EntryPoint` from `CopyMemory` to `RtlCopyMemory`
- Changed the `numberOfBytes` parameter from `uint` to `UIntPtr`

Note: I believe using an pointer type (`IntPtr` or `UIntPtr`) for the `numberOfBytes` parameter is correct even though pinvoke.net (and other sites) specify it as `int`, because the documentation for [`CopyMemory`/`RtlCopyMemory`](https://docs.microsoft.com/en-us/previous-versions/windows/desktop/legacy/aa366535%28v%3dvs.85%29) declares it as `SIZE_T` which is defined as pointer-sized integer (`ULONG_PTR`). Additionally, I have done some testing to verify that the value is actually interpreted as 64-bit integer when running as x64 (by specifying a value greater than 0xFFFFFFFF and verifying the correct number of bytes was copied).

Note: There is also a managed API [`Buffer.MemoryCopy()`](https://docs.microsoft.com/en-us/dotnet/api/system.buffer.memorycopy?view=netcore-3.0#System_Buffer_MemoryCopy_System_Void__System_Void__System_Int64_System_Int64_) that could be used instead of p/invoking `RtlCopyMemory`/`RtlMoveMemory`, available starting with .NET Framework 4.6, but that one seems to be slower.
      
## How Has This Been Tested?

Tested manually using `CefSharp.MinimalExample.NetCore.Wpf` (see https://github.com/cefsharp/CefSharp.MinimalExample/pull/57)

## Types of changes

- [ ] Updated documentation

**Checklist:**
- [X] Commented my code

### DIFF
--- a/CefSharp.Wpf/Rendering/AbstractRenderHandler.cs
+++ b/CefSharp.Wpf/Rendering/AbstractRenderHandler.cs
@@ -20,8 +20,8 @@ namespace CefSharp.Wpf.Rendering
     /// <seealso cref="CefSharp.Wpf.IRenderHandler" />
     public abstract class AbstractRenderHandler : IDisposable, IRenderHandler
     {
-        [DllImport("kernel32.dll", EntryPoint = "CopyMemory", SetLastError = false)]
-        protected static extern void CopyMemory(IntPtr dest, IntPtr src, uint count);
+        [DllImport("kernel32.dll", EntryPoint = "RtlMoveMemory", SetLastError = false)]
+        protected static extern void CopyMemory(IntPtr dest, IntPtr src, UIntPtr count);
 
         internal static readonly PixelFormat PixelFormat = PixelFormats.Pbgra32;
         internal static int BytesPerPixel = PixelFormat.BitsPerPixel / 8;
--- a/CefSharp.Wpf/Rendering/AbstractRenderHandler.cs
+++ b/CefSharp.Wpf/Rendering/AbstractRenderHandler.cs
@@ -20,7 +20,8 @@ namespace CefSharp.Wpf.Rendering
     /// <seealso cref="CefSharp.Wpf.IRenderHandler" />
     public abstract class AbstractRenderHandler : IDisposable, IRenderHandler
     {
-        [DllImport("kernel32.dll", EntryPoint = "RtlMoveMemory", SetLastError = false)]
+        // Note: In contrast to RtlMoveMemory, RtlCopyMemory requires that the buffers do not overlap.
+        [DllImport("kernel32.dll", EntryPoint = "RtlCopyMemory", SetLastError = false)]
         protected static extern void CopyMemory(IntPtr dest, IntPtr src, UIntPtr count);
 
         internal static readonly PixelFormat PixelFormat = PixelFormats.Pbgra32;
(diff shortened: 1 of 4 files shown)


amaitland:
> * Changed `EntryPoint` from `CopyMemory` to `RtlMoveMemory`

Is there a reason you chose `RtlMoveMemory` over `RtlCopyMemory`? I see no reason to use the slower move memory.

> Note: There is also a managed API [`Buffer.MemoryCopy()`](https://docs.microsoft.com/en-us/dotnet/api/system.buffer.memorycopy?view=netcore-3.0#System_Buffer_MemoryCopy_System_Void__System_Void__System_Int64_System_Int64_) that could be used instead of p/invoking `RtlMoveMemory`, available starting with .NET Framework 4.6, but that one seems to be slower.

As we only target `.Net 4.5.2` this isn't an option anyway.

kpreisser:
> Is there a reason you chose `RtlMoveMemory` over `RtlCopyMemory`? I see no reason to use the slower move memory.

No particular reason, I was just following the comments in https://github.com/dotnet/coreclr/issues/24008 for changing `CopyMemory` to `RtlMoveMemory` (I assume because it doesn't require buffers to not overlap and so is safer).

I updated the PR to use the faster `RtlCopyMemory` instead.
Thanks!

amaitland:
> (I assume because it doesn't require buffers to not overlap and so is safer).

Actually it would appear to be that `RtlCopyMemory` only exists for `x64` as detailed in https://stackoverflow.com/a/47016390/4583726

Confirmed in the output of `dumpbin /exports kernel32.dll` there is indeed no `RtlCopyMemory`.

> #define RtlCopyMemory(Destination,Source,Length) memcpy((Destination),(Source),(Length))

`winnt.h` calls `memcpy`, so changing to directly call [memcpy](https://www.pinvoke.net/default.aspx/msvcrt/memcpy.html) appears to be one option.

The other being to use [NativeMethodWrapper.CopyMemoryUsingHandle](https://github.com/cefsharp/CefSharp/blob/cefsharp/75/CefSharp.Core/NativeMethodWrapper.cpp#L13) which I had previous added with the intention of bench marking to see if the native call was quicker than the `p/invoke` call. (Crappy method name I know, from memory it I couldn't call it `CopyMemory` or `RtlCopyMemory` as the macros interfere with the method name).

amaitland:
Changed to use native wrapper around `RtlCopyMemory` in https://github.com/cefsharp/CefSharp/commit/b1b930605c0b9d2a74da54506bfdef4263d11a65

Originally the `Rendering` implementation in `WPF` was written in `VC++` and used [MemoryCopy](https://github.com/cefsharp/CefSharp/blob/cefsharp/57/CefSharp.Core/Internals/RenderClientAdapter.h#L182) so I'm comfortable it's well tested.

kpreisser:
Thank for your the background information and the fix! Sorry, I didn't test running as x86 after changing the PR from `RtlMoveMemory` to `RtlCopyMemory`, so I didn't notice it was missing on x86.

(As additional thought, since kernel32's `RtlMoveMemory` gets forwarded to ` NTDLL.RtlMoveMemory`, it might be more reliable to use that one as it is [documented](https://docs.microsoft.com/en-us/windows/win32/devnotes/rtlmovememory), whereas `kernel32.RtlMoveMemory` has only [archived documentation](https://docs.microsoft.com/en-us/previous-versions/windows/desktop/legacy/aa366788%28v%3dvs.85%29). But no need to change that of course since using `memcpy` from the CRT should be fine and is slightly faster.)

Thanks!

amaitland:
> Sorry, I didn't test running as x86 after changing

All good 😄 Makes no sense that it exists on `x64` and not `x86`.

PR #2891, opened 2019-09-06
BrowserSubprocess - Refactor to support .Net Core

amaitland:
Issue #2796

**Summary:**
Refactor the BrowserSubProcess so the core of it's implementation in `CefSharp.BrowserSubprocess.Core`.

**Changes:** 
- Added `BrowserSubprocessExecutable` for use with .Net Core (no WCF)
- Added `WcfBrowserSubprocessExecutable` for use with the existing CefSharp.BrowserSubprocess.exe which supports WCF (conditionally)
      
## How Has This Been Tested?
- The `Javascript Binding` tests pass (both async and sync)
- `Eval` or javascript happens correctly

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Updated documentation

**Checklist:**
- [x] Tested the code(if applicable)
- [ ] Commented my code
- [ ] Changed the documentation(if applicable)
- [x] New files have a license disclaimer
- [x] The formatting is consistent with the project (project supports .editorconfig)

**TODO**
- [x] Add comments to `WcfBrowserSubprocessExecutable`
- [ ] Consider better names for new classes (open to suggestions here)
- [x] Add comments to `Program.cs` that describe the new structure changes.

Issue #2795, opened 2019-05-31
Additional PackageReference Nuget packages


amaitland:
New packages will only support the `PackageReference` format (`VS2017/2019` will be required to use them).

- ~~[ ] Use the newer `runtimes\native` nuget feature for  new [cef.redist.win-x64/cef.redist.win-x86](https://github.com/cefsharp/cef-binary/pull/83)~~
- [x] Restructure packages to only copy `CEF` resources using targets file
- [x] Set `minClientVersion` to `Nuget` version supported by both `VS2017/2019` (make it impossible for older versions to install the packages)
- [x] Create `CefSharp.WinForms.x86/CefSharp.WinForms.x64`
- [x] Create `CefSharp.WPF.x86/CefSharp.WPF.x64`
- [x] Create `CefSharp.OffScreen.x86/CefSharp.OffScreen.x64`

There won't be `AnyCPU` support initially, the older packages will need to be used for that.

References
- https://docs.microsoft.com/en-us/nuget/create-packages/creating-a-package#including-msbuild-props-and-targets-in-a-package
- https://docs.microsoft.com/en-us/nuget/create-packages/supporting-multiple-target-frameworks
- https://docs.microsoft.com/en-au/nuget/guides/create-uwp-packages#adding-the-native-implementation-libraries

amaitland:
https://docs.microsoft.com/en-us/nuget/create-packages/supporting-multiple-target-frameworks

TonyValenti:
Just wanted to let everyone know I'm looking forward to .NET Core 3.0 being fully supported!

lifecoder-phoenix:
If help is needed with this then I'd gladly help out as I'm really looking forward to this feature

amaitland:
Pull Requests are welcome, finishing https://github.com/cefsharp/cef-binary/pull/83 is step one.

amaitland:
When three new individuals/companies signup to [GitHub Sponsors](https://github.com/sponsors/amaitland) then I'll make time to work on the new set of packages.

For a limited time `GitHub` will match any donated amount, even a `$5` per month donation will make a difference. See [About the GitHub Sponsors Matching Fund](https://help.github.com/en/github/supporting-the-open-source-community-with-github-sponsors/about-github-sponsors#about-the-github-sponsors-matching-fund)

amaitland:
Thanks to those that have started sponsoring my efforts!

[...]

PR #2895, opened 2019-09-06
Adjust references in .props files for .NET Core projects

kpreisser:
Issue #2796

**Summary:**
   - Adjusted the `<Reference>` elements in the NuGet `.props` file for better .NET Core support

**Changes:**
   - For .NET Core projects, a reference to `CefSharp.BrowserSubprocess.Core` is added, so that the app executable can run the subprocess logic (after #2891 is done).
   - `<Private>False</Private>` is no longer specified for the CefSharp assemblies in .NET Core projects, as otherwise these references would not be added to the generated `.deps.json` file and the CoreCLR wouldn't load them.
  (Note that this probably means we cannot support `AnyCPU` for .NET Core, but I think we can ignore that for now because e.g. when you publish a self-contained .NET Core application, it will already be platform-specific.)

Note: The `CefSharp.BrowserSubprocess.exe` file (using .NET Framework 4.5.2) is still copied to the output directory for .NET Core projects.
      
## How Has This Been Tested?
Tested by running `build.ps1`, and then using the built `.nupkg` packages (via local NuGet package source) in the `CefSharp.MinimalExample.NetCore.*` projects (see cefsharp/CefSharp.MinimalExample#57), and removing the `<Reference Update="CefSharp">` elements to verify the assemblies are correctly loaded by the runtime.
Additionally, I tested the new packages with the existing `CefSharp.MinimalExample.*` projects to verify classic .NET Framework projects using `packages.config` still work as expected.

## Screenshots (if appropriate):
grafik

## Types of changes

- [ ] Updated documentation

**Checklist:**
- [ ] Commented my code

### DIFF
--- a/NuGet/CefSharp.Common.props
+++ b/NuGet/CefSharp.Common.props
@@ -14,11 +14,17 @@
       <ItemGroup>
         <Reference Include="CefSharp">
           <HintPath>$(MSBuildThisFileDirectory)..\CefSharp\x64\CefSharp.dll</HintPath>
-          <Private>False</Private>
+          <Private Condition="!$(TargetFramework.StartsWith('netcoreapp'))">False</Private>
         </Reference>
         <Reference Include="CefSharp.Core">
           <HintPath>$(MSBuildThisFileDirectory)..\CefSharp\x64\CefSharp.Core.dll</HintPath>
-          <Private>False</Private>
+          <Private Condition="!$(TargetFramework.StartsWith('netcoreapp'))">False</Private>
+        </Reference>
+      </ItemGroup>      
+      <!-- Additional references for .NET Core -->
+      <ItemGroup Condition="$(TargetFramework.StartsWith('netcoreapp'))">
+        <Reference Include="CefSharp.BrowserSubprocess.Core">
+          <HintPath>$(MSBuildThisFileDirectory)..\CefSharp\x64\CefSharp.BrowserSubprocess.Core.dll</HintPath>
         </Reference>
       </ItemGroup>
     </When>
@@ -27,11 +33,17 @@
       <ItemGroup>
         <Reference Include="CefSharp">
           <HintPath>$(MSBuildThisFileDirectory)..\CefSharp\x86\CefSharp.dll</HintPath>
-          <Private>False</Private>
+          <Private Condition="!$(TargetFramework.StartsWith('netcoreapp'))">False</Private>
         </Reference>
         <Reference Include="CefSharp.Core">
           <HintPath>$(MSBuildThisFileDirectory)..\CefSharp\x86\CefSharp.Core.dll</HintPath>
-          <Private>False</Private>
+          <Private Condition="!$(TargetFramework.StartsWith('netcoreapp'))">False</Private>
+        </Reference>
+      </ItemGroup>      
+      <!-- Additional references for .NET Core -->
+      <ItemGroup Condition="$(TargetFramework.StartsWith('netcoreapp'))">
+        <Reference Include="CefSharp.BrowserSubprocess.Core">
+          <HintPath>$(MSBuildThisFileDirectory)..\CefSharp\x86\CefSharp.BrowserSubprocess.Core.dll</HintPath>
         </Reference>
       </ItemGroup>
     </Otherwise>
--- a/NuGet/CefSharp.OffScreen.props
+++ b/NuGet/CefSharp.OffScreen.props
@@ -12,7 +12,7 @@
       <ItemGroup>
         <Reference Include="CefSharp.OffScreen">
           <HintPath>$(MSBuildThisFileDirectory)..\CefSharp\x64\CefSharp.OffScreen.dll</HintPath>
-          <Private>False</Private>
+          <Private Condition="!$(TargetFramework.StartsWith('netcoreapp'))">False</Private>
         </Reference>
       </ItemGroup>
     </When>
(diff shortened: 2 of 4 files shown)


amaitland:
> * `<Private>True</Private>` is no longer specified for the CefSharp assemblies in .NET Core projects, as otherwise these references would not be added to the generated `.deps.json` file and the CoreCLR wouldn't load them.

This part baffles me, will need to do some more digging before I'm prepared to accept this change. Currently I don't have space to install `VS2019`, so it might be a while before that happens.

Happy to merge the additional `BrowserSubProcess` entries separately.

kpreisser:
> > * `<Private>True</Private>` is no longer specified for the CefSharp assemblies in .NET Core projects, as otherwise these references would not be added to the generated `.deps.json` file and the CoreCLR wouldn't load them.

Sorry, this should actually read "`<Private>False</Private>` is no longer specified"; I fixed my initial post.

> This part baffles me, will need to do some more digging before I'm prepared to accept this change. Currently I don't have space to install `VS2019`, so it might be a while before that happens.


Note that it's also possible to just install the [.NET Core 3.0 SDK](https://dotnet.microsoft.com/download/dotnet-core/3.0) and then run e.g. `dotnet build <project> -p:Platform=x64` to build the project, without installing the full VS 2019. For C# debugging in .NET Core projects, Visual Studio Code with the [MS C#](https://marketplace.visualstudio.com/items?itemName=ms-vscode.csharp) extension could be used.

I don't know the technical details, but when you build a .NET Core project, a [`<project>.deps.json` file](https://github.com/dotnet/cli/blob/v3.0.100-preview9.19426.10/Documentation/specs/runtime-configuration-file.md) is generated that contains entries for the references of the project. For example, with `CefSharp.MinimalExample.WinForms.netcore` it looks like this:
```json
{
  "runtimeTarget": {
    "name": ".NETCoreApp,Version=v3.0",
    "signature": ""
  },
  "compilationOptions": {},
  "targets": {
    ".NETCoreApp,Version=v3.0": {
      "CefSharp.MinimalExample.WinForms.netcore/1.0.0": {
        "dependencies": {
          "CefSharp.WinForms": "75.1.142",
          "CefSharp.Core": "75.1.142.0",
          "CefSharp": "75.1.142.0",
          "CefSharp.WinForms.Reference": "75.1.142.0"
        },
        "runtime": {
          "CefSharp.MinimalExample.WinForms.netcore.dll": {}
        }
      },
     ...
      "CefSharp.WinForms.Reference/75.1.142.0": {
        "runtime": {
          "CefSharp.WinForms.dll": {
            "assemblyVersion": "75.1.142.0",
            "fileVersion": "75.1.142.0"
          }
        }
      }
    }
  },
  ...
}
```

Now, when an assembly reference like `CefSharp.WinForms.dll` is set to `<Private>false</Private>` (Copy Local = False), it seems this reference is no longer included in the `.deps.json` file (probably because it also wouldn't be copied and therefore it is expected that it is loaded using another mechanism).

Then, when running the application, it cannot load `CefSharp.WinForms.dll` even though the file exists in the folder near the EXE (as it was copied by the `.targets` file when Platform is `x64` or `x86`):
```
System.IO.FileNotFoundException: 'Could not load file or assembly 'CefSharp.WinForms, Version=75.1.142.0, Culture=neutral, PublicKeyToken=40c4b6fc221f4138'. The system cannot find the file specified.'
```

Setting `<Private>true</Private>` fixed the issue as it adds the entry to the `.deps.json` file (which seems the correct way), but I don't know if there's another way to fix it.

Thank you!

amaitland:
> Note that it's also possible to just install the [.NET Core 3.0 SDK](https://dotnet.microsoft.com/download/dotnet-core/3.0) and then run e.g. `dotnet build <project> -p:Platform=x64` to build the project, without installing the full VS 2019. For C# debugging in .NET Core projects, Visual Studio Code with the [MS C#](https://marketplace.visualstudio.com/items?itemName=ms-vscode.csharp) extension could be used.

Free space permitting I'll install `.Net Core 3.0` when the official release comes out. Hadn't considered using `VS Code`, thanks for the suggestion 👍

kpreisser:
> This part baffles me, will need to do some more digging before I'm prepared to accept this change.

Hi @amaitland, did you already have a chance to look into this?

Thanks!

amaitland:
I did, there are a couple off issues that relates to the problem on `GitHub`, I'll see if I can find the links again.

I'm thinking instead of adding additional hacks to make `.Net Core` work we look at creating a more modern set of packages that specifically support `PackageReference`, issue #2795 will be used to track those changes.

I'm reluctant to merge this based on that, I'd rather not encourage people to use the current packages.

kpreisser:
Thank you!
Yes, I agree creating separate packages is probably the better option.

aesteves900:
Hello Guys, there was a big migration to .NET Core...I have an issue to run the BrowserSubProcess wrote In C# in CefSharp solution, I got badFormatedImageException all the time. I have some difficult to understand why CefSharp.Common.NETCore does not contain CefSharp.BrowserSubprocess.Core.netcore .dll.

I'm using the latest version from myget.org.

Can you help me?

image

image


Thank you!

amaitland:
The project can be compiled with .Net Core,  that part is correct. Issue  #3197 is still open and you can subscribe to it for progress updates. The NetCore set of packages on myget aren't usable yet, the Badimageformatexception is actually misleading.

You can use the current Nuget packages in .Net Core details at https://github.com/cefsharp/CefSharp.MinimalExample#net-core-support

Further questions please ask on https://gitter.im/cefsharp/CefSharp

PR #2924, opened 2019-10-06
Add support for compiling with Visual studio 2019

### LINKED ISSUE #2628: Feature Request - Add Support for compiling with VS2019
When `Visual Studio 2019` is released we can look at adding support for compiling `CefSharp`.

Research needs to be done into what `VC++` changes were made and if the version number has been bumped.

This is only required to compile `CefSharp` from source. There are no plans to change from targeting `VC++ 2015` as the default just yet.

- [ ] Remove `VS2013` support, this will allow for simplification as newer versions support targeting multiple platform toolsets.
- [x] Upgrade `CefSharp.props` to support `VS2019`.
- [ ] Add new `VS2019` target to `build.ps1`

mitchcapper:
Bumped sdk to 18362 CEF 77 now requires that version but we could lower it is desired.
Does not default to building with it.

Resolves #2628

### DIFF
--- a/CefSharp.props
+++ b/CefSharp.props
@@ -5,14 +5,14 @@
     <VisualStudioProductVersion>2013</VisualStudioProductVersion>
     <VisualStudioProductVersion Condition="'$(VisualStudioVersion)'=='14.0'">2015</VisualStudioProductVersion>
     <VisualStudioProductVersion Condition="'$(VisualStudioVersion)'=='15.0'">2017</VisualStudioProductVersion>
-    <VisualStudioProductVersion Condition="'$(VisualStudioVersion)'=='16.0'">2017</VisualStudioProductVersion>
+    <VisualStudioProductVersion Condition="'$(VisualStudioVersion)'=='16.0'">2019</VisualStudioProductVersion>
 
     <PlatformToolset>v120</PlatformToolset>
     <PlatformToolset Condition="'$(VisualStudioVersion)'=='14.0'">v140</PlatformToolset>
     <PlatformToolset Condition="'$(VisualStudioVersion)'=='15.0'">v141</PlatformToolset>
     <PlatformToolset Condition="'$(VisualStudioVersion)'=='16.0'">v142</PlatformToolset>
 	
-	<WindowsTargetPlatformVersion Condition="'$(VisualStudioVersion)'=='16.0'">10.0.17763.0</WindowsTargetPlatformVersion>
+	<WindowsTargetPlatformVersion Condition="'$(VisualStudioVersion)'=='16.0'">10.0.18362.0</WindowsTargetPlatformVersion>
     
     <CefSharpBrowserSubprocessPostBuildEvent>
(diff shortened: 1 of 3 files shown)


CefSharp.props
```diff
@@ -5,14 +5,14 @@
     <VisualStudioProductVersion>2013</VisualStudioProductVersion>
     <VisualStudioProductVersion Condition="'$(VisualStudioVersion)'=='14.0'">2015</VisualStudioProductVersion>
     <VisualStudioProductVersion Condition="'$(VisualStudioVersion)'=='15.0'">2017</VisualStudioProductVersion>
-    <VisualStudioProductVersion Condition="'$(VisualStudioVersion)'=='16.0'">2017</VisualStudioProductVersion>
+    <VisualStudioProductVersion Condition="'$(VisualStudioVersion)'=='16.0'">2019</VisualStudioProductVersion>
```
amaitland: This is where things get complicated, the default `cef.sdk` package only has `VS2015` and `VS2017` versions of the `wrapper` by default. With this change anyone using the packages from `Nuget.org` would be unable to build using `VS2019`

The `VisualStudioProductVersion` variable is used to determine which directory to load the `.lib` file from.

https://github.com/cefsharp/CefSharp/blob/a843407dab56741263f6783c368f85a7bb720c15/CefSharp.Core/CefSharp.Core.vcxproj#L136

Maybe it's possible to have `AppVeyor` build `VS2015`, `VS2017` and `VS2019` versions of the wrapper. See https://www.appveyor.com/docs/windows-images-software/ for list of software installed by image. 

Current image is https://github.com/cefsharp/cef-binary/blob/master/appveyor.yml#L1

mitchcapper: Ah, I could add the v142 to the default for cef-binary building but yeah we would have to change the appveyor default to vs2019 version.  Now it makes sense why you had the newer vs versions targeting the older option.

amaitland: `VS2019` appears perfectly capable of compiling with a `v141` lib file (as was promised by `Microsoft` that they'd be compatible).

I understand from your `Docker` build scripts you'd be looking to build everything with `VS2019`.

Open to suggestions here.

mitchcapper: Targeting v141 is no problem.  With docker could even install V2017 and V2019 side by side without issue.   In addition can easily patch in any changes to the code instead to manually change to v142.   Keeping in on V141 for now would seem the best solution.

amaitland:
I've been thinking about how to best support both use cases, it's probably worth investigating if we can use a File exists check. Check if the package contains the vs2019 compiled lib, set VisualStudioProductVersion to 2019 otherwise fallback to using 2017.

amaitland:
Additional check added in https://github.com/cefsharp/CefSharp/commit/12edb67220639d18fc6e88c170ae7b6a9c006080

I've not tested with a `VS2019` build of the libcefwrapper`, if you experience any problems please let me know.

PR #2947, opened 2019-10-30
WPF - Ensure ZoomLevelProperty is updated on UI thread when Visibility Changed

joaompneves:
Fixes crash in OnVisibilityChanged method
   - Call stack:
Unhandled Exception: System.InvalidOperationException: The calling thread cannot access this object because a different thread owns it.
at System.Windows.Threading.Dispatcher.VerifyAccess()
at System.Windows.DependencyObject.SetCurrentValue(DependencyProperty dp, Object value)
at CefSharp.Wpf.ChromiumWebBrowser.<OnIsVisibleChanged>d__325.MoveNext() in C:\projects\cefsharp\CefSharp.Wpf\ChromiumWebBrowser.cs:line 2039

**Summary:**
   - Fixed a crash on the OnVisibilityChanged method, due to the Zoom dependency property being set on a thread other than the dispatcher thread

**Changes:**
   - Set property on the dispatcher thread
      
## How Has This Been Tested?
   - Manual testing (switching tabs)

### DIFF
--- a/CefSharp.Wpf/ChromiumWebBrowser.cs
+++ b/CefSharp.Wpf/ChromiumWebBrowser.cs
@@ -2026,7 +2026,7 @@ namespace CefSharp.Wpf
                     }
                 });
 
-                if (browser != null)
+                if (browser != null) 
                 {
                     //Fix for #1778 - When browser becomes visible we update the zoom level
                     //browsers of the same origin will share the same zoomlevel and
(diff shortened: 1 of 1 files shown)


amaitland:
> * Unhandled Exception: System.InvalidOperationException: The calling thread cannot access this object because a different thread owns it.

How exactly do you reproduce this exception? In the normal course of events `OnIsVisibleChanged` is called on the `UI` thread, and the sync context is captured when you `await`. I don't see any errors in the `CefSharp.Wpf.Example` and nobody else is reporting an exception. What exactly are you doing differently?

The `UiThreadRunAsync` helper methods should be used rather than directly using the dispatcher.

amaitland:
> We have thousands of NUnit automated tests and this problem occurs frequently. With the fix the problem disappeared.

How does `OnIsVisibleChanged` end up being run on a different thread? Under normal circumstances this shouldn't occur. So what are you doing differently? If it's a unit test then you should be able to attach the debugger and see what's going on.

amaitland:
It's a harmless enough change, though I really would prefer to understand the reason it's required. It would be ideal if there was a code comment explaining the requirement for the Dispatcher Invoke.

- Manually merged in https://github.com/cefsharp/CefSharp/commit/3f285d55250a4bbc57919fb6821da0cd4be286d2
- Removed first `IsDisposed` check, in the expected scenario it would have been checked twice on the `UI` thread.

Issue #2831, opened 2019-07-09
Feature Request - Change property name that Cefsharp injects to window


cppish:
Can I change the property name that Cefsharp injects to window to hide from anti-robot detecting code?
Now it is `CefSharp`, used in `CefSharp.BindObjectAsync`

amaitland:
It's not currently configurable, you are welcome to submit a `PR` for consideration, there's a upper and lower case instance to consider.

https://github.com/cefsharp/CefSharp/blob/cefsharp/75/CefSharp.BrowserSubprocess.Core/CefAppUnmanagedWrapper.cpp#L121

amaitland:
Resolved with #3126

Issue #2977, opened 2019-11-25
Legacy Javascript Binding changed from a global setting to per browser


amaitland:
Previously the `Legacy` JavaScript binding behaviour was configured via a global static property [CefSharpSettings.LegacyJavascriptBindingEnabled](http://cefsharp.github.io/api/84.4.x/html/P_CefSharp_CefSharpSettings_LegacyJavascriptBindingEnabled.htm)

Now it's configured through IJavascriptObjectRepository.Settings.LegacyBindingEnabled
which defaults to [CefSharpSettings.LegacyJavascriptBindingEnabled](http://cefsharp.github.io/api/84.4.x/html/P_CefSharp_CefSharpSettings_LegacyJavascriptBindingEnabled.htm)

When `LegacyBindingEnabled` is enabled, the [ResolveObject](http://cefsharp.github.io/api/84.4.x/html/E_CefSharp_IJavascriptObjectRepository_ResolveObject.htm) event will now be called with `args.ObjectName == Legacy`. Which provides a much more predictable place to register objects with the legacy behaviour.

Example:
```c#
chromiumWebBrowser.JavascriptObjectRepository.Settings.LegacyBindingEnabled = true;

chromiumWebBrowser.JavascriptObjectRepository.ResolveObject += (s, e) =>
{
	if(e.ObjectName == JavascriptObjectRepository.LegacyObjects)
	{
		//ADD YOUR LEGACY
		e.ObjectRepository.Register("myObject", new MyObject(), isAsync:true);
	}
};
```

The [CefSharpSettings.LegacyJavascriptBindingEnabled](http://cefsharp.github.io/api/84.4.x/html/P_CefSharp_CefSharpSettings_LegacyJavascriptBindingEnabled.htm) property will be removed at some point in the future.

--- 
- [x] Raise `ResolveObject` event with `Legacy` param.
- [x] Map out how the property will be set in `WPF`. Make sure it's usable.
- [x] Add Unit test

PR #3126, opened 2020-05-14
Feature - Configure property name that Cefsharp injects to window

Geevo:
**Fixes:** [#2831] 

**Summary:** 
I have added a new feature to the Javascript Binding implementation that gives the ability to configure custom property names for both normal and camelCase binding objects.

I've never written in C++ before so lots of reading in and around the wrapper to grasp an idea of how i would implement this so apologies in advanced if i have missed something with regards to  anything unmanaged related, and of course i'm happy to try and change or refactor anything you're not happy with.

I've tried to make sure this doesn't cause anything breaking changes and default values are used if the properties are never configured.

**Changes:** 
+ Created two new properties in the `CefSharpSettings` and assigned them default values
+ Changes made to the `ManagedCefBrowserAdapter` to retrieve the properties from `CefSharpSettings`
+ Created a new `StringCheck` class/file (cs Internals) to validate the strings do not contain illegal characters
+ Changes to the `CefAppUnmanagedWrapper` to receive the new properties via the `extraInfo` parameter on `OnBrowserCreated`

      
**How Has This Been Tested?**  
Obviously this is a tad bit trickier to test. I had to make changes to all the html code to reflect the new values i was testing with (changes not included in this PR).

Using WinForms Example for testing, i set custom values to the new properties `CefSharpSettings.JavascriptBindingPropertyName` and `CefSharpSettings.JavascriptBindingPropertyNameCamelCase` and run the `Binding Test` accordingly and didn't receive any errors.

I tested with illegal characters to ensure an exception was thrown with an explanation as to why the exception was thrown.

I tested with VS 2019 Professional on Windows 10 (1809)

**Types of changes**

- [ ] Updated documentation

**Checklist:**

- [X] Commented my code
- [X] New files have a license disclaimer


CefSharp.Core/ManagedCefBrowserAdapter.cpp
```diff
@@ -61,6 +61,23 @@ void ManagedCefBrowserAdapter::CreateBrowser(IWindowInfo^ windowInfo, BrowserSet
...
+    auto jsBindingPropertyNameCamelCase = CefSharpSettings::JavascriptBindingPropertyNameCamelCase;
+    if (!StringCheck::EnsureLettersAndNumbers(jsBindingPropertyName) || !StringCheck::EnsureLettersAndNumbers(jsBindingPropertyNameCamelCase))
+    {
+        throw gcnew InvalidOperationException("CefBrowserHost::CreateBrowser invalid or illegal characters used for binding property names. Alphanumeric and underscores characters only.");
+    }
+
+    extraInfo->SetString("JsBindingPropertyName", StringUtils::ToNative(jsBindingPropertyName));
+    extraInfo->SetString("JsBindingPropertyNameCamelCase", StringUtils::ToNative(jsBindingPropertyNameCamelCase));
+
+    if (!CefBrowserHost::CreateBrowser(*cefWindowInfoWrapper->GetWindowInfo(), _clientAdapter.get(), addressNative,
```
amaitland, line 75: Unit tests are failing as CreateBrowser is incorrectly being called twice.

Geevo: Oops! Ok I've removed the duplicate method.

amaitland:
> Obviously this is a tad bit trickier to test. I had to make changes to all the html code to reflect the new values i was testing with (changes not included in this PR).

I think this is an argument against adding properties to `CefSharpSettings`, these should probably be per browser. Having `LegacyJavascriptBindingEnabled` and `WcfEnabled` as global was a mistake and once we come up with a per browser place to put these settings those should be migrated as well. 

I'm not sure exactly where they should go, possible a settings object should be added to the `JavascriptObjectRepository`.

Further comments inline.

CefSharp/CefSharpSettings.cs
```diff
@@ -20,8 +20,30 @@ static CefSharpSettings()
...
 
+        /// <summary>
+        /// Configure the browser binding property name.
+        /// Use a custom property name when calling registred objects, example 'CefSharp.BindObjectAsync'.
+        /// Alphanumeric and underscores characters only.
+        /// The default value is 'CefSharp'
+        ///
+        /// NOTE: There are two properties to be mindful of, this and <see cref="JavascriptBindingPropertyNameCamelCase"/>
+        /// </summary>
+        public static string JavascriptBindingPropertyName { get; set; }
```
amaitland, line 35: - We we need two properties for this? Camel case can be derived from the full name
- What happens if both properties are exactly the same?

If they both end up being the same name we should only generate one set of objects.

CefSharp/Internals/StringCheck.cs
```diff
@@ -0,0 +1,30 @@
+// Copyright © 2020 The CefSharp Authors. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style license that can be found in the LICENSE file.
+
+// Licensed to the .NET Foundation under one or more agreements.
```
amaitland, line 5: What's with the extra license block?

Geevo: The licence was copied from the following file I randomly selected.

https://github.com/cefsharp/CefSharp/blob/940034314b5961f8b80d84d8da7705eaa1dd95e6/CefSharp/Internals/PathCheck.cs#L5

I can submit a separate PR to remove from this page?

CefSharp/CefSharpSettings.cs
Geevo: How about exposing a single property and then manipulate the string to pascal/camel case on the first char of the property before injecting that way you can keep both objects and cater for either preference?

For example:
https://stackoverflow.com/a/36579722

Or simply leave it down to preference completely, and only have a single object.

CefSharp/Internals/StringCheck.cs
amaitland: That file has code copied from the .Net framework source. No need for additional license unless you've used code from somewhere else.

CefSharp/CefSharpSettings.cs
amaitland: Single property sounds fine, if the first character is already lower we should only create one set of objects. 

https://github.com/cefsharp/CefSharp/blob/master/CefSharp/Internals/JavascriptObjectRepository.cs#L543 should be refactored if we add an additional method for camel case conversion.

Geevo:
I think it maybe worth having a method to set the `WindowPropertyName`. That way you check if the browser is initialised beforehand, otherwise it could be too late and injected property name will remain the default configuration.

Maybe also perhaps then have get only property with the last known set name. 
And throw an exception when trying to set when the browser is already initialised?

CefSharp/Internals/JavascriptObjectRepository.cs
```diff
@@ -404,6 +411,8 @@ internal bool TrySetProperty(long objectId, string name, object value, out strin
         /// <param name="camelCaseJavascriptNames">camel case the javascript names of properties/methods</param>
         private void AnalyseObjectForBinding(JavascriptObject obj, bool analyseMethods, bool analyseProperties, bool readPropertyValue, bool camelCaseJavascriptNames)
         {
+            var isWindowPropertyFirstCharacterLowercase = StringCheck.IsFirstCharacterLowercase(WindowPropertyName);
+
```
amaitland, line 415: Please revert this set of changes. My previous comment was simply to reduce code duplication, I meant extract the core part of the function into say StringCheck(we'll need to me StringCheck to be more generic)

CefSharp/IJavascriptObjectRepository.cs
```diff
@@ -13,6 +13,7 @@ namespace CefSharp
     /// </summary>
     public interface IJavascriptObjectRepository : IDisposable
     {
+        string WindowPropertyName { get; set; }
```
amaitland, line 16: There will be multiple settings to add, so an object that holds the settings and performs the validation is what I can in mind.

CefSharp/Internals/JavascriptObjectRepository.cs
Geevo: Noted. reverted back

CefSharp/IJavascriptObjectRepository.cs
Geevo: I have now created a settings object currently only containing the `WindowPropertyName`.

amaitland:
I've implemented #2977 which is the reason you'll see conflicting changes in this.

- Went with a slightly different name in `JavascriptBindingSettings` as opposed to `JavascriptObjectRepositorySettings` (I'm still not sure which name I like best, we can change this any time up until this is included in a release branch).
- Used a `Freezable` implementation to throw exception of property modified after browser is created.

See https://github.com/cefsharp/CefSharp/commit/a70abc4437384a62bd187590636b6706250c93de for relevant commit.

I've squashed your changes and they're included in the https://github.com/amaitland/CefSharp/commits/feature/jsbpropertyname branch with some additional changes

- Validate the name in the property setter to provide immediate feedback. This will provide for a clearer call stack in the exception.
- `JsBindingPropertyName` is the name I chose, though I'm not convinced that's great either. I think we still need a better name. Really it's the name of the object that's created in the global javascript scope. Open to suggestions on this one.
- If `JsBindingPropertyName` is empty then we'll use the current behaviour as default.

Still need to fill in all the `xml doc` blanks and add some unit tests.

I'll finish the rest when I find some time.

Geevo:
> * `JsBindingPropertyName` is the name I chose, though I'm not convinced that's great either. I think we still need a better name. Really it's the name of the object that's created in the global javascript scope. Open to suggestions on this one.

Perhaps `JsBindingGlobalObjectName` would better represent the definition of its usage.

As per your description above and definition of the `window` object described here: 
(https://developer.mozilla.org/en-US/docs/Glossary/Global_object)

`The window object is the Global Object in the Browser.`

Nice touch of the Freezable base implementation 👍

amaitland:
> Perhaps `JsBindingGlobalObjectName` would better represent the definition of its usage.

Much better, thanks 👍

> Nice touch of the Freezable base implementation

Borrowing the idea from `WPF` 😄 A few other places we should be able to make use of the base class.

amaitland:
Manually merged in https://github.com/cefsharp/CefSharp/commit/66691105b43389eb9df43dca44b6996a7cc8e8a5 with additional changes in https://github.com/cefsharp/CefSharp/commit/ae41e10d712f9387321c32d0f86895144d6b5d1d

I ended up with the property named as `JavascriptBindingApiGlobalObjectName `, it's long I know, hopefully it's at least semi meaningful. Example below.

```c#
var settings = browser.JavascriptObjectRepository.Settings;
settings.JavascriptBindingApiGlobalObjectName = "customApi";
```

I've only got a single very basic test so far https://github.com/cefsharp/CefSharp/commit/ae41e10d712f9387321c32d0f86895144d6b5d1d#diff-59940d85e6a445f319b6a2b0a08bdf9cR77

If anyone has problems then please provide a test case.

PR #3173, opened 2020-07-07
Issue #2053 - Add accessibility support in WPF

justindarc:
**Fixes:** [issue-number] 
Fixes: #2053 

   - It is now possible to run the CefSharp.Wpf.Example and use a screen reader such as [Microsoft Narrator](https://en.wikipedia.org/wiki/Microsoft_Narrator), [NVDA](https://www.nvaccess.org/about-nvda/) or [JAWS](https://www.freedomscientific.com/products/software/jaws/) to traverse and navigate the web content hierarchy.

   - In `CreateOffscreenBrowserWindowInfo` in ChromiumWebBrowser.cs, it is necessary to use `SetAsChild` instead of `SetAsWindowless` in order for the UIAutomation/MSAA trees from CEF to be exposed.
   - In CefSettings.cs, the `force-renderer-accessibility` command line argument needs to be added in order for CEF to expose its UIAutomation/MSAA trees.
      
**How Has This Been Tested?**  
   - I have tested this by running the CefSharp.Wpf.Example app on a Windows 10 64-bit machine while using Microsoft Narrator, NVDA and JAWS (trial version) to navigate the web content.

**Types of changes**

- [ ] Updated documentation

**Checklist:**

- [X] Commented my code

### DIFF
--- a/CefSharp.Wpf/ChromiumWebBrowser.cs
+++ b/CefSharp.Wpf/ChromiumWebBrowser.cs
@@ -2018,7 +2018,14 @@ namespace CefSharp.Wpf
         protected virtual IWindowInfo CreateOffscreenBrowserWindowInfo(IntPtr handle)
         {
             var windowInfo = new WindowInfo();
-            windowInfo.SetAsWindowless(handle);
+            if (handle == IntPtr.Zero)
+            {
+                windowInfo.SetAsWindowless(handle);
+            }
+            else
+            {
+                windowInfo.SetAsChild(handle);
+            }
             return windowInfo;
         }
 
--- a/CefSharp.Wpf/CefSettings.cs
+++ b/CefSharp.Wpf/CefSettings.cs
@@ -25,6 +25,9 @@ namespace CefSharp.Wpf
             //Disable GPU Compositing
             //Issue https://github.com/cefsharp/CefSharp/issues/3114
             CefCommandLineArgs.Add("disable-gpu-compositing");
+
+            // Enables support for screen readers like Microsoft Narrator, NVDA and JAWS.
+            CefCommandLineArgs.Add("force-renderer-accessibility");
         }
     }
 }


justindarc:
I have left a comment on the #2495 PR that previously attempted to partially solve this problem where I explain why I believe that is the wrong technical approach. I also explain further in-depth my investigations that led me to this solution: https://github.com/cefsharp/CefSharp/pull/2495#issuecomment-654981429

justindarc:
I just force-pushed a new commit that fixes the one test failure. All tests pass locally now.

amaitland:
I have plans to release a native `Win32` implementation for `WPF` using a `HwndHost` that project is available at https://github.com/cefsharp/CefSharp.Wpf.HwndHost you are welcome to contribute to that project. It's pretty much ready to release a `Beta` version. I'm hoping to provide a drop in replacement for the `WPF` version. Alternatively you can use the `WinForms` version in a `WindowsFormsHost`.

> * In `CreateOffscreenBrowserWindowInfo` in ChromiumWebBrowser.cs, it is necessary to use `SetAsChild` instead of `SetAsWindowless` in order for the UIAutomation/MSAA trees from CEF to be exposed.

Fundamentally the `CefSharp.Wpf` implementation uses `OffScreen Rendering(OSR)` where every frame is rendered as a bitmap, mouse/keyboard/touch events are forwarded to the browser. This allows for transparency/transforms, basically a native `WPF` implementation. I'm the first to admit it's flawed and there are many limitations. For the screen reader use case it's unlikely we can provide a perfect solution, using a native `Win32` implementation is likely the best option.

> * In CefSettings.cs, the `force-renderer-accessibility` command line argument needs to be added in order for CEF to expose its UIAutomation/MSAA trees.

It's my understanding there is a performance impact when doing this and I think it's best to leave upto the individual user to enable this themselves. If you'd like to write a `Guide` to using `CefSharp` with a screen reader(accessibility) to provide better documentation then that would be welcomed.

> * I have tested this by running the CefSharp.Wpf.Example app on a Windows 10 64-bit machine while using Microsoft Narrator, NVDA and JAWS (trial version) to navigate the web content.

Did you try resizing the window? Where's the tab control and the menu after you applied your change?

I am going to decline this `PR` as it's a breaking change and overwrites the `WPF` windows child contents.

CefSharp.Wpf/ChromiumWebBrowser.cs
```diff
@@ -2018,7 +2018,14 @@ protected virtual bool CreateOffscreenBrowser(Size size)
...
         {
             var windowInfo = new WindowInfo();
-            windowInfo.SetAsWindowless(handle);
+            if (handle == IntPtr.Zero)
+            {
+                windowInfo.SetAsWindowless(handle);
+            }
+            else
+            {
+                windowInfo.SetAsChild(handle);
```
amaitland, line 2027: The `handle` in this case is the `HWND` for the `WPF Window`, you've effectively overridden the contents of the `Window`. There are many other changes that are required like resizing/minamize/maxamise hence the creation of https://github.com/cefsharp/CefSharp.Wpf.HwndHost

justindarc:
@amaitland no problem. I didn't notice that this approach wasn't behaving correctly in the same window as other native widgets as I was running just an example app with a full-window webview. That being said, not having the Chromium window attached to the WPF window is absolutely the root cause of the lack of accessibility support. Thank you for pointing me to the `HwndHost` repo. Interestingly enough, I was experimenting with converting the `ChromiumWebBrowser` control into an `HwndHost` a few days ago. I do think using the `HwndHost` is the proper approach for apps that cannot use the WinForms variant of the control that also need accessibility support.

Since I'm primarily a macOS/iOS developer, I do have a question. Will it still be possible to support transparency with the `HwndHost` control? I admit I haven't investigated this enough to know, so if you could shed some light on this, that would be helpful. Conversely, would it be possible to support transparency in the WinForms control? I'm aware of the "airspace" issue with rendering on Windows, so I'm not sure if that is a factor in whether or not transparency is possible.

Thanks for taking the time to look at this. 👍

justindarc:
@amaitland one last thing. I would be more than happy to help contribute to get CefSharp.Wpf.HwndHost to a state where it can be released. Do you have a list of remaining issues?

amaitland:
> Will it still be possible to support transparency with the `HwndHost` control? I admit I haven't investigated this enough to know, so if you could shed some light on this, that would be helpful. Conversely, would it be possible to support transparency in the WinForms control?

It's my understanding that `CEF` only supports transparency when using `OSR` which is used by `WPF` (I believe that's actually broken in version 83 at the moment https://bitbucket.org/chromiumembedded/cef/issues/2938/osr-transparent-background-broken-since).

See also
- https://bitbucket.org/chromiumembedded/cef/issues/2315/allow-transparent-browser-for-windowed 
- https://magpcss.org/ceforum/apidocs3/projects/(default)/_cef_settings_t.html#background_color (windowed rendering is used by the `WinForms/HwndHost` versions.

If you search on https://www.magpcss.org/ceforum/index.php you'll probably find some discussion on the topic, I'm pretty sure it's not possible currently.

amaitland:
> I would be more than happy to help contribute to get CefSharp.Wpf.HwndHost to a state where it can be released

@justindarc I've pushed a `-pre` release package to `Nuget.org` if you are interested it should appear shortly https://www.nuget.org/packages/CefSharp.Wpf.HwndHost/83.4.20-pre (have to wait for it to pass validation).

Any problems/issues please report them on https://github.com/cefsharp/CefSharp.Wpf.HwndHost

justindarc:
Awesome! Thank you!

jjbright:
I'm trying to automate the USD 4.1 (MS Dynamics CRM 365) which is using a CefSharp to render HTML content. 
While exploring the internet I found this GitHub CefSharp I am looking for some help.

1. Is this github project is for UI Automation purposes?
2. How to enable the force-renderer-accessibility in cefsharp, I do have the CefSharp.xml inside the USD 4.1 installation?

amaitland:
@jjbright Rather than asking questions on a declined pull request please use one of the options listed at https://github.com/cefsharp/CefSharp#contact

Issue #3114, opened 2020-04-28
WPF - Resize Glitches (Image glitches during resize)


amaitland:
- **What version of the product are you using?**
    81.3.100

- **What architecture x86 or x64?**
  Both
    
- **On what operating system?**
  Win 10

- **Are you using `WinForms`, `WPF` or `OffScreen`?**
  WPF
    
- **What steps will reproduce the problem?**
  - Open CefSharp.Wpf.Example
  - Change window state to normal    
  - Resize the browser fairly aggressively

- **What is the expected output? What do you see instead?**
  There are some glitches when resizing that aren't present when testing with `cefclient`

- **Please provide any additional information below.**
    DPI scale 150%

- **Does this problem also occur in the `CEF` Sample Application**
   No. The CEF Sample application isn't perfect when resizing, there's plenty of times the outer border is blank waiting for the image to resize correctly. The glitches aren't present to the same degree though.

amaitland:
Use the `disable-gpu-compositing` command line arg should workaround the issue in the short term.

amaitland:
Integrating `CEF` into the `WPF` message loop also appears to eliminate the glitches. Further debugging required.

Set `multiThreadedMessageLoop = false` in the example at https://github.com/cefsharp/CefSharp/blob/cefsharp/83/CefSharp.Wpf.Example/App.xaml.cs#L28

rstedman:
Any update on this bug? I've found that CefSharp 79+ causes significant rendering artifacts when resizing a window. I get the same behavior with or without the --disable-gpu-compositing flag. example:

![resize_artifacts](https://user-images.githubusercontent.com/5440997/96156167-c7a5ea00-0ede-11eb-9c55-5d99767e601a.gif)

I can confirm that integrating cef into the wpf message loop gets rid of the rendering issues, but I am unable to do that in my usage of cefsharp as it causes some other issues.

I am able to update my application to 75.1.143 (from 67.0.0) without experiencing this rendering issue. Any higher version and this issue starts occurring.

I would really like to take the latest cefsharp version, as it has a fix for a css grid layout bug that occurs when viewing a page using cefsharp that doesn't occur in browsers (https://bugs.chromium.org/p/chromium/issues/detail?id=1051039). This resize bug is the only issue currently blocking me from taking latest.

amaitland:
[...]

PR #3272, opened 2020-11-03
WPF - Add InteropBitmapRenderHandlerEx

timaiv:
**Fixes:**
https://github.com/cefsharp/CefSharp/issues/3114

This is my vision (**rather a hint than a pull request** how to speed up rendering)

**Summary:**
     Benefits:
     1) Rendering will be much faster. Only required bytes are copied (significanly improving (up to 50x faster) of fullscreens browsers, 2k, 4k monitors).
Now, cursor animation or text selection will take copying buffers only few ticks (will be), than 100ms (now) on 2k monitor.
     2) Only one dispatcher operation active.
     3) Dispatcher operation cancelling on dispose.
     4) If too many updates by cef thread with different sizes and small speed by dispatching, any image WILL be rendered, unlike in default implementation.
     5) No lambdas, no capturing.
     6) Image safe type checking
      
MethodImplOptions is used to allow set internal modifier and now u can use 1 benefit in other render handles, compiler will rather inject it directly.

TODO: @amaitland  change dirty rect struct passing in whole project by ref, or using c# 7.2 "readonly in structs"

**Types of changes**

- [ ] Updated documentation

**Checklist:**

- [x] Commented my code

Also i spent a lot of time (around 5 days) in recognizing cefsharp/ duce rendering and created many improvements that will be added in future.

### DIFF
--- a/CefSharp.Wpf/CefSharp.Wpf.csproj
+++ b/CefSharp.Wpf/CefSharp.Wpf.csproj
@@ -119,6 +119,7 @@
     <Compile Include="Rendering\Experimental\CompositionTargetRenderHandler.cs" />
     <Compile Include="Rendering\DirectWritableBitmapRenderHandler.cs" />
     <Compile Include="Rendering\Experimental\ByteArrayWritableBitmapRenderHandler.cs" />
+    <Compile Include="Rendering\Experimental\InteropBitmapRenderHandlerEx.cs" />
     <Compile Include="Rendering\InteropBitmapRenderHandler.cs" />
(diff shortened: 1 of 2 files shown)


amaitland:
Thanks for creating a PR with your idea 👍 I'll have a look in detail a bit later.

> TODO: @amaitland change dirty rect struct passing in whole project by ref, or using c# 7.2 "readonly in structs"

You can target a maximum of c# 7.1 currently, so no read-only struct support for .Net 4.5.2  for our default builds that use VS2015 

At this point in time I have no plans to change all the struct to be passed by ref.

Issue #3270, opened 2020-11-02
Core - Improve IPopupFeatures mapping and xml doc 


SirIntellegence:
In [`IPopupFeatures`](https://github.com/cefsharp/CefSharp/blob/57f00e0627332f8632fcd4245b7893d321291a10/CefSharp/IPopupFeatures.cs#L1), there are some properties that match the pattern of "*Set". The documentation for these items are terrible since they are ints instead of bools due to [this commit](https://bitbucket.org/chromiumembedded/cef/commits/72660c6fed539ff8e901825aca890bea195ad31d). If they were bools, that would have been easy to figure out, but they are not and I had to dig around and found this [lucky file](https://magpcss.org/ceforum/apidocs/projects/(default)/cef_types.h.html) to see that they were bools. Also, not all of the fields are accounted for (for reference). Could you change those to bools or give them proper documentation?

- **What version of the product are you using?**
    - Nuget: 85.3.130.0
- **What architecture x86 or x64?**
    Both, but that likely doesn't matter since since this is compile time
    
- **On what operating system?**
   Win10

- **Are you using `WinForms`, `WPF` or `OffScreen`?**
    WinForms

amaitland:
Your welcome to submit a PR. 

Just change the interface and https://github.com/cefsharp/CefSharp/blob/76c7918b83c2485f9ab664f343399b71368fa0bf/CefSharp.Core/PopupFeatures.h

amaitland:
See https://github.com/chromiumembedded/cef/blob/4240/include/internal/cef_types.h#L2077 for M86 branch header. 

As CEF doesn't document what the fields are actually for it's very difficult to provide accurate documentation. 

Looking a little closer and the references you've provided are all outdated. Should the set properties be Boolean? I have no idea.

amaitland:
A quick look through the CEF source and the `Set` properties should probably be removed and the `x,y,width,height` properties changed to `Nullable<int>`

https://github.com/chromiumembedded/cef/blob/cae194f493328a9890b27533ce1dc661e0d4b042/libcef/browser/browser_info_manager.cc#L36

SirIntellegence:
From what I could tell, they were bools, but they were changed to ints to be compatible with another compiler. C++ does implicit casts to bool, so it wasn't a problem for them, but C# does not. I will see if I can make a pull request.

SirIntellegence:
I am tempted to turn those into nullable ints. Would that be ok to do?

[...]

PR #3273, opened 2020-11-04
Fixed issue #3270 (Core - Improve IPopupFeatures mapping and xml doc)

SirIntellegence:
Now using nullable ints for optional values in IPopupFeatures

**Fixes:**   #3270  

**Summary:**
   - Now using nullable ints for optional values in IPopupFeatures

**Changes:**
   - Changed X, Y, Width, and Height to return nullable values
      - Also removed the corresponding *Set properties (XSet, YSet, WidthSet, HeightSet) since those are now covered by the nullable ints
      
**How Has This Been Tested?**  
There were no existing tests for this particular class, so I used `ExperimentalLifespanHandler` in CefSharp.Wpf.Example with a breakpoint at line 39 (the start of `OnBeforePopup`. I put it in `BrowserTabViewModel` in the same project after line 200. I then navigated to https://www.w3schools.com/html/html_links.asp and clicked "Try it Yourself" and made sure that sane values were returned from the `popupFeatures` instance. Nothing crashed, so I think we should be good.

**Types of changes**
- [x] Updated documentation

**Checklist:**

- [x] Commented my code

### DIFF
--- a/CefSharp/IPopupFeatures.cs
+++ b/CefSharp/IPopupFeatures.cs
@@ -2,6 +2,8 @@
 //
 // Use of this source code is governed by a BSD-style license that can be found in the LICENSE file.
 
+using System.Collections.Generic;
+
 namespace CefSharp
 {
     /// <summary>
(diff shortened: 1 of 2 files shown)


amaitland:
Thanks 👍 

Can you please remove the commented out properties, they are no longer available. 

https://github.com/chromiumembedded/cef/blob/4240/include/internal/cef_types.h#L2077

CefSharp/IPopupFeatures.cs
```diff
@@ -15,56 +17,28 @@ public interface IPopupFeatures
...
         /// </value>
-        int X { get; }
-        /// <summary>
-        /// Gets the set.
-        /// </summary>
-        /// <value>
-        /// The x coordinate set.
-        /// </value>
-        int XSet { get; }
+        int? X { get; }
```
amaitland, line 20: Can you change these to Nullable<int>, should hopefully resolve the build error in the ref assembly project. 
Thanks

SirIntellegence: Ah, Ok. I will get to that tomorrow.

amaitland:
RefAssembly generation was being difficult, still needs a little fine tuning, code was fine. Manually merged in https://github.com/cefsharp/CefSharp/commit/f9971730bea87d754c0a51082c98ad1bba8ffafb with follow up commit https://github.com/cefsharp/CefSharp/commit/6b5feafb708b739d739472f33f2fa438aa6ae320

PR #3291, opened 2020-11-23
Fix Task Deadlock During Interception

AndrewMD5:
**Summary:**
Currently custom interceptors are forced to call `.Result` as there is no async overload. In some scenarios this can lead to a deadlock as the task attempts to schedule it's continuation onto a thread that is being blocked by the call to Result. You can also trigger a deadlock by returning a Task directly without awaiting as CefSharp then calls `.Result` internally leading to the same issue mentioned prior.

**Changes:**
I have modified the method intercept call to allow for overloading with async. This now allows the entire execution to use async / await properly. A single call to `.Result` is made safely only in BrowserProcessService; by this time the results should be fully resolved so there is no risk of a deadlock.

It also disallows Task from being returned without first awaiting them. This removes the need for CefSharp to try and haphazardly resolve task itself. 

      
**How Has This Been Tested?**  
In our production app [Rainway](https://rainway.com/).

**Checklist:**
- [x] Commented my code

### DIFF
--- a/CefSharp.Example/ModelBinding/MethodInterceptorLogger.cs
+++ b/CefSharp.Example/ModelBinding/MethodInterceptorLogger.cs
@@ -4,17 +4,25 @@
 
 using System;
 using System.Diagnostics;
+using System.Threading.Tasks;
 using CefSharp.ModelBinding;
 
 namespace CefSharp.Example.ModelBinding
 {
     public class MethodInterceptorLogger : IMethodInterceptor
     {
+        public bool IsAsync => false;
+
         object IMethodInterceptor.Intercept(Func<object[], object> method, object[] parameters, string methodName)
         {
-            object result = method(parameters);
+            var result = method(parameters);
             Debug.WriteLine("Called " + methodName);
             return result;
+        }
+
+        public Task<object> InterceptAsync(Func<object[], object> method, object[] parameters, string methodName)
+        {
+            throw new NotImplementedException();
         }
     }
 }
--- a/CefSharp/AsyncExtensions.cs
+++ b/CefSharp/AsyncExtensions.cs
@@ -4,6 +4,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Threading;
 using System.Threading.Tasks;
 
 namespace CefSharp
@@ -13,6 +14,35 @@ namespace CefSharp
     /// </summary>
     public static class AsyncExtensions
     {
+        internal static T GetResultSafely<T>(this Task<T> task)
+        {
+            if (SynchronizationContext.Current == null)
+            {
+                return task.Result;
+            }
+
+            if (task.IsCompleted)
+            {
+                return task.Result;
+            }
+
+            var tcs = new TaskCompletionSource<T>();
+            task.ContinueWith(t =>
+            {
+                var ex = t.Exception;
+                if (ex != null)
+                {
+                    tcs.SetException(ex);
+                }
+                else
+                {
+                    tcs.SetResult(t.Result);
+                }
+            }, TaskScheduler.Default);
+
+            return tcs.Task.Result;
+        }
+
         /// <summary>
         /// Deletes all cookies that matches all the provided parameters asynchronously.
         /// If both <paramref name="url"/> and <paramref name="name"/> are empty, all cookies will be deleted.
--- a/CefSharp/Internals/ConcurrentMethodRunnerQueue.cs
+++ b/CefSharp/Internals/ConcurrentMethodRunnerQueue.cs
@@ -3,6 +3,7 @@
 // Use of this source code is governed by a BSD-style license that can be found in the LICENSE file.
 
 using System;
+using System.Diagnostics;
 using System.Threading;
 using System.Threading.Tasks;
 
@@ -33,9 +34,10 @@ namespace CefSharp.Internals
 
         public void Enqueue(MethodInvocation methodInvocation)
         {
-            var task = new Task(() =>
+          
+            var task = new Task(async () =>
             {
-                var result = ExecuteMethodInvocation(methodInvocation);
+                var result = await ExecuteMethodInvocation(methodInvocation).ConfigureAwait(false);
 
                 //If the call failed or returned null then we'll fire the event immediately
                 if (!result.Success || result.Result == null)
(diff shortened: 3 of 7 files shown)


amaitland:
Thanks for the `PR`, I'll have a look in more detail over the next day or so.

> In some scenarios this can lead to a deadlock as the task attempts to schedule it's continuation onto a thread that is being blocked by the call to Result. You can also trigger a deadlock by returning a Task directly without awaiting as CefSharp then calls `.Result` internally leading to the same issue mentioned prior.

Are you able to provide some unit tests that reproduce the two cases?

> CefSharp then calls `.Result` internally leading to the same issue mentioned prior.

From memory we are only accessing `Result` in a `ContinueWith` statement, can you provide a link to the source where we're accessing `Result` without the `Task` being completed?

> A single call to `.Result` is made safely only in BrowserProcessService; by this time the results should be fully resolved so there is no risk of a deadlock.

Are you using the sync `JSB` implementation in your application?

emily33901:
Hi

> Are you able to provide some unit tests that reproduce the two cases?

Im not well versed in this testing framework but any function that queues something to run on the threadpool it can (or will) eventually deadlock itself.


Returning a task here will block all other tasks from running whilst that task is being syncronously waited on. In the worst case it will completely deadlock straight away if a function waits for a result from a `Task.Run(...)` whilst if you return a task cefsharp will block any other requests from happening whilst it completes...

For example if I change some of the examples code and enable the not simple subprocess - AsyncBoundObject.cs's long running task example into:

```cs
        public async Task<string> test()
        {
            await Task.Run(async () => await Task.Delay(10000));
            return "Nice!";
        }

        public object DoSomething()
        {
            return test();
        }
```

This will block any other tests from running until it completes.

Same with 

```cs
        public async Task<string> test()
        {
            await Task.Delay(10000);
            return "Nice!";
        }

        public object DoSomething()
        {
            return test();
        }
```

Same with

```cs
        public async Task<string> test()
        {
            await Task.Delay(10000);
            return "Nice!";
        }

        public object DoSomething()
        {
            return Task.Run(async () => await test());
        }
```

Same with 

```cs
        public async Task<string> test()
        {
            await Task.Delay(10000);
            return "Nice!";
        }

        public object DoSomething()
        {
            return Task.Run(() => test().Result);
        }
```

(which i should note is the worst offender completely locking the threadpool  but its a known worst practise to do that)

I should note here that whilst you might be screaming for me to use async on `DoSomething()`, that also does not fix the issue.

amaitland:
> Im not well versed in this testing framework

Can you provide just a generic code example that I can turn into a Unit test?

> any function that queues something to run on the threadpool it can (or will) eventually deadlock itself.

Can you elaborate on this please. 

> Returning a task here will block all other tasks from running whilst that task is being syncronously waited on. In the worst case it will completely deadlock straight away if a function waits for a result from a `Task.Run(...)` whilst if you return a task cefsharp will block any other requests from happening whilst it completes...

Which `IMethodRunnerQueue` implementation are you using exactly? The default `MethodRunnerQueue` implementation will queue tasks for execution sequentially.

The `ConcurrentMethodRunnerQueue` implementation enqueues tasks on the `ThreadPool`. I'm not sure how you'd block this implementation exactly.

Can you confirm which implementation you are using in your application?

> For example if I change some of the examples code and enable the not simple subprocess

What exactly is the `not simple subprocess` that you've enabled?

> This will block any other tests from running until it completes.

It depends on how you ran the tests, if the default `MethodRunnerQueue` was used then this would likely be expected.

> I know for a fact we are not using the sync implementation of the jsb.

Thanks for confirming. As the changes to the `sync` implementation haven't been tested we should avoid those where possible. Specially as there are no `Unit tests` as part of this `PR`.

I believe we should leave the `TryCallMethod` implementation unchanged and add a `TryCallMethodAsync` implementation instead, use it only for the `IMethodRunnerQueue` implementation you have used and tested. 

Are you able to provide an example of the `IMethodInterceptor` implementation you are using?

To limit the breaking changes introduced here we should create an `IAsyncMethodInterceptor` that inherits from `IMethodInterceptor` and use type checking to determine which method is called.

amaitland:
I am aware of a problem with the `MethodRunnerQueue` implementation (took me a while to find the reproduction as the user didn't create an actual issue). `Task.Run` was incorrectly attempting to attach to the custom `TaskScheduler` and deadlocking, #3293 (which should now be resolved).

emily33901:
> > Im not well versed in this testing framework
> 
> Can you provide just a generic code example that I can turn into a Unit test?

Any of the code examples above will block up the concurrentmethodqueue

> > any function that queues something to run on the threadpool it can (or will) eventually deadlock itself.
> 
> Can you elaborate on this please.

To be honest this is probably just a reflection on what can normally happen on the threadpool if youre not careful with async await and blocking - Ill have to check through our codebase again and see whether anything else could also be creating this behaviour.

> > Returning a task here will block all other tasks from running whilst that task is being syncronously waited on. In the worst case it will completely deadlock straight away if a function waits for a result from a `Task.Run(...)` whilst if you return a task cefsharp will block any other requests from happening whilst it completes...
> 
> Which `IMethodRunnerQueue` implementation are you using exactly? The default `MethodRunnerQueue` implementation will queue tasks for execution sequentially.
> 
> The `ConcurrentMethodRunnerQueue` implementation enqueues tasks on the `ThreadPool`. I'm not sure how you'd block this implementation exactly.
> 
> Can you confirm which implementation you are using in your application?

We use the Concurrent method runner and for this test code that im trying to use to recreate this i am also using the concurrent method runner.

From any of the little snippets that I had before will cause all the other methods that need executing to wait for the `DoSomething()` function to return, whether async or not, whether returning a task or not.

> What exactly is the `not simple subprocess` that you've enabled?

the `simpleSubProcess` bool in Program.cs of the winforms example code

> > This will block any other tests from running until it completes.
> 
> It depends on how you ran the tests, if the default `MethodRunnerQueue` was used then this would likely be expected.

I believe that the example does also use the concurrent method runner - it sets `CefSharpSettings.ConcurrentTaskExecution = true;`

> > I know for a fact we are not using the sync implementation of the jsb.
> 
> Thanks for confirming. As the changes to the `sync` implementation haven't been tested we should avoid those where possible. Specially as there are no `Unit tests` as part of this `PR`.
> 
> I believe we should leave the `TryCallMethod` implementation unchanged and add a `TryCallMethodAsync` implementation instead, use it only for the `IMethodRunnerQueue` implementation you have used and tested.
> 
> Are you able to provide an example of the `IMethodInterceptor` implementation you are using?

```csharp
        public object Intercept(Func<object[], object> method, object[] parameters, string methodName)
        {
            // Runs the method and consitently serialises the result from the function

            var interceptTask = Task.Run(() => InterceptAsync(method, parameters, methodName));


            return interceptTask.Result;
        }
        private async Task<object> InterceptAsync(Func<object[], object> method, object[] parameters, string methodName)
        {
            // in here method is normally always async
            // this makes returnValue probably a Task
            var returnValue = method(parameters);

            // Do some checks to make sure its properly a Task or similar
            // pseudocodeish
            if (returnValue is Task) 
            {
                return Serialise(await returnValue);
            } 
            else 
            {
                return Serialise(returnValue);
            }

            // Otherwise await result and serialise
            
        }
```

The syncronous wait for the result here is to avoid some funky frontend behaviour where it completely freezes for several hundred milliseconds whilst processing these execution interception results....

amaitland:
> From any of the little snippets that I had before will cause all the other methods that need executing to wait for the `DoSomething()` function to return, whether async or not, whether returning a task or not.

The ConcurrentMethodRunnerQueue is not used by default, the `QUnit` tests are run sequentially.

For testing purposes you can launch the `CefSharp.Wpf.Example` an use the `Tests->Async JSB Task Tests` menu item to run the `Task` related tests.

There's some refactoring to do before this can be run as part of the build process, this is on my `TODO` list.

> I believe that the example does also use the concurrent method runner - it sets `CefSharpSettings.ConcurrentTaskExecution = true;`

Only for a the `Async JSB Task Tests` is it enabled.

If you can provide a clear list of steps/example to reproduce the behaviour you are seeing I'll happy look into it.

amaitland:
I've created `PR` #3296 as my proposed simplification or this `PR`, it aims to 

- Eliminate the breaking changes introduced by this `PR`.
- Add an example
- Add some additional tests.
- To have all method calls go through `InterceptAsync` I've added `JavascriptBindingSettings.AlwaysInterceptAsynchronously` to configure this behaviour. (If anyone has a better name for the property then I'm open to suggestions).

Commit https://github.com/cefsharp/CefSharp/pull/3291/commits/eea134c91f4c7501ac5a96fe4a50718d3e0050fd is the base of this `PR`.

If throwing an exception in `TryCallMethodAsync` when a `Task` is returned (as shown in https://github.com/cefsharp/CefSharp/pull/3291/files#diff-df112b02559a8f6e4974da2943e081757cde96d23e5e7ec39ffcf3ab70a10de3R52) is a desired behaviour then we can add a configuration option for this behaviour. It will be disabled by default to avoid inconveniencing upgrading users.

There is still some code duplication that needs to be addressed before merge.

I understand there are certain cases where customising the behaviour of the `MethodRunnerQueue` and `JavascriptObjectRepository` are desirable. I plan to make these pluggable with base classes to allow for further customisation of existing behaviour.

AndrewMD5:
Your solution seems like the better approach. Closing in favor of that.

amaitland:
Sorry for the delayed response, I've been away for the Christmas holidays. 

I'll finish #3296 shortly (will be included in the next `-pre` release which hopefully will be out next week).

I'd like to also personally thank you for the continued support, it means a lot to my family 😄

Issue #3197, opened 2020-07-23
Net Core/Net 5 - Implementation


amaitland:
Track the progress remaining on actually compiling the project with .Net Core

Basic support for compiling with .Net Core was added with #3181 with additional commit (https://github.com/cefsharp/CefSharp/commit/e8cf336bf3f2871c400cd01ee090fa42df97adf8) to get the core projects compiling/tests passing,

- [x] Integrate PDB into dll (bug in package reference makes it near impossible to distribute a pdb, see https://github.com/dotnet/sdk/issues/1458)
- [x] Integrate source code directly into the dll if the size doesn't increase dramatically see https://github.com/cefsharp/CefSharp.Wpf.HwndHost/commit/901c6a7095dd5fb1124174593302947224a40943#diff-5418bd97c668e509388f5d466d94a436R10 for an example
- [x] Generate [reference assemblies](https://docs.microsoft.com/en-us/dotnet/standard/assembly/reference-assemblies) for csproj
- [x] Generate reference assembly for CefSharp.Core.dll using `GenApi`
- [x] Improve `Reference Assembly` generation as relies on `x64` hardcoded in project file. See https://github.com/cefsharp/CefSharp/issues/3197#issuecomment-672719684
- [x] Check if we need to make the BrowserSubProcess TSAware (it's now a WinExe  rather than a console exe so not sure that we have to).
- [x] Add Net Core Unit Test Project
- [x] Update appveyor to build the matrix (Net Core requires a VS2019 build worker image)
- [x] Add assembly manifest for VC++ projects to prevent loading issues of `ijwhost.dll` see https://github.com/dotnet/runtime/issues/38231 and https://github.com/dotnet/runtime/issues/37972#issuecomment-647726082
- [x] Distribute `ijwhost.dll` as required to load `C++/CLI`
- [x] Dynamically set version number in `chromiumembeddedframework.runtime.json`
- [x] Finish Nuget packages (partially done, they don't work yet)
- [x] Fix type in `.targets` file as detailed in https://github.com/cefsharp/CefSharp/issues/3197#issuecomment-733534204
- [ ] Update Dependency Check (possibly create a `Net Core` variant).
- [x] Update `BrowserSubProcess` to support running on `.Net 5.0` when `.Net Core` is not installed.
- [x] Add `.Net Core` specific Nuget Readme.txt
- [ ] `VC++ 2019` is required for `.Net Core` (add to readme, document in other places).
[...]

PR #3295, opened 2020-11-26
.NET Core - Allow non-portable RuntimeIdentifiers

kpreisser:
**Issue:** cefsharp/CefSharp#3197

**Summary:**
   - Adjust the `RuntimeIdentifier` check in `CefSharp.Common.NETCore.targets` to make it work with non-portable RIDs like `win7-x64` or `win10-x86`.

**Changes:** 
- Changed the comparison of the `RuntimeIdentifier` to not check for exact RIDs like `win-x64`, but check that the string starts with `win` and that it contains the architecture like `-x64`.
- Note: I used `.Contains('-x64')` rather than `.EndsWith('-x64')` as there seem to be some RIDs that have [another suffix](https://github.com/dotnet/runtime/blob/f8e9156244606582c202c9f01be571baf523cccf/src/libraries/pkg/Microsoft.NETCore.Platforms/runtime.json#L2974), like `win7-x64-aot`, which is described as `[additional qualifiers]` in [.NET Core RID catalog](https://docs.microsoft.com/en-us/dotnet/core/rid-catalog).  However I don't know exactly when/how such RIDs would be used.
      
**How Has This Been Tested?**  

- Built the `CefSharp.Common.NETCore` NuGet package, referenced it in a app that uses `CefSharp.WinForms.NETCore`, and published it using runtime identifiers `win7-x64` and `win7-x86`.

**Types of changes**

- [ ] Updated documentation

**Checklist:**

- [ ] Commented my code

### DIFF
--- a/NuGet/PackageReference/CefSharp.Common.NETCore.targets
+++ b/NuGet/PackageReference/CefSharp.Common.NETCore.targets
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <Choose>
-    <When Condition="'$(RuntimeIdentifier)' == 'win-x64'">
+    <When Condition="$(RuntimeIdentifier.StartsWith('win')) and $(RuntimeIdentifier.Contains('-x64'))">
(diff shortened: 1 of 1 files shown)


amaitland:
Excellent, thanks 👍

Issue #3284, opened 2020-11-16
Feature Request - Add .Net 5.0 Support


Thawing75:

chris-kruining:
yes please

Haltroy:
It seems that .Net 5.0 gets its core from .Net Core, meaning that it's not simple to just update like in .Net Framework.
.Net Core implementations should be done first in order to support .Net 5.0.
Issue #3197 contains the progress on .Net Core implementation.
Here's the [.Net Portability Analyzer](https://docs.microsoft.com/en-us/dotnet/core/porting/) results:
[REPORT.zip](https://github.com/cefsharp/CefSharp/files/5604948/REPORT.zip)
I thought this might help but I'm not happy with the results. Felt like a waste of time.

julesx:
A big draw of .NET 5 is the single file executable. I am investigating if using Costura +  CefSharp is possible, but it's certainly not simple.

amaitland:
> It seems that .Net 5.0 gets its core from .Net Core, meaning that it's not simple to just update like in .Net Framework.
> .Net Core implementations should be done first in order to support .Net 5.0.

Firstly supporting `.Net 5.0` doesn't necessarily mean compiling a `.Net 5.0` specific version. As it is my understanding `.Net 5.0` is the version after `.Net Core 3.1` and is backwards compatible. https://github.com/cefsharp/CefSharp/issues/3197#issuecomment-732809648 suggests testing with `.Net 5.0` to this point has been fairly successful, just a few quirks with the packaging. 

At this point in time I believe the `.Net Core` implementation (https://github.com/cefsharp/CefSharp/issues/3197) will be usable by `.Net Core 3.1` and `.Net 5.0`. 

> Here's the [.Net Portability Analyzer](https://docs.microsoft.com/en-us/dotnet/core/porting/) results:
> [REPORT.zip](https://github.com/cefsharp/CefSharp/files/5604948/REPORT.zip)
> I thought this might help but I'm not happy with the results. Felt like a waste of time.

Thanks, not actually necessary. The `CI` builds have been successfully compiling a .`Net Core` version for a few months now. There's a `CefSharp3.netcore.sln` solution in the project source. The problem is actually creating a set of `Nuget` packages, progress been slow(I've spent countless hours on this already), there's very little in the way of official guidance from `MS` in regards to our specific use case.

> A big draw of .NET 5 is the single file executable. I am investigating if using Costura + CefSharp is possible, but it's certainly not simple.

[...]

PR #3359, opened 2021-01-16
Add support for Windows ARM64 to CefSharp.NETCore packages

### LINKED ISSUE #2944: Feature Request - ARM64 support

kpreisser:
**Fixes #2944**
Relates to #3284

**Summary:** 
   - Added support for runtime identifier `win-arm64` to the `CefSharp.NETCore` packages, so that they can be used to publish apps for `win-arm64` on .NET 5.0 and higher.

**Changes:**
   - Added platform `arm64` to the platform-specific .NET Core projects (`CefSharp.Core.Runtime.netcore`, `CefSharp.BrowserSubprocess.Core.netcore`, `CefSharp.BrowserSubprocess.netcore`, `CefSharp.Test.netcore`, `CefSharp.WinForms.Example.netcore`, `CefSharp.WPF.Example.netcore`, `CefSharp.OffScreen.Example.netcore`).
     - Note: For .NET Core projects that now specify `<Platforms>x86;x64;arm64</Platforms>`, it seems we explicitly need to set `<PlatformTarget>` to `arm64` when `Platform` is `arm64`. Without this, the `PlatformTarget` seems to be `AnyCPU` and e.g. the `CefSharp.BrowserSubprocess.exe` would be an `x64` EXE. See: https://github.com/dotnet/sdk/issues/15434
   - In `CefSharp.BrowseSubprocess.netcore`, added `net5.0` as target framework, so that we can use the `net5.0` EXE for `win-arm64` (as .NET 5.0 officially supports creating Windows ARM64 apps).
   - Updated `Initializer.ModuleInitializer` to use `RuntimeInformation.ProcessArchitecture` for building the platform-specific folder path.
   - Switched the `CefSharp.*.Example.netcore` and `CefSharp.Test.netcore` projects from `cef.redist.*` to `chromiumembeddedframework.runtime`.
   - Added the `win-arm64` artifacts to the NuGet packages.
   - Changed `build.netcore.ps1` to also build for platform `arm64`.
      
**How Has This Been Tested?**  
- Installed .NET SDK `6.0.100-preview.2.21155.3` (x64), switched `CefSharp.MinimalExample.[WinForms|Wpf|OffScreen].netcore` to `net6.0-windows`, switched to self-hosting the subprocess, referenced the built NuGet packages, and published the projects for `win-arm64`.
- Run the published apps on Windows 10 Pro (ARM64) Build 21337 on a Raspberry Pi 4 Model B (4 GB).
- Additionally, installed .NET SDK `6.0.100-preview.2.21155.3` (arm64) on the target machine and run the non-published version of `CefSharp.MinimalExample.Winforms` using the default `CefSharp.BrowserSubprocess.exe`.

## Screenshots:
You can see in `Architecture` column of the task manager that the application is running natively as `ARM64`.
(Windows 10 on ARM also supports running `x86` apps (and `x64` in insider builds) using emulation, but they would run significantly slower than native ARM64 apps.)

### CefSharp.MinimalExample.WinForms.netcore:

I changed the form title to output `RuntimeInformation.FrameworkDescription` and `RuntimeInformation.ProcessArchitecture`:

**Self-hosting subprocess**:
grafik

**Default subprocess**:
grafik

### CefSharp.MinimalExample.Wpf.netcore:

grafik

### CefSharp.MinimalExample.OffScreen.netcore:
grafik

**Types of changes**

- [ ] Updated documentation

**Checklist:**

- [ ] Commented my code


amaitland (changes requested):
> * Upgrade to `88.1.2+gac8e5f0+chromium-88.0.4324.27 / Chromium 88.0.4324.27`

Thanks for keeping this as it's own commit, I've cherry picked and have just pushed an update to `88.1.3`. Hopefully it should be easy enough to rebase this on `master`.

> * Note: For .NET Core projects that now specify `<Platforms>x86;x64;arm64</Platforms>`, it seems we explicitly need to set `<PlatformTarget>` to `arm64` when `Platform` is `arm64`. Without this, the `PlatformTarget` seems to be `AnyCPU` and e.g. the `CefSharp.BrowserSubprocess.exe` would be an `x64` EXE.

Seems like a bug, were you able to find an open issue?

> Additionally, the current `Windows ARM64` installer for .NET 5.0 doesn't include the Desktop Runtime (WinForms/WPF).

I was under the impression the `WinForms ARM64` implementation was complete for `.Net 5.0`, how does one install `WinForms` on `ARM64`?

I'd like to remove the dependency on `cef.redist.arm64` as it's only used in development. Switching the `Net Core` projects to using the `chromiumembeddedframework.runtime` package and specifying a `RuntimeIdentifier` should be feasible now as the packages are now on the verge of an official release.

I'll install the `ARM64` cross compiler and build this locally when I get a chance. 

Overall looks very promising, thanks for your efforts 👍

kpreisser:
Thanks a lot for your feedback!



> Hopefully it should be easy enough to rebase this on `master`.

I will look at rebasing it at current master when I have some time.



> Seems like a bug, were you able to find an open issue?

I didn't find one in dotnet/sdk. I created a new issue here: https://github.com/dotnet/sdk/issues/15434



> I'd like to remove the dependency on `cef.redist.arm64` as it's only used in development. Switching the `Net Core` projects to using the `chromiumembeddedframework.runtime` package and specifying a `RuntimeIdentifier` should be feasible now as the packages are now on the verge of an official release.

OK, I will look at switching the projects to the `chromiumembeddedframework` package.

Thank you!

CefSharp.BrowserSubprocess.Core/CefSharp.BrowserSubprocess.Core.netcore.vcxproj
```diff
@@ -138,6 +179,25 @@
       <AdditionalManifestFiles>Assembly.manifest</AdditionalManifestFiles>
     </Manifest>
   </ItemDefinitionGroup>
+  <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Debug|arm64'">
+    <ClCompile>
+      <WarningLevel>Level3</WarningLevel>
+      <AdditionalIncludeDirectories>$(SolutionDir)packages\$(CefSdkVer)\CEF;%(AdditionalIncludeDirectories);$(ProjectDir)</AdditionalIncludeDirectories>
+      <Optimization>Disabled</Optimization>
+      <PreprocessorDefinitions>WIN32;_DEBUG;EXPORT;NETCOREAPP;%(PreprocessorDefinitions)</PreprocessorDefinitions>
```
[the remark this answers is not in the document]
kpreisser: It seems `WIN32` is always be defined implicitely by `minwindef.h`, regardless of the architecture. Maybe we can remove it completely from the project file. See also: https://github.com/MicrosoftDocs/cpp-docs/issues/1459

amaitland:
> OK, I will look at switching the projects to the chromiumembeddedframework package

I can do that tomorrow.

kpreisser:
> I was under the impression the `WinForms ARM64` implementation was complete for `.Net 5.0`, how does one install `WinForms` on `ARM64`?

The current .NET 5.0 installers for Windows ARM64 don't include the Windows Desktop components (WinForms+WPF) because WPF isn't yet ready for ARM64. See the [5.0-preview-8 blog post](https://devblogs.microsoft.com/dotnet/announcing-net-5-0-preview-8/):

> We added support for Windows ARM64 as part of this release. We’ve made the relatively late decision to delay the Windows Desktop component (Windows Forms, WPF). Windows Forms is near ready, but WPF is not, and we don’t want to release only half the Windows Desktop component, in part because we don’t test it in a split configuration. We hope to add the Windows Desktop component back as part of a 5.0 servicing update.

Sorry, I previously thought that publishing a WinForms app using RID `win-arm64` already works with .NET 5.0.2, but apparently I had used a daily build of .NET 6.0 SDK. It doesn't yet seem work with the .NET 5.0 SDK.

CefSharp.BrowserSubprocess.Core/CefSharp.BrowserSubprocess.Core.netcore.vcxproj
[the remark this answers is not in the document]
kpreisser: I removed the definition of `WIN32` in the `netcore` projects.

kpreisser:
I rebased the PR on `master` branch. (Note: File `CefSharp.Common.NETCore.targets` still referenced version `87.1.13` of `chromiumembeddedframework.runtime`, I updated it also to `88.1.3`.)

 I have also updated the `CefSharp.*.Example.netcore` and `CefSharp.Test.netcore` projects to use `chromiumembeddedframework.runtime` instead of `cef.redist` and to specify a `RuntimeIdentifier`, so I think the `cef.redist.arm64` package is now no longer needed.

However, I had to exclude these projects for build in the solution file for platform `arm64`, as otherwise it fails because WinForms and WPF are not yet available for ARM64 in .NET 5.0 SDK.
Once a .NET 5.0 release with full Windows ARM64 support is available, we can include these projects again (and add `win-arm64` to `RuntimeIdentifiers`).

Thank you!

amaitland:
> The current .NET 5.0 installers for Windows ARM64 don't include the Windows Desktop components (WinForms+WPF) because WPF isn't yet ready for ARM64. See the [5.0-preview-8 blog post](https://devblogs.microsoft.com/dotnet/announcing-net-5-0-preview-8/):

Thanks for the clarification. Makes sense they want to ship them together.

> I rebased the PR on `master` branch.

Great, thanks 👍 

> (Note: File `CefSharp.Common.NETCore.targets` still referenced version `87.1.13` of `chromiumembeddedframework.runtime`, I updated it also to `88.1.3`.)

The build script should take care of updating this. https://github.com/cefsharp/CefSharp/blob/master/build.netcore.ps1#L293

> I have also updated the `CefSharp.*.Example.netcore` and `CefSharp.Test.netcore` projects to use `chromiumembeddedframework.runtime` instead of `cef.redist` and to specify a `RuntimeIdentifier`, so I think the `cef.redist.arm64` package is now no longer needed.

Excellent, thank you.

> However, I had to exclude these projects for build in the solution file for platform `arm64`, as otherwise it fails because WinForms and WPF are not yet available for ARM64 in .NET 5.0 SDK (and because they currently target `netcoreapp3.1` only).
> Once a .NET 5.0 release with full Windows ARM64 support is available, we can increase the `TargetFramework` to `net5.0-windows` and can include these projects again (and add `win-arm64` to `RuntimeIdentifiers`).

As it's not clear when the final release will happen I might try and merge this and comment out the minimum to keep things ticking along. 

We should also look at updating our `.vsconfig` file. Adding [Microsoft.VisualStudio.Component.VC.Tools.ARM64](https://github.com/dotnet/runtime/blob/master/.vsconfig#L45) should hopefully do the trick.

CefSharp.OffScreen.Example/CefSharp.OffScreen.Example.netcore.csproj
```diff
@@ -17,22 +17,28 @@
...
+        <Platforms>x86;x64;arm64</Platforms>
+        <!-- Explicitly set the PlatformTarget for arm64, see: https://github.com/dotnet/sdk/issues/15434 -->
+        <PlatformTarget Condition="'$(Platform)' == 'arm64'">arm64</PlatformTarget>
         <StartupObject>CefSharp.OffScreen.Example.Program</StartupObject>
         <!--
         Run on a newer version of .Net if the specified version (.Net Core 3.1) is not installed
         https://github.com/dotnet/docs/issues/12237
         -->
         <RollForward>Major</RollForward>
+        <RuntimeIdentifiers>win-x86;win-x64</RuntimeIdentifiers>
```
amaitland, line 29: Is the `<RuntimeIdentifiers/>` (plural) element required here? I was under the impression it was one or other.

CefSharp.Example/CefExample.cs
```diff
@@ -168,10 +168,11 @@ public static void Init(CefSettingsBase settings, IBrowserProcessHandler browser
 
             if (DebuggingSubProcess)
             {
-                var architecture = Environment.Is64BitProcess ? "x64" : "x86";
 #if NETCOREAPP
-                settings.BrowserSubprocessPath = Path.GetFullPath("..\\..\\..\\..\\..\\CefSharp.BrowserSubprocess\\bin.netcore\\" + architecture + "\\Debug\\netcoreapp3.1\\CefSharp.BrowserSubprocess.exe");
+                var architecture = System.Runtime.InteropServices.RuntimeInformation.ProcessArchitecture.ToString().ToLowerInvariant();
```
amaitland, line 172: I've updated the example and test projects to `.Net 4.7.2` in commit https://github.com/cefsharp/CefSharp/commit/0cdbef1d44c953beacd78384f51e1aa079710936

If I'm reading the API doc correctly `System.Runtime.InteropServices.RuntimeInformation.ProcessArchitecture` was added in `.Net 4.7.1` so we should be able to use this in all example/test projects now without needing a `if else`

amaitland:
The `CefSharp.AfterBuild.targets` likely needs to be updated now the switch to using the `chromiumembeddedframework.runtime` packages.

amaitland:
After installing the `ARM64` build tools I was able to successfully build using `Visual Studio` 👍 (I have no means of actually testing the resulting binaries)

CefSharp.OffScreen.Example/CefSharp.OffScreen.Example.netcore.csproj
kpreisser: It seems to be required here for `build.netcore.ps1` to work (using `MSBuild` from VS, not `dotnet build`). Otherwise, MSBuild complains about the runtime identifier used (e.g. `win-x64`) not being present in `obj\project.assets.json`, and suggests to add it to `<RuntimeIdentifiers>`.

I think it is for the `restore` operation which needs to know what runtime packages to restore before building the project (though I'm not sure why it works without specifying `<RuntimeIdentifiers>` when building directly in Visual Studio).

CefSharp.Example/CefExample.cs
kpreisser: Thank you! Yes, `RuntimeInformation` was introduced in .NET Standard 2.0 and is therefore available in .NET Framework 4.7.1. I will update the code to remove the `#if`.

kpreisser: Done.

kpreisser:
I updated the `.vsconfig` to include the VC ARM64 tools, and updated `CefSharp.AfterBuild.targets` to remove the targets for copying the `cef.redist` files, which should no longer be needed after switching to `chromiumembeddedframework.runtime`.

Thank you!

kpreisser:
> ❌ [Build CefSharp 88.1.6-CI3887 failed](https://ci.appveyor.com/project/cefsharp/cefsharp/builds/37511914) (commit [a8b4cd8](https://github.com/cefsharp/CefSharp/commit/a8b4cd803a) by @kpreisser)

I think the build failure is unrelated to the changes in this PR.

kpreisser:
Publishing WPF apps for `win-arm64` now seems to work with .NET SDK `6.0.100-preview.1.21078.9` (daily build). I updated the first post to include a screenshot of `CefSharp.MinimalExample.Wpf.netcore`.

kpreisser:
I checked that the `CefSharp.BrowserSubprocess.exe` also works when building `CefSharp.MinimalExample.WinForms.netcore` for `win-arm64` (see screenshot in first post). It seems the `CefSharp.BrowserSubprocess.exe` now also works when publishing the application as self-contained (thanks to 7c45907c76b7936f623a761a7a920dad56ae79ce) :+1:

@amaitland Is there anything still left to be done before this can be merged? Thanks!

CefSharp.Wpf.Example/CefSharp.Wpf.Example.netcore.csproj
```diff
@@ -18,22 +18,28 @@
...
+        <Platforms>x86;x64;arm64</Platforms>
+        <!-- Explicitly set the PlatformTarget for arm64, see: https://github.com/dotnet/sdk/issues/15434 -->
+        <PlatformTarget Condition="'$(Platform)' == 'arm64'">arm64</PlatformTarget>
         <StartupObject>CefSharp.Wpf.Example.Program</StartupObject>
         <!--
         Run on a newer version of .Net if the specified version (.Net Core 3.1) is not installed
         https://github.com/dotnet/docs/issues/12237
         -->
         <RollForward>Major</RollForward>
+        <RuntimeIdentifiers>win-x86;win-x64</RuntimeIdentifiers>
```
amaitland, line 30: Do we need `RuntimeIdentifiers` here?

amaitland:
Thanks for the hard work on this 👍 I was hoping to find a way to runup a virtual machine to test this, wasn't clear that was possible currently on an `Intel CPU`.

I'll merge into `master` now, might be a few tweaks that I'll make later, probably switch the examples to always being SelfContained = false.

kpreisser:
Great, thanks a lot! :+1:

FYI, there is an unofficial/unsupported way to install a Windows 10 ARM64 build on a Raspberry Pi 3/4 (that has an ARM64 processor), which is what I was using to test the builds. You can search for "Windows on Raspberry" to find the tool that can be used for deploying the Windows image.

As for VMs, it seems be possible to [use QEMU](https://virtuallyfun.com/wordpress/2018/03/22/windows-10-arm64-on-qemu/) on Windows to emulate an ARM64 processor on x64 hardware, but I haven't tried that yet.

Issue #3390, opened 2021-02-03
Every project with transitive reference to CefSharp now needs CefSharpAnyCpuSupport property set.


deonberlin:
- **What version of the product are you using?**
    Using Nuget version 87.1.132

- **What architecture x86 or x64?**
    AnyCPU
    
- **On what operating system?**
    Win10

- **Are you using `WinForms`, `WPF` or `OffScreen`?**
    WPF/OffScreen
    
- **What steps will reproduce the problem?**
I have the following project structure (all projects using the new SDK project style):
_Project UI_ ---> references _Project Browser_ --> references CefSharp packages (OffScreen and WPF)
Both projects have CefSharpAnyCpuSupport set to true
_Project UI Extensions_ ---> references _Project UI_ ---> transitive reference to CefSharp
Project has not set CefSharpAnyCpuSupport

When building, the TransformXML task in CefSharpCommonAnyCPUConfigTransform target raises errors about missing source parameter for all projects which have not set CefSharpAnyCpuSupport and target AnyCPU with a transitive reference to CepSharp.

- **What is the expected output? What do you see instead?**
I would have expected that the build raises no errors, like it was in versions before 87.

Currently my workaround is to set the CefSharpAnyCpuSupport property in Directory.Build.props for the whole solution.

amaitland:
PackageReference assets are controllable on a per package basis. See https://docs.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files#controlling-dependency-assets

Specifically via setting privateAssets in this instance. 

> I would have expected that the build raises no errors, like it was in versions before 87.

If I remove the targets file from the buildTransative folder then there's no way to have it as transitive. If I leave it as is then transitive references should work properly and the user can configure the behaviour.

I'm not aware of a way to tell if the targets file is being used as a transitive reference and adjust the check,  I did look. If you can find a way a PR would be most welcomed. The error message can be improved to include details of how to control assets.

amaitland:
Additionally if you create a CefSharpBuildAction property in your project file and set the value to None (case  sensitive) and tweaking the assets you should hopefully get a similar behaviour.

deonberlin:
Okay, the actual problem was, that the libraries with transitive references did not had an app.config file, so the Source for the target was empty. It could be resolved by adding an additional condition to the build target which checks if ther was an App config file copied by checking '@(AppConfigWithTargetPath).

[...]

Issue #3389, opened 2021-02-03
Release Notification - 88.2.x


amaitland:
This version includes is a minor bug fix release with an updated version of `CEF/Chromium`. 

---

#### Release 88.2.x-pre
  - [x] Update to latest CEF build
  - [x] Create release notes
  - [x] Create `cefsharp/88` release branch
  - [x] Release `-pre` packages to nuget.org

#### Release 88.2.x

  - [x] Resolve remaining issues attached to [88 Milestone](https://github.com/cefsharp/CefSharp/milestone/53)
  - [x] Update `DevTools` client `API`
  - [x] Update https://github.com/cefsharp/CefSharp/blob/cefsharp/88/NuGet/Readme.txt#L30
  - [x] Create release notes
  - [x] Update General Usage API doc links
  - [x] Update Issue Template CEF Sample application links
  - [x] Generate new `API Doc` and upload to http://cefsharp.github.io/api/ 
  - [x] Release 88.2.x packages to nuget.org

amaitland:
- `PDF Viewer` toolbar buttons displaying their `$i18n` values should be fixed in this release see https://bitbucket.org/chromiumembedded/cef/issues/3081/chromium-85-pdf-viewer-cannot-handle-i18n

amaitland:
The `88.2.40-pre` release packages are now on Nuget.org. 

For the `.Net 4.5.2` and above
- https://www.nuget.org/packages/CefSharp.WinForms/88.2.40-pre
- https://www.nuget.org/packages/CefSharp.Wpf/88.2.40-pre
- https://www.nuget.org/packages/CefSharp.OffScreen/88.2.40-pre

These packages required **Visual C++ 2015** or greater

For `.Net Core 3.1/.Net 5.0`
- https://www.nuget.org/packages/CefSharp.WinForms.NETCore/88.2.40-pre
- https://www.nuget.org/packages/CefSharp.Wpf.NETCore/88.2.40-pre
- https://www.nuget.org/packages/CefSharp.OffScreen.NETCore/88.2.40-pre

These packages required **Visual C++ 2019**

**NOTE** A minimum of `.Net Core 3.1` is required (for `.Net 3.0` which is no longer supported by `Microsoft` you'll need to use the older packages).

---

- The [CefSharp.MinimalExample/tree/cefsharp/88](https://github.com/cefsharp/CefSharp.MinimalExample/tree/cefsharp/88) branch has been created for testing purposes.
[...]

PR #3391, opened 2021-02-04
Resolve build issues for projects without app.config

deonberlin:
**Fixes:** #3390

**Summary:**
   - Added condition to build target to check if app.config exists

   - Build target condition for @(AppConfigWithTargetPath) added
      
**How Has This Been Tested?**  

  - build package
  - referenced newly crated package in a library project without app.config
  - build library project successfully

**Types of changes**

- [ ] Updated documentation

**Checklist:**

- [ ] Commented my code

### DIFF
--- a/NuGet/CefSharp.Common.targets
+++ b/NuGet/CefSharp.Common.targets
@@ -283,7 +283,7 @@
   --> 
   <UsingTask TaskName="TransformXml" AssemblyFile="$(CefSharpTransformXmlDllPath)" Condition="Exists('$(CefSharpTransformXmlDllPath)') AND '$(PlatformTarget)' == 'AnyCPU' AND '$(CefSharpAnyCpuSupport)' == ''" />
   
-  <Target Name="CefSharpCommonAnyCPUConfigTransform" AfterTargets="_CopyAppConfigFile" Condition="Exists('$(CefSharpTransformXmlDllPath)') AND '$(PlatformTarget)' == 'AnyCPU' AND '$(CefSharpAnyCpuSupport)' == ''">
+  <Target Name="CefSharpCommonAnyCPUConfigTransform" AfterTargets="_CopyAppConfigFile" Condition="'@(AppConfigWithTargetPath)' != '' AND Exists('$(CefSharpTransformXmlDllPath)') AND '$(PlatformTarget)' == 'AnyCPU' AND '$(CefSharpAnyCpuSupport)' == ''">
     <TransformXml Source="@(AppConfigWithTargetPath->'$(OutDir)%(TargetPath)')" Transform="$(MSBuildThisFileDirectory)..\build\app.config.x86.transform" Destination="@(AppConfigWithTargetPath->'$(OutDir)%(TargetPath)')"/>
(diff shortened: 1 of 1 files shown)


amaitland:
Thanks 👍 This is only part of it, we need to add the equals empty check to the CefSharpPlatformCheck target as well.

Users with a PlatformTarget of AnyCPU and without an app.config need to be informed extra steps are required to make AnyCPU work.

amaitland:
Testing this change and it fails with 

```
error MSB4099: A reference to an item list at position 120 is not allowed in this condition "Exists('$(CefSharpTransformXmlDllPath)') AND '$(PlatformTarget)' == 'AnyCPU' AND '$(CefSharpAnyCpuSupport)' == '' AND '@(AppConfigWithTargetPath)' != ''". [C:\projects\CefSharp.MinimalExample\CefSharp.MinimalExample.WinForms\CefSharp.MinimalExample.WinForms.net472.csproj
```

deonberlin:
Interesting. How did you test that, because the condition in the error message doesn't match my code changes, it matches my comment.

I've tested by simply modifying the buildTransitive\CefSharp.Common.targets file in the nuget package cache, and for me it works.

The reference to the list condition can of course only evaluated after the target _CopyAppConfigFile has run.

amaitland:
I applied the changes manually then generated a new set of Nuget packages, it's when I perform a Nuget upgrade it fails. I've also tried your exact commit (Please target PRs again master in future)

> I've tested by simply modifying the buildTransitive\CefSharp.Common.targets file in the nuget package cache, and for me it works.

Did you close and reopen your solution to make sure visual studio reflected the changes? Targets files are cached. 

> The reference to the list condition can of course only evaluated after the target _CopyAppConfigFile has run.

The condition should still evaluate without exception regardless of when it's run as it's a simple equals and not equals.

deonberlin:
> Please target PRs again master in future

I had the hope for an update on the 87 package, that's why I targeted this branch.

> Did you close and reopen your solution to make sure visual studio reflected the changes? Targets files are cached.

Yes I did, and actually each change in the target file was applied in the next build, at least here on my machine.

> The condition should still evaluate without exception regardless of when it's run as it's a simple equals and not equals.

I had thought so too. Actually I've copied that condition from the definition of _CopyAppConfigFile, but the error messages states it's invalid, even if the same condition must have been evaluated correctly, because _CopyAppConfigFile was executed before.

I'll try to reproduce the error.

amaitland:
I've squashed your changes and merged into master in commit https://github.com/cefsharp/CefSharp/commit/d36c0bceb3335d76b355ef7cf3c74dafe4a80805

Build 88.2.10-CI3900 should be available from https://www.myget.org/gallery/cefsharp for testing purposes. I'll run another Nuget upgrade tomorrow,  make sure I didn't make a mistake somewhere. Your welcome to test out the packages.

deonberlin:
The packages 88.2.10-CI3900 are working fine on my side with all platforms and for AnyCPU with and without CefSharpAnyCpuSupport.

amaitland:
This has been merged into the `cefsharp/88` branch and is included in the `88.2.40-pre` release, see https://github.com/cefsharp/CefSharp/issues/3389#issuecomment-776506045 for details.

PR #3426, opened 2021-03-07
Fix typos in CefSharp.Common.targets

jozefizso:


NuGet/CefSharp.Common.targets
```diff
@@ -89,15 +89,15 @@
...
+	For class libraries that are often AnyCPU and consuming Exe/WinExe projects that are x64 or x86 the resulting build output
 	is polluted with extra copies in sub folders.
     -->
     <CefSharpBuildAction Condition="'$(CefSharpBuildAction)' == '' AND '$(NuGetProjectStyle)' != 'PackageReference'">None</CefSharpBuildAction>
     <CefSharpBuildAction Condition="'$(CefSharpBuildAction)' == '' AND '$(NuGetProjectStyle)' == 'PackageReference'">Content</CefSharpBuildAction>
 	<CefSharpBuildAction Condition="'$(CefSharpBuildAction)' == '' AND '$(NuGetProjectStyle)' == 'PackageReference' AND $(NuGetToolVersion) > '5.0' AND '$(OutputType)' == 'Library'">NoAction</CefSharpBuildAction>
 	
     <CefSharpTargetDir Condition="'$(CefSharpTargetDir)' == ''"></CefSharpTargetDir>
-    <!-- Ideally we could use EnsureTrailingSlash, was only added on 2016 so unlikely supported in older version of VS -->
+    <!-- Ideally we could use EnsureTrailingSlash, was only added on 2017 so unlikely supported in older version of VS -->
```
amaitland, line 100: 2016 is a reference to the year EnsureTrailingSlash was added to msbuild as the commit didn't have details of the milestone or any tags associated. Not actually a reference to VS2017 as msbuild is independent,  not that it really matters.

amaitland:
Thanks 👍

NuGet/CefSharp.Common.targets
jozefizso: I see, thanks for explanation. When I googled it, I did not notice it was in fact implemented in msbuild in the year 2016: https://github.com/dotnet/msbuild/issues/1302

PR #3432, opened 2021-03-09
Fix the copyright symbol in native libraries

jozefizso:
A small fix for `.rc` files so the © symbol is compiled correctly.

Old behavior: LegalCopyright is compiled as `Copyright Â© 2021 The CefSharp Authors`
New behavior: LegalCopyright will have correct value `Copyright © 2021 The CefSharp Authors`

image


amaitland:
Great, thanks 👍

PR #3447, opened 2021-03-18
Additional notes for "renderer-process-limit" to avoid freezed DevTools.

cbries:
I spend several days on analyzing the issue of a freezed DevTool when a Breakpoint hit, i.e. DevTools became unresponsible and were not usable anymore. To help other I have extended the documentation for the specific setting "renderer-process-limit".


amaitland:
I'd rather remove the line from the example, there was a time when limiting the number of render processes was a viable workaround for a bug in Chromium,  that was like version 57. 

 Setting renderer-process-limit is something I'd strongly advise against with recent versions (any version that supports chromium site isolation). It's not an option that's tested or supported from a CEF point of view. 

Adding a generic disclaimer about behaviour changing between versions for any command line is possibly worthwhile.

PR #3467, opened 2021-04-03
Javascript Binding - Allow binding to classes in the global namespace

pedoc:
**Fixes:** 

**Summary:** Sometimes, the registered Javascript object does not have a namespace, at this time this function will throw a ```NullReferenceException```.

**Changes:** 
      - Add null reference check for IsComplexType function 
      
**How Has This Been Tested?**  
This problem is obvious and does not need to be tested 

**Types of changes**

- [ ] Updated documentation

**Checklist:**

- [ ] Commented my code

### DIFF
--- a/CefSharp/Internals/JavascriptObjectRepository.cs
+++ b/CefSharp/Internals/JavascriptObjectRepository.cs
@@ -738,7 +738,7 @@ namespace CefSharp.Internals
                 baseType = Nullable.GetUnderlyingType(type);
             }
 
-            if (baseType == null || baseType.IsArray || (baseType.Namespace != null && baseType.Namespace.StartsWith("System")))
+            if (baseType == null || baseType.IsArray || baseType.Namespace?.StartsWith("System") == true)
             {
                 return false;
             }
--- a/CefSharp.Test/CefSharp.Test.csproj
+++ b/CefSharp.Test/CefSharp.Test.csproj
@@ -163,6 +163,7 @@
     <Compile Include="Framework\RequestContextFacts.cs" />
     <Compile Include="Framework\TestMemberInfo.cs" />
     <Compile Include="JavascriptBinding\IntegrationTestFacts.cs" />
+    <Compile Include="JavascriptBinding\JavaScriptObjectRepositoryFacts.cs" />
     <Compile Include="OffScreen\OffScreenBrowserBasicFacts.cs" />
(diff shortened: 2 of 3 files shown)


amaitland:
Please add a unit test.

amaitland:
A new call is the https://github.com/cefsharp/CefSharp/tree/master/CefSharp.Test/JavascriptBinding folder please,  JavaScriptObjectRepositoryFacts or something similar as a name please.

CefSharp/Internals/JavascriptObjectRepository.cs
```diff
@@ -738,7 +738,7 @@ private static bool IsComplexType(Type type)
                 baseType = Nullable.GetUnderlyingType(type);
             }
 
-            if (baseType == null || baseType.IsArray || baseType.Namespace.StartsWith("System"))
+            if (baseType == null || baseType.IsArray || (baseType.Namespace != null && baseType.Namespace.StartsWith("System")))
```
amaitland, line 741: Should be able to simplify this with a null conditional operator.

https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/member-access-operators#null-conditional-operators--and-

amaitland (changes requested):

CefSharp/Internals/JavascriptObjectRepository.cs
pedoc: For the time being, this will not work. Because the result of the null conditional operator is Nullable<bool>, it cannot satisfy the bool type constraint of if

pedoc:
@amaitland Hi,unit test has been added

CefSharp/Internals/JavascriptObjectRepository.cs
campersau:
```suggestion
            if (baseType == null || baseType.IsArray || baseType.Namespace?.StartsWith("System") == true)
```

amaitland: I believe `(baseType?.Namespace.StartsWith("System") ?? false)` would also have worked though the `== true` syntax is nicer so going with that. @campersau Thanks 👍

amaitland:
Updated Test case to run under .Net Core and added license disclaimer in commit https://github.com/cefsharp/CefSharp/commit/138cbca996a615eb7a0da971ce1f198ec6967c3e

PR #3627, opened 2021-06-17
WPF - Add support for Shift+AltGr

ptrsky:
The handling of AltGr characters in CefBrowsterHostWrapper did not cover capital diacritic letters. This commit adds handling for the Shift + AltGr + modifier combination.

**Fixes: #3626**

**Summary:**
   - I have added support for Shift+AltGr character input in CefSharp WPF

**Changes:** 
   - I have modified the current handling of AltGr characters in the CefBrowsterHostWrapper class by extending support to Shift+Ctrl+Alt combination
      
**How Has This Been Tested?**  
   - ran CefSharp.WPF.Example
   - navigate to www.google.com
   - enter capital polish letters (Shift + AltGr + EOASLZXCN) 
   - enter small polish letters (AltGr + eoaslzxcn), 
   - enter special characters (Shift + 1-9)

**Screenshots (if appropriate):**
obraz

**Types of changes**

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Updated documentation

**Checklist:**

- [x] Tested the code(if applicable)
- [x] Commented my code
- [ ] Changed the documentation(if applicable)
- [ ] New files have a license disclaimer
- [x] The formatting is consistent with the project (project supports .editorconfig)

PR #3769, opened 2021-08-22
WPF - Support ALT(Left) and ALT GR (Right) Key.

spinalcord:
**Fixes:** 

**Summary:**
- Support ALT(Left) and ALT GR (Right) Key.
- This allows: CTRL+ ALT +[SPECIFIC KEY] to input symbols with WPF (e.g. German keyboard layout)

**Changes:**
- Extended a if condition ([here](https://github.com/cefsharp/CefSharp/blob/b9de94303eb87141fbad7e28ad1399ea2d73ac67/CefSharp.Core.Runtime/Internals/CefBrowserHostWrapper.cpp#L382)) to proof whether the ALT key is pressed and not only ALT GR.
      
**How Has This Been Tested?**  
- Tested with a vanilla project and also with the Cef example browser, which is inside the master src.
- Entered symbols ² ³ { [ ] } } \ @ on certain search engines within Cef example browser

(Layout: German)
new

**Types of changes**

- [ ] Updated documentation

**Checklist:**

- [ ] Commented my code

### DIFF
--- a/CefSharp.Core.Runtime/Internals/CefBrowserHostWrapper.cpp
+++ b/CefSharp.Core.Runtime/Internals/CefBrowserHostWrapper.cpp
@@ -379,7 +379,7 @@ void CefBrowserHostWrapper::SendKeyEvent(int message, int wParam, int lParam)
 
         // mimic alt-gr check behaviour from
         // src/ui/events/win/events_win_utils.cc: GetModifiersFromKeyState
-        if (IsKeyDown(VK_RMENU))
+        if (IsKeyDown(VK_RMENU) || IsKeyDown(VK_LMENU))
         {
             // reverse AltGr detection taken from PlatformKeyMap::UsesAltGraph
             // instead of checking all combination for ctrl-alt, just check current char


amaitland:
Thanks, in future please target pull requests against master branch. Also please update the pull request title, `Update CefBrowserHostWrapper.cpp` doesn't describe the change. 

Do we need a left and right alt check or can we switch to VK_MENU? 

@campersau Thoughts/comments? I know you recently made a refinement to alt gr support.

campersau:
:one: `AltGr+[Key]` was working before.

:two: `AltGr+Shift+[Key]` was fixed by https://github.com/cefsharp/CefSharp/pull/3627 (I just ported it back to cef)

:three: `(L|R)Strg+Alt+[Key]` and `(L|R)Strg+Alt+Shift+[Key]` will be fixed by this PR, which I have confirmed locally as well.

So all of these combinations are now working in WPF when using a german keyboard layout:

| Key | (L\|R)Strg+Alt+[Key] | (L\|R)Strg+Alt+Shift+[Key] | AltGr+[Key] | AltGr+Shift+[Key] |
| --- | --------------- | --------------------- | ----------- | ----------------- |
| `q`   | `@`          :three:     |                       | `@`           :one:|                   |
| `e`   | `€`            :three:   |                       | `€`           :one:|                   |
| `m`   | `µ`          :three:     |                       | `µ`           :one:|                   |
| `2`   | `²`           :three:    |                       |  `²`           :one:|                   |
| `3`   | `³`             :three:  |                       | `³`           :one:|                   |
| `7`   | `{`            :three:   |                       | `{`           :one:|                   |
| `8`   | `[`           :three:    |                       | `[`           :one:|                   |
| `9`   | `]`          :three:     |                       | `]`           :one:|                   |
| `0`   | `}`           :three:    |                       | `}`           :one:|                   |
| `+`   | `~`              :three: |                      | `~`           :one:|                   |
| `<`   | `\|`              :three: |                       | `\|`           :one:|                   |
| `ß`   | `\`               :three:| `ẞ`                     :three:| `\`           :one:| `ẞ`                 :two:|

Instead of checking for both `VK_LMENU` and `VK_RMENU` checking only for `VK_MENU` also worked in my tests.

This PR could then also be backported:  https://bitbucket.org/chromiumembedded/cef/src/cfdec92624a82fa59119e0aeb2d4744978f0c9c8/tests/cefclient/browser/osr_window_win.cc#lines-796

@spinalcord FYI the screencapture shows `WinForms` browser instead of `WPF`.

spinalcord:
I updated the screenshot.

amaitland:
> Instead of checking for both `VK_LMENU` and `VK_RMENU` checking only for `VK_MENU` also worked in my tests.

@campersau Thanks, sounds like we can simply check for `VK_MENU`

> This PR could then also be backported: https://bitbucket.org/chromiumembedded/cef/src/cfdec92624a82fa59119e0aeb2d4744978f0c9c8/tests/cefclient/browser/osr_window_win.cc#lines-796

Can you clarify what you are referring to? The `CefSharp` code is structured slightly differently and the `IsKeyDown` check only happens for `KEYEVENT_CHAR` currently.

campersau:
@amaitland I think they are logical equal because of the lines above this. The main difference is that our call to set the `event.modifiers` comes earlier and thus we can put all the special handling inside the same `else` (`keyEvent.type = KEYEVENT_CHAR`) path whereas cef needs another check for `event.type == KEYEVENT_CHAR` because the `event.modifiers` are set later.
https://bitbucket.org/chromiumembedded/cef/src/cfdec92624a82fa59119e0aeb2d4744978f0c9c8/tests/cefclient/browser/osr_window_win.cc#lines-786:796

image

amaitland:
Merged into `master` in commit https://github.com/cefsharp/CefSharp/commit/b86aa00c4106e859c4d2502c96df706703f7750c

Switched to using `VK_MENU` in commit https://github.com/cefsharp/CefSharp/commit/e00b1ce0bf2cb79cc0019fe1c41abf9eae67ebe6

Assigning this to the `93` milestone.

Looking at the `Chromium` source and I think it might be worth looking at implementing a simplified version of [PlatformKeyMap::UsesAltGraph](https://source.chromium.org/chromium/chromium/src/+/main:ui/events/win/events_win_utils.cc;l=443?q=GetModifiersFromKeyState&ss=chromium) so the code path is only run for supported keyboard layouts.

PR #3801, opened 2021-09-11
Add Direct download to Cefsharp required files

Rolexboy88cool:
instead of downloading the source, and building the project, there's a download link where the owners update and we can use that download link to auto download cefsharp.winforms,cefsharp.wpf, etc files faster and easier.


amaitland:
If you have something to discuss then use https://github.com/cefsharp/CefSharp/discussions

PR #3811, opened 2021-09-14
Fix for arrays with empty elements

### LINKED ISSUE #3808: EvaluateScriptAsync crash when serializing an partially empty array
We managed to crash the embedded browser by simply evaluating code that requires returning an partially empty array ( [1,2, ,,,,, 5] ).

Here are the details:

We are using the latest version, 93.1.111

X64
	
- **What version of .Net?**
.Net Framework 4.8
    
Win10

WinForms
    
Take the provided code into a new Winforms 4.8 project and replace the Form1.cs code with that one. Add CefSharp via NuGet (Version 93.1.111 at that point of time). The error is easy to reproduce, just click the button on top of the page. In our real-live example, its a bit more random, most likely due to randomly reusing of "legal" memory.

- Code:
`
using System;
using System.Windows.Forms;
using CefSharp;
using CefSharp.WinForms;

namespace CefPureRepro
{
	public partial class Form1 : Form
	{
		private ChromiumWebBrowser _browser;

		public Form1()
		{
			InitializeComponent();

			Cef.Initialize(new CefSettings());

			Button button = new Button();
			button.Text = "Click me";
			button.Dock = DockStyle.Top;
			button.Click += ButtonClick;
			Controls.Add(button);

			_browser = new ChromiumWebBrowser("https://www.google.com");
			_browser.Dock = DockStyle.Fill;
			Controls.Add(_browser);
		}

		private async void ButtonClick(object sender, EventArgs e)
		{
			var result = await _browser.EvaluateScriptAsync("[1,2,,5]");
			Console.WriteLine(result.Result);
		}
	}
}
`

Expected is that the code returns the Array as an dynamic object. Unfortunately, instead, the embedded browser crashes.

I'd like to provide a better stack-trace but I can't download the pdbs as the URL contains the word "spotify" which is blocked by our IT.
That's what I figured out from VS:
libcef.dll!122a0175()	Unknown
libcef.dll!	Unknown
libcef.dll!122a1996()	Unknown
libcef.dll!0fe96bc6()	Unknown
CefSharp.BrowserSubprocess.Core.dll!CefV8ValueCToCpp::GetValue()	Unknown
[Managed to Native Transition]	
CefSharp.BrowserSubprocess.Core.dll!CefSharp.BrowserSubprocess.Serialization.SerializeV8Object<>() Line 68	C++
CefSharp.BrowserSubprocess.Core.dll!CefSharp.BrowserSubprocess.Serialization.SerializeV8Object<>() Line 111	C++
CefSharp.BrowserSubprocess.Core.dll!CefSharp::BrowserSubprocess::CefAppUnmanagedWrapper::OnProcessMessageReceived() Line 452	C++
[Native to Managed Transition]	
CefSharp.BrowserSubprocess.Core.dll!CefCToCppRefCounted::Wrap()	Unknown

The Log file seems unsuspicious:
[0914/092556.831:ERROR:gpu_init.cc(441)] Passthrough is not supported, GL is disabled
[0914/092806.924:ERROR:dns_config_service.cc(259)] DNS config watch failed.
[0914/092806.924:ERROR:dns_config_service.cc(259)] DNS config watch failed.
[0914/093534.409:ERROR:gpu_init.cc(441)] Passthrough is not supported, GL is disabled

Thank you for your great product! I hope this entry helps!
Best regards,
Ramin

Romout:
**Fixes:**
 Resolves #3808

   - Array serialization now correctly handles empty elements within the source array

   -No structural changes - use HasValue to check if an element is empty
      
**How Has This Been Tested?**  
Used WinForms example with several inputs to try and break it but - it works.

**Types of changes**

- [ ] Updated documentation

**Checklist:**

- [ ] Commented my code

### DIFF
--- a/CefSharp.BrowserSubprocess.Core/Serialization/V8Serialization.cpp
+++ b/CefSharp.BrowserSubprocess.Core/Serialization/V8Serialization.cpp
@@ -59,15 +59,12 @@ namespace CefSharp
                 else if (obj->IsArray())
                 {
                     int arrLength = obj->GetArrayLength();
-                    std::vector<CefString> keys;
                     auto array = CefListValue::Create();
-                    if (arrLength > 0 && obj->GetKeys(keys))
+                    if (arrLength > 0)
                     {
-                        int index = 0;
                         for (int i = 0; i < arrLength; i++)
                         {
-                            if (obj->HasValue(i))
-                                SerializeV8Object(obj->GetValue(keys[index++]), array, i, callbackRegistry, seen);
+                            SerializeV8Object(obj->GetValue(i), array, i, callbackRegistry, seen);
                         }
                     }
 
--- a/CefSharp.BrowserSubprocess.Core/Serialization/V8Serialization.cpp
+++ b/CefSharp.BrowserSubprocess.Core/Serialization/V8Serialization.cpp
@@ -63,11 +63,12 @@ namespace CefSharp
                     auto array = CefListValue::Create();
                     if (arrLength > 0 && obj->GetKeys(keys))
                     {
+                        int index = 0;
                         for (int i = 0; i < arrLength; i++)
                         {
-                            SerializeV8Object(obj->GetValue(keys[i]), array, i, callbackRegistry, seen);
+                            if (obj->HasValue(i))
+                                SerializeV8Object(obj->GetValue(keys[index++]), array, i, callbackRegistry, seen);
                         }
-
                     }
 
                     list->SetList(index, array);
--- a/CefSharp.Test/JavascriptBinding/IntegrationTestFacts.cs
+++ b/CefSharp.Test/JavascriptBinding/IntegrationTestFacts.cs
@@ -230,5 +230,28 @@ namespace CefSharp.Test.JavascriptBinding
                 Assert.Equal(JavascriptObjectRepository.AllObjects, evt.Arguments.ObjectName);
             }
         }
+
+        [Fact]
+        public async Task CanSerializePartiallyEmptyArrays()
+        {
+            using (var browser = new ChromiumWebBrowser(CefExample.HelloWorldUrl))
+            {
+                var testCases = new[]
+                {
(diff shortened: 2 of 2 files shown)


amaitland:
> Used WinForms example with several inputs to try and break it but - it works

Unit test will need to be added, https://github.com/cefsharp/CefSharp/blob/master/CefSharp.Test/OffScreen/OffScreenBrowserBasicFacts.cs#L139 Can be used as a rough guide.

amaitland:
Looking at the actual code now and the [history](https://github.com/cefsharp/CefSharp/blob/cefsharp/43/CefSharp.BrowserSubprocess.Core/Serialization/V8Serialization.cpp#L62) would suggest the section is basically unchanged since 2015. So the questions I have are was there a reason we used keys to begin with? If yes is that reason still valid? A straight index would have made more sense (I didn't write the code, though I likely reviewed it all those years ago).

I'd be inclined to simplify the code as below:

```c++
else if (obj->IsArray())
{
	int arrLength = obj->GetArrayLength();
	auto array = CefListValue::Create();

	for (int i = 0; i < arrLength; i++)
	{
		SerializeV8Object(obj->GetValue(i), array, i, callbackRegistry, seen);
	}

	list->SetList(index, array);
}
```

I did test your code and it differs slightly in it's output when I compare to [JSON.stringify()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify) for one edge case (at least that's all I found in the 2 mins of testing).

```
[1, 2, ,] is returned as "[1,2,null]" for JSON.stringify where it's "[1, 2]" for the updated code.
```

As above test cases will be required before any chance can be merged.

Romout:
Hm, I like your version much more. You are right that keys are not even required if GetValue returns null if there isn't a value.

amaitland:
Test case rewritten to use `Theory with Inline Data` in https://github.com/cefsharp/CefSharp/commit/fa2b6e22e3a1c3368764d9c73421a96484eb10e6

PR #3841, opened 2021-10-09
Net Core - Don't check for presence of "d3dcompiler_47.dll" on ARM64

kpreisser:
Don't check for presence of `d3dcompiler_47.dll` when running as ARM64, because that file isn't present in the CEF redist for this architecture.
Currently, the .NET SDK copies a stub `D3DCompiler_47.dll` file to the application's directory (so the dependency check currently succeeds), but once that is fixed and that file is no longer deployed, the dependency check would fail on ARM64.

Issue #2944

**Summary:**
   - Don't check for presence of `d3dcompiler_47.dll` when current architecture is `arm64`.

**Changes:**
  - Changed field initializer of `DependencyChecker.CefOptionalDependencies` to not include `d3dcompiler_47.dll` in the array when running on .NET Core and `RuntimeInformation.ProcessArchitecture` is `Architecture.Arm64`.
      
**How Has This Been Tested?**  
- Published `CefSharp.MinimalExample.Wpf.netcore` for `win-arm64` (target framework `net5.0-windows`).
- Deleted file `D3DCompiler_47.dll` from the published files and removed the corresponding entry from `CefSharp.MinimalExample.Wpf.netcore.deps.json`.
- Copied the published files to a Windows 11 ARM64 Build 22000 machine and run the application.
- Verified that before the change, the app fails with exception `Unable to locate required Cef/CefSharp dependencies: Missing:C:\..\d3dcompiler_47.dll`, and after the change, the app starts successfully.
- Also verified that when performing the same steps for a `x64` machine, after the change when deleting file `d3dcompiler_47.dll`, the exception is still thrown as expected.

**Types of changes**

- [ ] Updated documentation

**Checklist:**

- [ ] Commented my code

### DIFF
--- a/CefSharp/DependencyChecker.cs
+++ b/CefSharp/DependencyChecker.cs
@@ -5,7 +5,13 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+#if NETCOREAPP
+using System.Linq;
+#endif
 using System.Reflection;
+#if NETCOREAPP
+using System.Runtime.InteropServices;
+#endif
 using System.Text;
 
 namespace CefSharp
@@ -20,6 +26,11 @@ namespace CefSharp
         /// en-US Locales pak file location
         /// </summary>
         public const string LocalesPackFile = @"locales\en-US.pak";
+
+        /// <summary>
+        /// File name of the Direct3D Compiler DLL.
+        /// </summary>
+        private const string D3DCompilerDll = "d3dcompiler_47.dll";
 
         /// <summary>
         /// List of Cef Dependencies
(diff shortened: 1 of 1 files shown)


kpreisser:
Thank you, I changed the code accordingly to remove the `d3dcompiler_47.dll` entry in the static constructor.

Do you prefer to avoid having the `#if NETCOREAPP` directives in the top `using` block? It would be possible, but (I think) make the code less readable, e.g.:
```cs
#if NETCOREAPP
        static DependencyChecker()
        {
            // win-arm64 doesn't ship with a copy of the D3D Compiler, it's included with the OS.
            if (System.Runtime.InteropServices.RuntimeInformation.ProcessArchitecture is System.Runtime.InteropServices.Architecture.Arm64)
            {
                CefOptionalDependencies = System.Linq.Enumerable.ToArray(System.Linq.Enumerable.Where(CefOptionalDependencies, x => x != D3DCompilerDll));
            }
        }
#endif
```

amaitland:
Great, thanks 👍 

I think the build errors are totally unrelated.

> Do you prefer to avoid having the `#if NETCOREAPP` directives in the top `using` block? It would be possible, but (I think) make the code less readable, e.g.:

Either is fine really. If it's just a single entry then I just normally use the VS fixup and have to specify the full namespace.

PR #3846, opened 2021-10-10
WPF - IME Only use GCS_CURSORPOS if an adjacent character is new input

igandrews:
This is the change I discussed [here](https://github.com/cefsharp/CefSharp/issues/1262#issuecomment-934661876) in issue 1262. Namely, we should follow the WPF IME implementation and only use the GCS_CURSORPOS if either the character prior or after the cursor position is ATTR_INPUT (i.e. a new character entered by the user). Otherwise the cursor should be left at the end of the string. This fixes the problem described in that issue where pressing the space key was positioning the caret at the beginning of the input when it should have been placed at the end.

**Fixes:** [[1262](https://github.com/cefsharp/CefSharp/issues/1262)] 

**Changes:**
   - Changed the GetCompositionSelectionRange to return the attributes it was already retreiving as that contains the state of the characters in the composition.
   - Use those attributes and only keep/honor the cursor position if the attributes of the character before or after that offset is new. Otherwise use the end of the string as the cursor position.
      
**How Has This Been Tested?**  
Manually using the Microsoft Japanese IME keyboard with Hiragana.

**Types of changes**
- [ ] Updated documentation

**Checklist:**
- [x] Commented my code

### DIFF
--- a/CefSharp.Wpf/Internals/IMEHandler.cs
+++ b/CefSharp.Wpf/Internals/IMEHandler.cs
@@ -101,7 +101,7 @@ namespace CefSharp.Wpf.Internals
 
             if (attributes != null &&
                 // character before
-                ((compositionStart >= 0 && (compositionStart - 1) < attributes.Length && attributes[compositionStart - 1] == ImeNative.ATTR_INPUT)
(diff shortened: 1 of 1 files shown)


amaitland:
Thanks 👍 

Same disclaimer as always, merging `IME` changes without testing them. 

Will include in `M95`.

PR #3998, opened 2022-02-11
Update docs for IResourceRequestHandlerFactory

craigfowler:
I have just discovered that when `HasHandlers` returns `false`, the `GetResourceRequestHandler` method is never executed.  Presumably this is an optimisation which happens inside of CEF, as I can see that the `HasHandlers` property value goes all way back to the C++ native project.

I was about to suggest removing this optimisation because it'd be trivial to optimise and return early inside the default factory impl in C# world.  If it goes all the way down to CEF then that's less easy and so the next best thing would be to document it.

**Fixes:** _Has no related issue_

**Summary:** Addition to docs (more info above)

**Changes:** `IResourceRequestHandlerFactory` (comments only)
      
**How Has This Been Tested?**  It has not, no functional changes made

**Types of changes**
- [x] Updated documentation

**Checklist:**
- [x] Commented my code

### DIFF
--- a/CefSharp/IResourceRequestHandlerFactory.cs
+++ b/CefSharp/IResourceRequestHandlerFactory.cs
@@ -16,6 +16,13 @@ namespace CefSharp
         /// <summary>
         /// Are there any <see cref="ResourceHandler"/>'s registered?
         /// </summary>
+        /// <remarks>
+        /// <para>
+        /// Implementors must return <see langword="true"/> from this method if the factory is to be used.
+        /// The <see cref="GetResourceRequestHandler"/> method will not be executed if this property returns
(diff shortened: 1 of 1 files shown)


amaitland:
Always appreciated when someone helps improve the documentation, thanks 👍

amaitland:
Looking at the property xml doc doesn't make sense anymore, it likely predates the `Network Service` changes in version `75`.

Additional changes made in commit https://github.com/cefsharp/CefSharp/commit/e4e6cb8f7a0023b1b7e05f57ca44fe2314691447

Technically `HasHandlers` could be removed, as it's a breaking change we'll leave as is for now.

PR #4063, opened 2022-04-13
AnyCPU - Simply support via ModuleInitializer

amaitland:
**Summary:**
Now that we use the SDK Style csproj files and target VS2019 we can use a custom attribute
(same as in .Net Core 3.1) to get the compiler to output a ModuleInitializer.
ModuleInitializer can be used to resolve AnyCpu dlls at runtime. Should greatly simplify AnyCpu
support.

**Changes:**
  - Checks if the Assembly.GetEntryAssembly() is MSIL
  - Change CefSharp.Core.csproj to use LangVersion 9 so we can get the compiler to output
    the ModuleInitializer
  - Remove CefSharpPlatformCheck and other CefSharpAnyCpuSupport checks
  - CefSharp.Common.targets still has CefSharpAnyCpuSupport references for backwards compatibility with those with it set and Prefer32bit
    The default behaviour for this case is to simply copy the 32bit libs
  - Calling CefRuntime.SubscribeAnyCpuAssemblyResolver multiple times now will simply Unsubscribe the previous resolver

**How Has This Been Tested?**  
Further testing required.

**Types of changes**
- [x] Updated documentation

**Checklist:**
- [x] Commented my code


amaitland:
This will need more testing before merging.

amaitland:
On the `MSBuild` related code, rather than aggressively removing the code, we could just set the value to `True` effectively disabling the checks, just incase they needed to be rolled back quickly. Then actively removed at a later point in time.

Some additional extensibility points for the user and an even that'll make it easier to debug when the module initializer has run, possibly the ability to cancel it's execution allowing the user to perform their own action directly.

- [ ] Allow event to trigger when module initializer has run (with option to cancel the execution)
- [ ] Allow path configuration
- [ ] Default `CefSharpAnyCpuSupport` to true for short term.

PR #4067, opened 2022-04-16
UpdateNugetPackages takes parameters and fixes common.nuspec requirements based on arch

### LINKED ISSUE #4066: Building only select architectures does not have proper package requirements for .net framework.
If you want to only build x64 or x86 both cef dist files are still required by the CefSharp.Common package.

mitchcapper:
Closes #4066 

**Summary:** 
The primary goal here is to only include the architectures the user wants to build in the required packages.  There are also some code cleanup options  for UpdateNugetPackages.ps1 to make it usable by more people.   I debated if this should go in build.ps1 or UpdateNugetPackages.ps1 but it seemed more appropriate here (as the rest of the script is also doing version updates).   

**Changes:** 

- Have the script take parameters rather than hard coded values, default to old hardcoded values if not specified
- Add BuildArches parameter to the script like build.ps1, this is a requirement to know what final architectures the user wants to build for
- Enabled script mode for the script, this required changing some of the node selection to xpath to avoid null child accesses
- Use same nuget path and download function as build.ps1, I would assume no reason to use ../nuget
- Major change: Edit the `NuGet\CefSharp.Common.nuspec` file to only require the cef dist files for the build users want.  Rather than having an input file, or hardcoding the node data for the xml node we want added/removed we will just use the existing one as the template.
      - Added support for Async binding to return Task<T>
      - Added new QUnit Test cases
      
**How Has This Been Tested?**  
Yes tried setting different values and still seemed to update everything. I didn't see any files with EnsureNuGetPackageBuildImports targets any more but I believe that xpath should be correct.

**Types of changes**

- [ ] Updated documentation

**Checklist:**

- [x] Commented my code

### DIFF
--- a/UpdateNugetPackages.ps1
+++ b/UpdateNugetPackages.ps1
@@ -1,19 +1,85 @@
 #requires -Version 5
+[CmdletBinding()]
 
+param(
+	[Parameter(Position = 1)]
+	[string] $CefVersion = "100.0.12",
+	[Parameter(Position = 2)]
+	[string] $CefSharpVersion = "",
+	[Parameter(Position = 3)]
+	[string] $BuildArches = "x86 x64 arm64"
+	)
 # Update projects files
 # I haven't found a clean solution that allows for using just nuget.exe and dotnet.exe to do this
 # Update the vcxproj files first
 # Update the .Net csproj files modifying the xml file directly
+$ARCHES = [System.Collections.ArrayList]$BuildArches.ToLower().Split(" ");
+Set-StrictMode -version latest;
+$ErrorActionPreference = "Stop";
 
-$CefVersion = '100.0.12'
-$CefSharpVersion = $CefVersion + "0"
+if ($CefSharpVersion -eq "")
+{
+	$CefSharpVersion = $CefVersion + "0"
+}
+
+function UpdateNupsecCommonArches()
+{
+	$nuspecFile = (Resolve-Path 'NuGet\CefSharp.Common.nuspec');
+	$CefSharpNugetXml = [xml](Get-Content $nuspecFile );
+	$ns = new-object Xml.XmlNamespaceManager $CefSharpNugetXml.NameTable
+	$ns.AddNamespace("ns", $CefSharpNugetXml.DocumentElement.NamespaceURI)
+
+	$srcNode = $CefSharpNugetXml.SelectSingleNode("//ns:package/ns:metadata/ns:dependencies/ns:group/ns:dependency",$ns);
+
+	$parentNode = $srcNode.parentNode;
+	$parentNode.RemoveAll();
+
+
+	$ARCHES.Remove("arm64")
+
+	foreach ($arch in $ARCHES)
+	{
+		$clone = $srcNode.CloneNode($true);
+		$clone.SetAttribute("id","cef.redist." + $arch);
+		$parentNode.AppendChild( $clone ) | Out-Null;
+
+
+	}
+
+	if ( $parentNode.get_HasChildNodes() -eq $false )
+	{
+		$parentNode.AppendChild( $srcNode ); # we need to keep one node on there as it is used as a template, for building other arches this group is ignored anyway
+	}
+
+	$CefSharpNugetXml.Save( $nuspecFile )
+}
+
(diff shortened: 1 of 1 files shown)


amaitland:
> * Use same nuget path and download function as build.ps1, I would assume no reason to use ../nuget

Your likely the first person to use `UpdateNugetPackages.ps1`, it was something I hacked together to make things easier when upgrading 😄 Reason for using a relative nuget.exe is that's what I always have on my machine. Having the script download if required makes sense 👍 

> * Major change: Edit the `NuGet\CefSharp.Common.nuspec` file to only require the cef dist files for the build users want.  Rather than having an input file, or hardcoding the node data for the xml node we want added/removed we will just use the existing one as the template.

This should have already been covered by [build.ps1](https://github.com/cefsharp/CefSharp/blob/master/build.ps1#L272) as part of generating the packages. Is it not working?

I'd prefer to keep all changes to the `nuspec` files as part of `build.ps1` if possible as it already has logic for dealing with the different architectures.

Do you need a specific target to use as part of your workflow? Happy to split that out as a new target if required

mitchcapper:
Mostly rather than using my own scripts before to change versions I figure if UpdateNugetPackages.ps1 can do it all the better.  Plus as it already has to happen on new versions officially makes it a bit easier.

As for the build.ps1 instead for the feature change, no problem and will investigate why it did not work for me in current form.  Last time I built with a custom version number and just x64 for the arches it still referenced the x86 redist package in the Common cefsharp package and failed to build.     

As for why, there is an annoying windows 10 21H2 docker issue where they did not put out windows base images that match the kernel version.    This forces you to use hyperV rather than straight process isolation and it takes forever.   This is fixed in w11 but not currently looking to upgrade. Building only x64 cuts time way down.

amaitland:
> Mostly rather than using my own scripts before to change versions I figure if UpdateNugetPackages.ps1 can do it all the better. Plus as it already has to happen on new versions officially makes it a bit easier.

Makes sense to use the same script if possible 👍 I'll merge the base changes later today so I can test when when I upgrade to `M101` shortly.

> As for the build.ps1 instead for the feature change, no problem and will investigate why it did not work for me in current form. Last time I built with a custom version number and just x64 for the arches it still referenced the x86 redist package in the Common cefsharp package and failed to build.

It's possible there's a mistake, happy to take a look when I get a minute. 

> As for why, there is an annoying windows 10 21H2 docker issue where they did not put out windows base images that match the kernel version. This forces you to use hyperV rather than straight process isolation and it takes forever. This is fixed in w11 but not currently looking to upgrade. Building only x64 cuts time way down.

Building only the arch you need sounds very reasonable to me 👍 I'm sticking with `Win 10` as hardware I have is too old to run  `Win 11`.

amaitland:
> It's possible there's a mistake, happy to take a look when I get a minute.

Confirming there is a problem, I've forgotten to remove the dependency. Should be a simple fix.

amaitland:
> I didn't see any files with EnsureNuGetPackageBuildImports targets any more but I believe that xpath should be correct.

I've tested locally and the `EnsureNuGetPackageBuildImports` aren't removed, adding the namespace to the xpath query got it working again.

amaitland:
Merged via cherry pick https://github.com/cefsharp/CefSharp/commit/cfa5b4185a2a0fce91969e0b7a8b81f4fc2d3dbf

Additional changes made in commit https://github.com/cefsharp/CefSharp/commit/78b682a8729bc5bd452697be047b318413aa57ec

`build.ps1` handles changing `CefSharp.common.nuspec`. Fixed the `EnsureNuGetPackageBuildImports` and did a little reordering of code so functions are first.

First commit to use the updated script is https://github.com/cefsharp/CefSharp/commit/684e76c83641a620bf72540a4c14a1972f7cda24

Everything looks OK so far.

> Confirming there is a problem, I've forgotten to remove the dependency. Should be a simple fix.

The nuspec dep should be fixed in commit https://github.com/cefsharp/CefSharp/commit/80ea9212c9b774c884499d437c4f2c4a4dfeb58e

If you are still having problems with https://github.com/cefsharp/CefSharp/issues/4066 then please let me know.

PR #4119, opened 2022-05-25
Fix FolderSchemeHandlerFactory locking files

uecasm:
**Fixes:**
  - #2535 (better than this)

   - Fix `FolderSchemeHandlerFactory` keeping files locked for the process lifetime

If a page or other resource is loaded via a scheme that uses `FolderSchemeHandlerFactory`, subsequently attempting to delete or overwrite the corresponding file results in a "file in use" exception -- even if the browser has navigated elsewhere or disposed beforehand.

This appears to be because `ResourceHandler.FromStream` will by default not close streams even though (in this case) they are only used for a single request.  So we ought to tell it to do so.  (It would be safer if `FromStream` did dispose by default, but that would be a breaking change.)
      
**How Has This Been Tested?**  
I created a copy of this class in my test app with this change and performed some navigations where it deletes the loaded file (and creates a new one) before doing the next navigation.  With this change, it works.

**Types of changes**

- [ ] Updated documentation

**Checklist:**

- [ ] Commented my code

### DIFF
--- a/CefSharp/SchemeHandler/FolderSchemeHandlerFactory.cs
+++ b/CefSharp/SchemeHandler/FolderSchemeHandlerFactory.cs
@@ -117,7 +117,7 @@ namespace CefSharp.SchemeHandler
                 var fileExtension = Path.GetExtension(filePath);
                 var mimeType = GetMimeTypeDelegate(fileExtension);
                 var stream = new FileStream(filePath, FileMode.Open, FileAccess.Read, resourceFileShare);
-                return ResourceHandler.FromStream(stream, mimeType);
+                return ResourceHandler.FromStream(stream, mimeType, autoDisposeStream: true);
             }
 
             return ResourceHandler.ForErrorMessage("File Not Found - " + filePath, HttpStatusCode.NotFound);


uecasm:
FWIW, `ResourceHandler.FromFilePath` is similarly vulnerable and really ought to default to disposing the stream.  Where the resource handler instance is reused for multiple requests (e.g. if cached rather than a new one being created each time) then not disposing is required, but on the other hand that method does not actually expose the underlying stream to allow disposing it when the resource handler itself is eventually destroyed, so the stream handle is invariably leaked for the entire application lifetime.

amaitland:
Thanks 👍 

I think a unit test should be added, I'll look to see which current ones can be used as a reference. 

> FWIW, ResourceHandler.FromFilePath is similarly vulnerable and really ought to default to disposing the stream. 

Please submit a PR with your suggest.  Thanks.

uecasm:
> Please submit a PR with your suggest.

I was deliberately not changing that since it's a breaking change and thus beyond my pay grade. 😁

Fundamentally it seems like a potential lifetime mismatch for `ResourceHandler`s in general.  I'm not familiar enough with the code to trace it but the implication of this interaction is that the `ResourceHandler.Dispose` is called at the end of each request, and yet `ResourceHandler`s can be either one-shot (as in this particular case) or cached and reused for multiple requests (as in `ResourceRequestHandlerFactory`).  If they're one-shot, then you want disposing the request to dispose the handler and the underlying stream.  If they're cached, then disposing the request should not really dispose the handler **or** the stream (and instead whatever is caching the instances ought to be the thing responsible for disposing the handler and the stream).  Or to put it another way: disposing the handler ought to always dispose the stream; the Real Bug™ is that the handler dispose happens at the wrong time sometimes, because it has too many "owners".  But that's probably not straightforward to redesign, especially when working around CEF's behaviour and trying to not break existing code too much.

This PR, at least, is trying to be much simpler than that.

amaitland:
> I was deliberately not changing that since it's a breaking change and thus beyond my pay grade. 😁

A PR will allow for further discussion and consideration. The suggestion will likely be lost otherwise. 

> I'm not familiar enough with the code to trace it but the implication of this interaction is that the ResourceHandler.Dispose is called at the end of each request, and yet ResourceHandlers can be either one-shot (as in this particular case) or cached and reused for multiple requests (as in ResourceRequestHandlerFactory). If they're one-shot, then you want disposing the request to dispose the handler and the underlying stream

You cannot reuse an instant of ResourceHandler, the stream reference is freed.

ResourceRequestHandlerFactory stores response data as a byte[] from  memory. 

You can reuse a Stream. In hindsight this would have been done differently with each stream being disposed of. This is only the default implementation and if you require finer grained control you can implement the interfaces directly. 

As for when the ResourceHandler is disposed, that's entirely up to CEF. 

> but it leaks the stream by default, which is not a great default.

When CefSharp creates the stream then disposing of the stream makes sense.

The default should have been to auto dispose, making some of the static methods obsolete and new ones added with different name and behaviour is probably a reasonable move.

uecasm:
> ResourceRequestHandlerFactory stores response data as a byte[] from memory.

Hmm, you're right.  I was interpreting the doc comments for the `autoDisposeStream` parameter to imply that there was some other situation in which the `ResourceHandler` might be re-used for multiple requests.  But if that's not the case then it seems even less sensible for `FromFilePath` to work the way it currently does.

amaitland:
The code has evolved over time, if you feel the documentation needs clarification then by all means make a suggest. 

If you'd like to expand the scope of the PR then please update the title and description.

amaitland:
This won't be included in the upcoming `102` release as more work/through is required.

You should easily be able to workaround the limitation with older versions using something like the following:

```c#
public class CustomFolderSchemeHandlerFactory : FolderSchemeHandlerFactory
{
	public CustomFolderSchemeHandlerFactory(string rootFolder, string schemeName = null, string hostName = null, string defaultPage = "index.html", FileShare resourceFileShare = FileShare.Read) : base(rootFolder, schemeName, hostName, defaultPage, resourceFileShare)
	{
	}

	protected override IResourceHandler Create(IBrowser browser, IFrame frame, string schemeName, IRequest request)
	{
		var resourceHandler = base.Create(browser, frame, schemeName, request) as ResourceHandler;
		
		if(resourceHandler == null)
		{
			return null;
		}

		resourceHandler.AutoDisposeStream = true;

		return resourceHandler;
	}
}
```

Code compiles, haven't run it so might need minor tweaks.

uecasm:
What work?  The discussion here was mostly unrelated to this PR.

amaitland:
> What work?

Firstly, as I stated above this will need unit tests. There's not really a copy and paste example for this, so I'll have to write one. I'm working on something else at the moment.

I'd also like for the other related issue to be addressed at the same time.

amaitland:
If your company is currently [sponsoring](https://github.com/sponsors/amaitland) the project then please confirm the GitHub account name, I'll look to reprioritise this.

amaitland:
Commit https://github.com/cefsharp/CefSharp/commit/550552f2108d36f6a073aab6dd8b5dcfb297fb82 adds a basic `FolderSchemeHandler` factory test. This should be usable as a base for additional tests to validate this change.

`M103` should be out in a week or so, if you can write a test before then we can look at include in the next release.

PR #4266, opened 2022-10-07
Copy link to highlight how can we implement this

Daszbin:
I was wondering how can we implement copy link to highlight through the contextmenuhandler and is their an API in CEF to implement this feature.


amaitland:
Use https://github.com/cefsharp/CefSharp/discussions to ask questions.

PR #4307, opened 2022-11-16
WPF - Add simplified resize hack (disabled by default)

Wupval:
**Fixes:**
   - Fixes issue 4274

**Summary:**
   - Re-Added the resize hack for issue 2779, as the issue still occurs. Disabled by default as mentioned by @amaitland in https://github.com/cefsharp/CefSharp/discussions/4274#discussioncomment-3946056

**Changes:**
   - Re-Added the resize hack for issue 2779
   - Ported changes to match master (Execution of the resize hack has been moved to OnBrowserWasHidden)
      
**How Has This Been Tested?**  
   - This was tested with the CefSharp minimal example, creating multiple browsers (at least 6 are needed) and toggling the visibility of the browsers.
   - OS: Win10, x64

**Types of changes**
- [ ] Updated documentation

### DIFF
--- a/CefSharp.Wpf/ChromiumWebBrowser.cs
+++ b/CefSharp.Wpf/ChromiumWebBrowser.cs
@@ -140,11 +140,28 @@ namespace CefSharp.Wpf
         /// </summary>
         private static bool DesignMode;
 
+        private bool resizeHackForIssue2779Enabled;
+        private Structs.Size? resizeHackForIssue2779Size;
+
         /// <summary>
         /// This flag is set when the browser gets focus before the underlying CEF browser
         /// has been initialized.
         /// </summary>
         private bool initialFocus;
+
+        /// <summary>
+        /// Hack to work around issue https://github.com/cefsharp/CefSharp/issues/2779
+        /// Disabled by default
+        /// </summary>
+        public bool EnableResizeHackForIssue2779 { get; set; }
+
+        /// <summary>
+        /// Number of milliseconds to wait after resizing the browser when it first
+        /// becomes visible. After the delay the browser will revert to it's
+        /// original size.
+        /// Hack to work around issue https://github.com/cefsharp/CefSharp/issues/2779
+        /// </summary>
+        public int ResizeHackForIssue2779DelayInMs { get; set; }
 
         /// <summary>
         /// Gets a value indicating whether this instance is disposed.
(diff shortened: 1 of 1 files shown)


CefSharp.Wpf/ChromiumWebBrowser.cs
```diff
@@ -1759,11 +1795,26 @@ private void OnWindowStateChanged(object sender, EventArgs e)
...
-        protected virtual void OnBrowserWasHidden(bool hidden)
+        protected virtual async void OnBrowserWasHidden(bool hidden)
         {
             if (browser != null)
             {
                 browser.GetHost().WasHidden(hidden);
+
+                if (hidden)
+                {
+                    if (EnableResizeHackForIssue2779)
```
amaitland, line 1806: This if statement should be moved up so the whole set of conditions is only evaluated if hack enabled.

CefSharp.Wpf/ChromiumWebBrowser.cs
```diff
@@ -2696,5 +2747,48 @@ public IBrowser GetBrowser()
...
+                    {
+                        action();
+                    });
+                }
+            }
+        }
+
+        private async Task ResizeHackForIssue2779()
+        {
+            if (EnableResizeHackForIssue2779)
```
amaitland, line 2771: Remove this if statement, method needs to only be called if enabled.

CefSharp.Wpf/ChromiumWebBrowser.cs
```diff
@@ -1759,11 +1795,26 @@ private void OnWindowStateChanged(object sender, EventArgs e)
...
+                if (hidden)
+                {
+                    if (EnableResizeHackForIssue2779)
+                    {
+                        resizeHackForIssue2779Enabled = true;
+                    }
+                }
+                else
+                {
+                    await CefUiThreadRunAsync(async () =>
```
amaitland, line 1813: Let's remove the async modifier from the method and discard the Task returned from the method. 

There's no continuation.

amaitland (changes requested):
Comments inline.

Wupval:
@amaitland Is there an error in the code or any other reason the build seems to always fail?

amaitland:
Build failure is unrelated to this PR.

amaitland:
Follow up commit https://github.com/cefsharp/CefSharp/commit/0131dd8a331a3f500cf1c06da2b85307e9fc6de5

Properties have been renamed, no longer reference issue #2779, directly reference the upstream issue.

amaitland:
This is a greatly simplified version of the resize hack originally used. If anyone experiences any issues then I suggest you override `OnBrowserWasHidden` and `OnPaint` to implement your own version.

Original version for reference https://github.com/cefsharp/CefSharp/commit/9140c08a34e113b8865a80a7ac9f8b4aa19631c4

Issue #4362, opened 2023-01-13
Net Core - Unable to use Central Package Management


hhu-mahmoud:
The project need the Nuget Package : "CefSharp.Wpf.HwndHost" to be installed, but if I want to use the "Central Package Version Management" it throw always exception: 
"error NU1008: Projects that use central package version management should not define the version on the PackageReference items but on the PackageVersion items: CefSharp.Wpf.HwndHost.
15>Done building project "Application.csproj" -- FAILED."

To produce the exception:

1 - Create an empty project
2 - Add CefSharp and the Nuget Package "CefSharp.Wpf.HwndHost" 
3 - Use CPM => Create under "Solution Items" The CPM file "Directory.Package.props" and enable the "ManagePackageVersionsCentrally"
```xml
<Project>
  <PropertyGroup>
    <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
  </PropertyGroup>
  <ItemGroup>
    <PackageVersion Include="CefSharp.Common" Version="108.4.130" />
    <PackageVersion Include="CefSharp.Wpf.HwndHost" Version="108.4.130" />
    <PackageVersion Include="CefSharp.Wpf.NETCore" Version="108.4.130" />
  </ItemGroup>
</Project>
```

Then remove the Version of Nuget Packages from the "csproj"

4- Build the project and you will see the exception

amaitland:
Please use [Fenced Code Blocks](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-and-highlighting-code-blocks) to format code.

> The project need the Nuget Package : "CefSharp.Wpf.HwndHost" to be installed

Was it working when you just referenced `CefSharp.Wpf.NETCore`?

> "error NU1008: Projects that use central package version management should not define the version on the PackageReference items but on the PackageVersion items: CefSharp.Wpf.HwndHost.

Likely someone will need to reach out to the relevant `Microsoft` team for some guidance.

hhu-mahmoud:
> Was it working when you just referenced `CefSharp.Wpf.NETCore`?

Yes it works also but with the same problem

I think, because this 3 Packages are automatically installed and locked with the CefSharp.Wpf.NETCore or CefSharp.Wpf.HwndHost =>

![grafik](https://user-images.githubusercontent.com/32834850/212469861-9ea75438-b05a-4d79-bdbf-5aef6e7f9b94.png)

amaitland:
[...]

PR #4381, opened 2023-01-28
Add Compatibility with Central Package Management

ljdorey:
When a runtime identifier is NOT chosen, the default Nuget logic is to reference all possible compilations for compatibility purposes. However if Central Package Management, introduced in Nuget 6.2 ( https://devblogs.microsoft.com/nuget/introducing-central-package-management/ ) is enabled referencing the package directly in a targets file will causes an error. Additional since it is only imported at compile time, it triggers a infinite restore loop in visual studio.

**Fixes:** https://github.com/cefsharp/CefSharp/issues/4362

**Summary:** 
Updated CefSharp.Common.Netcore.Targets file to be compatible with Central Package Management. 

**Changes:** 
Added a check to see if Central package management is enabled and if so use "VersionOverride" to specify the fallback packages. 

**How Has This Been Tested?**  
Tested locally by modifying the existing, nuget packge to make sure it works, wasn't sure how to generate a Nuget test package, if you can point me at some docs I can do so if you'd like, my Organization has approved me some time to contribute and this change. 

**Types of changes**
- [ ] Updated documentation

**Checklist:**

- [ ] Commented my code

### DIFF
--- a/NuGet/PackageReference/CefSharp.Common.NETCore.targets
+++ b/NuGet/PackageReference/CefSharp.Common.NETCore.targets
@@ -142,14 +142,18 @@
     <Otherwise>
       <Choose>
         <When Condition="$(ManagePackageVersionsCentrallytrue) == 'true'">
-          <PackageReference Include="chromiumembeddedframework.runtime.win-x64" VersionOverride="109.1.11" />
-          <PackageReference Include="chromiumembeddedframework.runtime.win-x86" VersionOverride="109.1.11" />
-          <PackageReference Include="chromiumembeddedframework.runtime.win-arm64" VersionOverride="109.1.11" />
+          <ItemGroup>
+            <PackageReference Include="chromiumembeddedframework.runtime.win-x64" VersionOverride="109.1.11" />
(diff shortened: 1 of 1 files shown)


amaitland:
Thanks for the `PR` 👍 

> wasn't sure how to generate a Nuget test package

The build script will create the nuget packages. See https://github.com/cefsharp/CefSharp/blob/master/appveyor.yml#L11 for example.

The build script will also need to be modified to update the version numbers of the new entries you've added.
https://github.com/cefsharp/CefSharp/blob/master/build.ps1#L440

NuGet/PackageReference/CefSharp.Common.NETCore.targets
```diff
@@ -140,10 +140,23 @@
       </ItemGroup>
     </When>
     <Otherwise>
+      <Choose>
+        <When Condition="$(ManagePackageVersionsCentrallytrue) == 'true'">
```
amaitland, line 144: This looks like a mistake, should be:

```
'$(ManagePackageVersionsCentrally)' == 'true'
```

amaitland:
Closing as issue resolved in commit https://github.com/cefsharp/CefSharp/commit/5b61574a51c943289739573e6ebc414111158c3a

PR #4639, opened 2023-11-30
WPF - Fixed Html Dropdown positioning when near bottom of screen

DidiDerDenker:
**Fixes:** #2820

**Summary:** Fixed positioning in order to prevent html dropdown to go off screen when its near the bottom of a page.

**Changes:** Added mouse-teleport-class in order to reset the position of a dropdown, used in chromium-web-browser.
     
**How Has This Been Tested?**  
The assemblies are able to build. The example of the solution worked. The example that was pointed out in issue 2820 worked.

**Types of changes**
- [x ] Bug fix (non-breaking change which fixes an issue)
- [ ] Updated documentation

**Checklist:**
- [x ] Tested the code (if applicable)
- [x] Commented my code
- [ ] Changed the documentation (if applicable)
- [x ] The formatting is consistent with the project (project supports .editorconfig)

Big thanks to "carrierdown". He provied the solution a few month ago, but it was never merged as a pull request.


CefSharp.Wpf/Internals/MouseTeleport.cs
```diff
@@ -0,0 +1,55 @@
+using CefSharp.Structs;
+
+namespace CefSharp.Wpf.Internals
+{
+    public class MouseTeleport
+    {
+        public Rect teleportingRect { get; private set; }
```
amaitland, line 7: `public` properties should start with a capital letter.

These should probably be private fields.

CefSharp.Wpf/Internals/MouseTeleport.cs
```diff
@@ -0,0 +1,55 @@
+using CefSharp.Structs;
+
+namespace CefSharp.Wpf.Internals
+{
+    public class MouseTeleport
```
amaitland, line 5: Terrible class name IMHO, doesn't tell you anything about it's purpose for being.

It translates/adjusts mouse coordinates when popup is visible.

CefSharp.Wpf/ChromiumWebBrowser.cs
```diff
@@ -2276,7 +2323,11 @@ protected override void OnMouseMove(MouseEventArgs e)
                 var point = e.GetPosition(this);
                 var modifiers = e.GetModifiers();
 
-                browser.GetHost().SendMouseMoveEvent((int)point.X, (int)point.Y, false, modifiers);
+                if (!mouseTeleport.IsInsideOriginalRect((int)point.X, (int)point.Y))
```
amaitland, line 2326: Under what instance do we want to ignore the mouse move?

These methods should simple call `GetAdjustedMouseCoords`, no logic should be in any of these methods.

CefSharp.Wpf/Internals/MouseTeleport.cs
```diff
@@ -0,0 +1,55 @@
...
+
+        public void Reset()
+        {
+            this.originalRect = new Rect();
+            this.originalRect = new Rect();
+            this.xOffset = 0;
+            this.yOffset = 0;
+        }
+
+        public bool IsInsideOriginalRect(int x, int y)
```
amaitland, line 34: This should be part of the internal implementation, I don't think this should be public

CefSharp.Wpf/Internals/MouseTeleport.cs
```diff
@@ -0,0 +1,55 @@
...
+
+        private bool IsInsideTeleportingRect(int x, int y)
+        {
+            return x >= this.teleportingRect.X &&
+                   x < this.teleportingRect.X + this.teleportingRect.Width &&
+                   y >= this.teleportingRect.Y &&
+                   y < this.teleportingRect.Y + this.teleportingRect.Height;
+        }
+
+        public Point GetAdjustedMouseCoords(int x, int y)
```
amaitland, line 50: This should take the `Point`, if `popup` is not visible then simple return the same point.

If `popup` is visible and requires translation then return translated point.

CefSharp.Wpf/ChromiumWebBrowser.cs
```diff
@@ -1037,7 +1042,13 @@ void IRenderWebBrowser.OnPopupShow(bool isOpen)
         /// <param name="isOpen">if set to <c>true</c> [is open].</param>
         protected virtual void OnPopupShow(bool isOpen)
         {
-            UiThreadRunAsync(() => { popupImage.Visibility = isOpen ? Visibility.Visible : Visibility.Hidden; });
+            UiThreadRunAsync(() => {
```
amaitland, line 1045: Opening brace should be on new line for consistency.

CefSharp.Wpf/ChromiumWebBrowser.cs
```diff
@@ -1037,7 +1042,13 @@ void IRenderWebBrowser.OnPopupShow(bool isOpen)
...
+                popupImage.Visibility = isOpen ? Visibility.Visible : Visibility.Hidden;
+                if (!isOpen)
+                {
+                    mouseTeleport.Reset();
```
amaitland, line 1049: Should pass in `isopen` to the class.

CefSharp.Wpf/ChromiumWebBrowser.cs
```diff
@@ -2108,8 +2119,44 @@ private void SetPopupSizeAndPositionImpl(Rect rect)
             popupImage.Width = rect.Width;
             popupImage.Height = rect.Height;
 
-            Canvas.SetLeft(popupImage, rect.X);
-            Canvas.SetTop(popupImage, rect.Y);
+            int x = rect.X,
```
amaitland, line 2122: This whole block of logic would be better in the actual translation class.

Then it can be easily unit tested.

CefSharp.Wpf/ChromiumWebBrowser.cs
```diff
@@ -2293,15 +2344,18 @@ protected override void OnMouseWheel(MouseWheelEventArgs e)
...
-                var pointX = (int)point.X;
-                var pointY = (int)point.Y;
 
-                browser.SendMouseWheelEvent(
-                    pointX,
-                    pointY,
-                    deltaX: isShiftKeyDown ? e.Delta : 0,
-                    deltaY: !isShiftKeyDown ? e.Delta : 0,
-                    modifiers: modifiers);
+                if (!mouseTeleport.IsInsideOriginalRect((int)point.X, (int)point.Y))
```
amaitland, line 2348: Under what instance do we want to ignore the mouse wheel?

These methods should simple call `GetAdjustedMouseCoords`, no logic should be in any of these methods.

CefSharp.Wpf/ChromiumWebBrowser.cs
```diff
@@ -2444,12 +2498,21 @@ private void OnMouseButton(MouseButtonEventArgs e)
...
                     var clickCount = e.ClickCount;
 
-                    if(clickCount > 3)
+                    if (clickCount > 3)
                     {
                         clickCount = 1;
                     }
 
-                    browser.GetHost().SendMouseClickEvent((int)point.X, (int)point.Y, (MouseButtonType)e.ChangedButton, mouseUp, clickCount, modifiers);
+                    if (mouseTeleport.IsInsideOriginalRect((int)point.X, (int)point.Y))
```
amaitland, line 2506: Under what instance do we want to ignore the mouse click?

These methods should simple call `GetAdjustedMouseCoords`, no logic should be in any of these methods.

amaitland (changes requested):
Thanks for the PR!

This will need some major refactoring before it can be considered for merge.

amaitland:
> * Commented my code

New class needs to be commented.

CefSharp.Wpf/Internals/MouseTeleport.cs
DidiDerDenker: I changed them to private fields.

CefSharp.Wpf/Internals/MouseTeleport.cs
DidiDerDenker: Thanks for your advice, I appreciate that and agree. I renamed it to "MouseAdjustor". Is that okay?

CefSharp.Wpf/ChromiumWebBrowser.cs
DidiDerDenker: I moved the logic to the class.

CefSharp.Wpf/Internals/MouseTeleport.cs
DidiDerDenker: I made it private.

CefSharp.Wpf/Internals/MouseTeleport.cs
DidiDerDenker: I adapted your comment as I unterstood it. Hope it's okay.

CefSharp.Wpf/ChromiumWebBrowser.cs
DidiDerDenker: Done.

CefSharp.Wpf/ChromiumWebBrowser.cs
DidiDerDenker: Moved it to the class.

CefSharp.Wpf/ChromiumWebBrowser.cs
DidiDerDenker: I moved the logic to the class.

CefSharp.Wpf/ChromiumWebBrowser.cs
DidiDerDenker: I moved the logic to the class.

DidiDerDenker:
> This will need some major refactoring before it can be considered for merge.

🢂 Thank you so much for reviewing the PR! Your comments were very helpful and I tried my best to change the code accordingly. I hope the PR gets closer to being considered for merge.

> New class needs to be commented.

🢂 Done.

CefSharp.Wpf/ChromiumWebBrowser.cs
```diff
@@ -1037,7 +1042,15 @@ void IRenderWebBrowser.OnPopupShow(bool isOpen)
...
         protected virtual void OnPopupShow(bool isOpen)
         {
-            UiThreadRunAsync(() => { popupImage.Visibility = isOpen ? Visibility.Visible : Visibility.Hidden; });
+            UiThreadRunAsync(() =>
+            {
+                popupImage.Visibility = isOpen ? Visibility.Visible : Visibility.Hidden;
+
+                if (!isOpen)
+                {
+                    mouseAdjustor.Reset();
```
amaitland, line 1051: If you rename `Reset` to `OnPopupShow` and always call it passing in `isOpen` then we can track if the popup is open internally.

CefSharp.Wpf/Internals/MouseAdjustor.cs
```diff
@@ -0,0 +1,132 @@
...
+        }
+
+        /// <summary>
+        /// Adjusts the mouse-coordinates when the popup is visible.
+        /// </summary>
+        /// <param name="point">The original point.</param>
+        /// <returns>The adjusted point if needed, else the original point.</returns>
+        public Point GetAdjustedMouseCoords(System.Windows.Point point)
+        {
+            return !this.IsInsideOriginalRect(point) && IsInsideTeleportingRect(point)
```
amaitland, line 101: We shouldn't need to call these methods when there is no popup currently active or when then the popup is active and it it's shown in it's default location.

We should track this with a simple boolean property for performance reasons.

CefSharp.Wpf/Internals/MouseAdjustor.cs
```diff
@@ -0,0 +1,132 @@
...
+                ? new Point((int)point.X + this.xOffset, (int)point.Y + this.yOffset)
+                : new Point((int)point.X, (int)point.Y);
+        }
+
+        /// <summary>
+        /// Checks if the given point is inside the original-rect.
+        /// </summary>
+        /// <param name="point">The point.</param>
+        /// <returns>A boolean.</returns>
```
amaitland, line 110: Ideally `returns` should tell you about the meaning of the boolean.

e.g. Returns true if the is inside original rectangle otherwise returns false.

CefSharp.Wpf/Internals/MouseAdjustor.cs
```diff
@@ -0,0 +1,132 @@
...
+                   point.X < this.originalRect.X + this.originalRect.Width &&
+                   point.Y >= this.originalRect.Y &&
+                   point.Y < this.originalRect.Y + this.originalRect.Height;
+        }
+
+        /// <summary>
+        /// Checks if the given point is inside the teleporting-rect.
+        /// </summary>
+        /// <param name="point">The point.</param>
+        /// <returns>A boolean.</returns>
```
amaitland, line 123: Again, the meaning of the return value should be in the comment, not just that it's a boolean.

CefSharp.Wpf/Internals/MouseAdjustor.cs
```diff
@@ -0,0 +1,132 @@
...
+                y = 0;
+                yOffset = prevY;
+            }
+
+            if (x != prevX || y != prevY)
+            {
+                this.xOffset = xOffset;
+                this.yOffset = yOffset;
+
+                Rect teleportingRect = new Rect(x, y, x + originalRect.Width, y + originalRect.Height);
```
amaitland, line 67: There's still some `teleporting` naming references that should ideally be updated to something nicer

CefSharp.Wpf/Internals/MouseAdjustor.cs
```diff
@@ -0,0 +1,132 @@
...
+        /// The original-rect.
+        /// </summary>
+        private Rect originalRect;
+
+        /// <summary>
+        /// The teleporting-rect.
+        /// </summary>
+        private Rect teleportingRect;
+
+        public System.Windows.Point Update(Rect originalRect, Rect viewRect)
```
amaitland, line 27: Method needs xml comment and probably to be renamed.

Maybe `UpdatePopupSizeAndPosition`?

CefSharp.Wpf/Internals/MouseAdjustor.cs
```diff
@@ -0,0 +1,132 @@
...
+
+        /// <summary>
+        /// The original-rect.
+        /// </summary>
+        private Rect originalRect;
+
+        /// <summary>
+        /// The teleporting-rect.
+        /// </summary>
+        private Rect teleportingRect;
```
amaitland, line 25: Please rename this variable.

amaitland (changes requested):
Thanks 👍 

Looks like you've made decent progress, still a little bit of tidying up to go.

CefSharp.Wpf/ChromiumWebBrowser.cs
amaitland: I've added another comment with more details in latest review.

CefSharp.Wpf/Internals/MouseTeleport.cs
amaitland: I've added another comment in the latest review with more details.

CefSharp.Wpf/Internals/MouseTeleport.cs
amaitland: Hmm, go with that for now. See if we can come up with something better later.

CefSharp.Wpf/ChromiumWebBrowser.cs
DidiDerDenker: Done.

CefSharp.Wpf/Internals/MouseAdjustor.cs
DidiDerDenker: Done.

CefSharp.Wpf/Internals/MouseAdjustor.cs
DidiDerDenker: Done.

CefSharp.Wpf/Internals/MouseAdjustor.cs
DidiDerDenker: Done.

CefSharp.Wpf/Internals/MouseAdjustor.cs
DidiDerDenker: Done.

CefSharp.Wpf/Internals/MouseAdjustor.cs
DidiDerDenker: Done.

CefSharp.Wpf/Internals/MouseAdjustor.cs
DidiDerDenker: Thank you again! This is the only comment where I am unsure about the implementation. Unfortunately, I have no idea how to get a bool for "isOpen" from the stack above. Do you have an idea?

amaitland: If you add a class level variable (at private field of type bool), you can keep track of if the popup is open and the position was transformed. 

if the bool is false simply return the point, if it's true then perform the checks.

- If popup is closed, then bool will be false
- If popup is open, yet not transformed then bool will be false
- If popup is open and transformed then bool will be true.

DidiDerDenker: Thank you. I added a commit to the PR with the changes.

CefSharp.Wpf/Internals/MouseAdjustor.cs
```diff
@@ -0,0 +1,150 @@
...
+        }
+
+        /// <summary>
+        /// Adjusts the mouse-coordinates when the popup is visible.
+        /// </summary>
+        /// <param name="point">The original point.</param>
+        /// <returns>The adjusted point if needed, else the original point.</returns>
+        public Point GetAdjustedMouseCoords(System.Windows.Point point)
+        {
+            this.isOpen = !this.IsInsideOriginalRect(point) && IsInsideAdjustedRect(point);
```
amaitland, line 116: `isOpen` should be calculated in the other methods, not in here. By this point you should already know if the popup is open, and it's position adjusted.

amaitland (changes requested):

CefSharp.Wpf/Internals/MouseAdjustor.cs
```diff
@@ -0,0 +1,150 @@
...
+        /// </summary>
+        private bool isOpen;
+
+        /// <summary>
+        /// Updates the size and the position of the popup.
+        /// </summary>
+        /// <param name="originalRect"></param>
+        /// <param name="viewRect"></param>
+        /// <returns>The adjusted point.</returns>
+        public System.Windows.Point UpdatePopupSizeAndPosition(Rect originalRect, Rect viewRect)
```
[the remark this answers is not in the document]
DidiDerDenker: Done.

CefSharp.Wpf/Internals/MouseAdjustor.cs
```diff
@@ -0,0 +1,150 @@
...
+
+        /// <summary>
+        /// Resets the offsets and original-rect.
+        /// <param name="isOpen">If the popup is open or not.</param>
+        /// </summary>
+        public void OnPopupShow(bool isOpen)
+        {
+            if (!isOpen)
+            {
+                this.xOffset = 0;
```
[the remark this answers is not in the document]
DidiDerDenker: Done.

CefSharp.Wpf/Internals/MouseAdjustor.cs
DidiDerDenker: Are you sure? I looked for a solution but it was not that clean, since I had to call "IsInsideOriginalRect" and "IsInsideAdjustedRect" multiple times instead of right here. How would you envisage this in concrete terms?

amaitland: I think you've misunderstood. I would have expected this method to look something like the following (written in notepad, so not tested)

```c#
public Point GetAdjustedMouseCoords(System.Windows.Point point)
{
// Popup isn't open, or it's open and it's position was unchanged, so we don't need to perform any transform
	if (!this.isOpen)
		return new Point((int)point.X, (int)point.Y);
		
	if(!this.IsInsideOriginalRect(point) && IsInsideAdjustedRect(point))
		return new Point((int)point.X + this.xOffset, (int)point.Y + this.yOffset);
		
	return new Point((int)point.X, (int)point.Y);
}
```

- Popup is closed, so we don't need to do anything
- Popup is open, though we didn't change it's position, so no need to do anything
- Check if we need to perform a translation, not required when no over the popup

The logic can probably be cleaned up a little still. Does that make more sense now?

DidiDerDenker: Yes, thank you, I have understood that now 😄
I committed the changes.

DidiDerDenker:
@amaitland How do we continue from here? I updated this branch and all comments are solved as far as I know.

amaitland:
The changes need testing. As there's currently no unit tests for the new class, it may be a take some time to validate this.

You can see if anyone subscribed to #2820 is prepared to test and report back.

amaitland:
As nobody has volunteered to test this we can look at including it as an opt in feature. 
- Extract interface
- Add two implementations one that does nothing and one that performs the transform.
- Allow for opting in to using the new implementation.

DidiDerDenker:
> As nobody has volunteered to test this we can look at including it as an opt in feature.
> 
> * Extract interface
> * Add two implementations one that does nothing and one that performs the transform.
> * Allow for opting in to using the new implementation.

@amaitland Thank you! I extraced an interface and added an implementation that doesn't adjust anything. I'm not sure how to implement the "opt-in". I added prototypical code in the constructor of "ChromiumWebBrowser". Could you give me an advice on where and how to implement the "opt-in" in the best way? Thanks :)

amaitland:
Thanks for all the hard work. This has been merged, with follow up changes from PR #4759

You can enable the transform via an extension method:

```
using CefSharp.Wpf.Experimental;

chromiumWebBrowser.UsePopupMouseTransform();
```

Once this has been tested sufficently we can remove the method and enable it by default.

DidiDerDenker:
@amaitland Great, thank you! Thanks for your help as well. In which version is it going to be released?

amaitland:
Next major version (`M123`).

Hopefully released in about a week.

PR #4682, opened 2024-01-16
fix: actually support vswhere not finding something

mitchcapper:
By default vswhere returns a non-zero error code making our error handling fail

I don't remember if I am mis-remembering a line suppress error option in powershell but I couldn't find it, and I didn't want to try/catch this.


amaitland:
Thanks! Let me know if you need this merged into the 120 branch.

Issue #4621, opened 2023-11-01
Error while fulfilling of JS: Request BrowserId : 1 not found it's likely the browser is already closed


maksimrudn:
### Is there an existing issue for this?

- [X] I have searched both open/closed issues, no issue already exists.

### CefSharp Version

117.2.40

### Operating System

Windows 10

### Architecture

x64

### .Net Version

.net 6.0

### Implementation

OffScreen

### Reproduction Steps

Version: Above 108

Steps:
Create instance of CefSharp.OffScreen.ChromiumWebBrowser
br = CefSharp.OffScreen.ChromiumWebBrowser();

Go by url 2 times:
br.Load(url)
await br.EvaluateScriptAsPromiseAsync(script);
br.Load(url)
await br.GetSourceAsync();
br.Load(url)

Fulfill JS
_browser.EvaluateScriptAsPromiseAsync(script)

### Expected behavior

Fulfilling of JS

### Actual behavior

JS Error Request BrowserId : 1 not found it's likely the browser is already closed

### Regression?

After regression to version 108 problem dissapeared.

### Known Workarounds

Likely problem is here
https://github.com/cefsharp/CefSharp/blob/master/CefSharp.BrowserSubprocess.Core/CefAppUnmanagedWrapper.cpp
Look at history..

### Does this problem also occur in the CEF Sample Application

Yes using WPF/OffScreen command line args

amaitland:
> ### Reproduction Steps
> Version: Above 108

Please create a Unit test that reliably reproduces the problem and submit a `PR`.

https://github.com/cefsharp/CefSharp/tree/master/CefSharp.Test

Your code is incomplete and not enough for me to investigate this.

> Likely problem is here
> https://github.com/cefsharp/CefSharp/blob/master/CefSharp.BrowserSubprocess.Core/CefAppUnmanagedWrapper.cpp
> Look at history..

What is the problem **exactly**? Please provide details of your analysis.

Just saying look at history is meaningless. If you can see a specific error that needs fixing then submit a `PR`.

Likely this is the result of `upstream` changes to `CEF/Chromium`.

maksimrudn:
> > ### Reproduction Steps
> > Version: Above 108
> 
> Please create a Unit test that reliably reproduces the problem and submit a `PR`.
> 
> https://github.com/cefsharp/CefSharp/tree/master/CefSharp.Test
> 
> Your code is incomplete and not enough for me to investigate this.
> 
> > Likely problem is here
[...]

PR #4741, opened 2024-03-03
Add Test to replicate issue #4621

nebojsasandin:
Modified Google test to replicate this issue: https://github.com/cefsharp/CefSharp/issues/4621

### DIFF
--- a/CefSharp.BrowserSubprocess.Core/BindObjectAsyncHandler.h
+++ b/CefSharp.BrowserSubprocess.Core/BindObjectAsyncHandler.h
@@ -151,14 +151,14 @@ namespace CefSharp
                                     auto rootObjectWrappers = _browserWrapper->JavascriptRootObjectWrappers;
 
                                     JavascriptRootObjectWrapper^ rootObject;
-                                    if (!rootObjectWrappers->TryGetValue(frame->GetIdentifier(), rootObject))
+                                    if (!rootObjectWrappers->TryGetValue(StringUtils::ToClr(frame->GetIdentifier()), rootObject))
                                     {
 #ifdef NETCOREAPP
                                         rootObject = gcnew JavascriptRootObjectWrapper(browser->GetIdentifier());
 #else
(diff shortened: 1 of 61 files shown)


amaitland:
- Please fix the formatting (you've reformatted the whole file). 
- Please create a new test for the issue rather than rewriting an existing one (we still need to keep the existing one).

> [Add files via upload](https://github.com/cefsharp/CefSharp/pull/4741/commits/aa9497109a6cd82083cc55805cb15fe3e3accd8d)

It appears you've uploaded the file which makes me wonder if you've actually run the test?

nebojsasandin:
@amaitland 
Is it better now?
- Previously you've given me a direct line to a Google test method, made me think you wanted me to include this Google-related issue into that test. Forgive my foolishness.
- I have created a new test, and ran it, it still produces the same issue previously stated.
- Originally I could not run any tests on my end, when I compiled the code I got a whole bunch of C++ errors I'm unfamiliar with, and I could not work out how to make it run, so I just changed the test I needed to. Then I got error 403 - Forbidden when I tried to commit from my computer, so I simply uploaded the file via git web. I did not think it would mess up the entire file formatting. Sorry about that. Desktop git is working now, and the entire source is compiling properly, so I was able to do both: run the test locally, and commit changes properly.
- Let me know if there's anything else you need from me.

amaitland:
Thanks. I'll have a look closer on the weekend.

amaitland:
The issue does indeed look like an `upstream` one. If you isolate the two origins in different render processes then it should work as expected.

```c#
settings.CefCommandLineArgs.Add("isolate-origins", "https://www.google.com,https://accounts.google.com/");
```

amaitland:
I've manually merged this test in commit https://github.com/cefsharp/CefSharp/commit/800cbcf0fd58d85bfa0c3f57636ddfa823a1b830

It's now in a separate file that references the original issue.

Thanks for the unit tests 👍

PR #4771, opened 2024-03-27
Further demonstration of issue: https://github.com/cefsharp/CefSharp/issues/4621

nebojsasandin:
Previously, it was concluded that the bypass for issue: https://github.com/cefsharp/CefSharp/issues/4621 can be resolved by using:

`settings.CefCommandLineArgs.Add("isolate-origins", "https://accounts.google.com,https://chrome.google.com,https://chromewebstore.google.com,https://mail.google.com,https://www.google.com,https://google.com");`

...however, there's a new test which demonstrates that this workaround isn't always helpful.

GoogleSearchToGoogleAccountsBreaksJS() test method works with the workaround method, however: GoogleSearchToGmailBreaksJS() test method does not work.


amaitland:
Test added in https://github.com/cefsharp/CefSharp/commit/527a062924925e686ef7f6c816be09288764a633

They'll only run locally for now.

See https://github.com/cefsharp/CefSharp/issues/4621

PR #4796, opened 2024-05-07
Implement AcceleratedPaintInfo

Scalibq:
Implement AcceleratedPaintInfo so the shared texture can be used for accelerated rendering in a custom IRenderHandler.
Code was successfully tested against Direct3D11.1 using SharpDX on Windows 11.
Use SharpDX.Direct3D11.Device1.OpenSharedResource1< Texture2D >(AcceleratedPaintInfo.SharedTextureHandle);
Then copy the contents of this resource to your own Texture2D before returning from OnAcceleratedPaint().

### DIFF
--- a/CefSharp/AcceleratedPaintInfo.cs
+++ b/CefSharp/AcceleratedPaintInfo.cs
@@ -8,11 +8,30 @@ using CefSharp.Enums;
 namespace CefSharp
 {
     /// <summary>
-    /// AcceleratedPaintInfo
+    /// Class representing accelerated paint info.
     /// </summary>
     public sealed class AcceleratedPaintInfo
     {
-        public IntPtr SharedTextureHandle;
-        public ColorType Format;
+        /// <summary>
+        /// Handle for the shared texture. The shared texture is instantiated
+        /// without a keyed mutex.
+        /// </summary>
+        public IntPtr SharedTextureHandle { get; }
+
+        /// <summary>
+        /// The pixel format of the texture.
+        /// </summary>
+        public ColorType Format { get; }
+
+        /// <summary>
+        /// AcceleratedPaintInfo
+        /// </summary>
+        /// <param name="sharedTextureHandle">Handle to the shared texture resource</param>
+        /// <param name="format">The pixel format of the shared texture resource</param>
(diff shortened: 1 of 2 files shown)


CefSharp/AcceleratedPaintInfo.cs
```diff
@@ -2,12 +2,17 @@
...
+using CefSharp.Enums;
+
 namespace CefSharp
 {
     /// <summary>
     /// AcceleratedPaintInfo
     /// </summary>
     public sealed class AcceleratedPaintInfo
     {
+        public IntPtr SharedTextureHandle;
```
amaitland, line 15: These should be immutable properties

Please add a `get` and set via constructor. for both properties

Also please include the appropriate xml doc based on the `CEF API`.

CefSharp/AcceleratedPaintInfo.cs
```diff
@@ -2,12 +2,17 @@
 //
 // Use of this source code is governed by a BSD-style license that can be found in the LICENSE file.
 
+using System;
+using CefSharp.Enums;
+
 namespace CefSharp
 {
     /// <summary>
     /// AcceleratedPaintInfo
```
amaitland, line 11: Please update to match the `CEF API`.

```
Class containing shared texture information for the OnAcceleratedPaint
callback. Resources will be released to the underlying pool for reuse when
the callback returns from client code.
```

amaitland (changes requested):
Thanks for the `PR`.

Comments inline.

CefSharp/AcceleratedPaintInfo.cs
Scalibq: The underlying CEF code takes care of the resource management.
The callback only passes a handle to the resource.
When using this callback, you just need to know that you cannot use the handle after returning. That's what this documentation says.
As I already said above: make a copy of the resource inside the callback, if you intend to use it outside of the callback.

amaitland: Please add the relevant `xml doc` and change the public fields (the project doesn't use public fields unless they're static) to properties set via the constructor.

Thanks!

amaitland (approved):

Issue #4801, opened 2024-05-13
WPF - JavaScript DragEnd event not emitted


WilliamTomOBrien:
### Is there an existing issue for this?

- [X] I have searched both open/closed issues, no issue already exists.

### CefSharp Version

124.3.50.0

### Operating System

Windows 11

### Architecture

x64

### .Net Version

4.7.2

### Implementation

WPF

### Reproduction Steps

1. Run CefSharp.Wpf.Example
2. Load the [following gist](https://gist.github.com/WilliamTomOBrien/5d6a457062302dd3b9df8d722e1fa94c) in the browser
3. Drag the blue box from one grey box to the other
4. View the emit drag and drop events listed below

### Expected behavior

I would expect the `dragend` event to be emit when the drag operation ended, as described [here](https://developer.mozilla.org/en-US/docs/Web/API/HTML_Drag_and_Drop_API#drag_events). In this example, this would be reflected by `dragend` being printed at the bottom of the event list.

This behavior does occur when using the CEF browser, as shown below.
![cef-124-drag-and-drop-events](https://github.com/cefsharp/CefSharp/assets/29744773/a2ec2b59-d64e-4ed9-ad01-be86abc35e95)

### Actual behavior

The `dragend` event is not emit after dragging and dropping an element. 

![cefsharp-124-drag-and-drop-events](https://github.com/cefsharp/CefSharp/assets/29744773/e57b2f0f-f44c-4a9f-9b2c-e6549b109bd0)

Apologies for printing the events to the screen - I wanted the gif showing the behavior in CEF to match, and unfortunately trying to load DevTools when running the `cefclient.exe --multi-threaded-message-loop --no-sandbox --off-screen-rendering-enabled --enable-gpu` command would crash my instance of CEF.

### Regression?

[This issue](https://github.com/cefsharp/CefSharp/issues/3941) seems like it may potentially be relevant, however was not followed up on by the creator. Unfortunately, I was unable to verify if this was the case, as I was unable to build the [cefsharp/86](https://github.com/cefsharp/CefSharp/tree/cefsharp/86) branch.

### Known Workarounds

No known workarounds, aside from avoiding any functionality that would rely on an event handler for `dragend`.

### Does this problem also occur in the CEF Sample Application

No

amaitland:
If `cefclient` with `--off-screen-rendering-enabled` works correctly then you can compare the behaviour and see what changes are required in the `WPF` implementation.

[...]

PR #4814, opened 2024-05-29
WPF - Ensure only single drag operation is emit on DragSourceEndedAt

WilliamTomOBrien:
**Fixes:** #4801 

**Summary:** 
   - This change ensures that only `none`, `move`, `copy`, or `link` can be passed the the browser host to signify a DragEnd event. Specifying multiple flags on the bitmask seems to prevent the event from firing, and therefore the result that is returned from DragDrop.DoDragDrop is compared to the last drag drop effect we have on record.
   - When doing further testing after creating the issue, I noticed that the dragEnd event *was* emit when I cancelled the drag drop. Beyond this, when testing and manually specifying the flag, I noticed the event was emit when using the `none`, `move`, `copy`, or `link` flags by themselves in the bitmask, and seemed to correlate with the event's `dataTransfer.dropEvent` field. Given the API specifies that can only be one of those four values (and not a mixture of [them](https://developer.mozilla.org/en-US/docs/Web/API/DataTransfer)), that seemed reasonable. Furthermore, it seemed that the `DoDragDrop` function used within the `StartDragging` function would consistently return the mask with more than one value set.
  
**Changes:** 
   - I have modified the implementation of the CefSharp.Wpf ChromiumWebBrowser's StartDragging method, but no significant changes were made to the structure of the code.
      
**How Has This Been Tested?**  

This has been tested by going through the same process detailed in the issue [here](https://github.com/cefsharp/CefSharp/issues/4801#issue-2293862100). This was done on Windows 11, with an x64 architecture, .NET Version 4.7.2, and the WPF implementation.

**Types of changes**

- [ ] Updated documentation

**Checklist:**

- [X] Commented my code
- [ ] Changed the documentation(if applicable) N/A
- [X] New files have a license disclaimer

### DIFF
--- a/CefSharp.Wpf/ChromiumWebBrowser.cs
+++ b/CefSharp.Wpf/ChromiumWebBrowser.cs
@@ -920,7 +920,7 @@ namespace CefSharp.Wpf
             {
                 if (browser != null)
                 {
-                    //DoDragDrop will fire DragEnter event
+                    // DoDragDrop will fire DragEnter event
                     var result = DragDrop.DoDragDrop(this, dataObject, allowedOps.GetDragEffects());
                     var dragEffects = currentDragDropEffects.HasValue ? currentDragDropEffects.Value : DragDropEffects.None;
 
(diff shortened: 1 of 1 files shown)


amaitland:
Thanks for the PR 👍 

We should compare this to the cefclient reference implementation.

amaitland:
Some quick testing and this look ok 👍 It probably needs more thorough testing though.

Rather than let it sit here I'll extract the logic out into a method and allow for customisation of the behaviour.

Someone can implement their own custom logic (or revert to the previous behaviour if they wanted).

Follow up PR #4845

PR #4845, opened 2024-06-23
WPF - Add GetDragOperationsMask virtual method to allow custom behaviour

amaitland:
Extract behaviour into method so user can override if required.

Follow up to PR# 4814


amaitland:
The only thing I don't like is the name being the same as the extension method. I don't have a better name, so going with that for now.

Issue #4867, opened 2024-07-06
Change Cef.IsInitialized from bool to bool?


amaitland:
Change to nullable to we can track the not initialized state, and true/false.

We can then provide better error messages for the Chromium process singleton #4668

PR #4869, opened 2024-07-06
Core - Change Cef.IsInitialized from bool to bool?

amaitland:
**Fixes:**
#4867

**Summary:**
Change to  nullable bool

- change type
- Throw exception if false
      
**How Has This Been Tested?**  
Needs more testing

**Types of changes**

- [ ] Updated documentation

**Checklist:**

- [ ] Commented my code


CefSharp.Core.Runtime/Cef.h
```diff
@@ -555,11 +555,11 @@ namespace CefSharp
             /// </summary>
             static void Shutdown()
             {
-                if (_initialized)
```
amaitland, line 558: If `_initialized` is null potentially we should just return

CefSharp.Core.Runtime/Cef.h
```diff
@@ -615,11 +615,11 @@ namespace CefSharp
             /// </summary>
             static void ShutdownWithoutChecks()
             {
-                if (_initialized)
+                if (_initialized.GetValueOrDefault())
```
amaitland, line 618: Same as above

amaitland:
I've asked https://www.magpcss.org/ceforum/viewtopic.php?f=6&t=19918 to see if it's possible to call `CefInitialize` twice if the first time fails due to singleton.

Assuming the answer will be `no`. If it's `yes` then we can change some of the checks later.

PR #4900, opened 2024-08-05
Example program for rendering with D3D using shared textures and OnAcceleratedPaint()

Scalibq:
**Summary:**
A simple proof-of-concept program that hosts an offline CefSharp instance, and initializes it to render via shared D3D11 textures.
It can run in either fullscreen or windowed mode (switch with ALT+ENTER), with or without VSYNC, and with or without external frame trigger.
These are controlled by the FULLSCREEN, VSYNC and EXTERNALTRIGGER consts, found at the top of DXForm.cs.

**Changes:**
- Added a new project to the netcore solution: CefSharp.Direct3D.Example.netcore
      
**How Has This Been Tested?**  
The program itself is a test application.
It's meant to act as a simple example of how to use the accelerated interfaces, and can also be modified to test and verify new CEF features or alternative implementations.

**Types of changes**

- [ ] Updated documentation

**Checklist:**

- [ ] Commented my code


Scalibq:
Note that there appears to be an issue with performance of the CEF-rendering when vsync is enabled.
To the best of my knowledge, the code follows all guidelines of the CEF-developers by waiting on the copy-operation to complete via a Query.
It appears that CEF's internal rendering gets confused when handling of OnAcceleratedPaint() takes longer.
The CEF developers however deny any such issue exists, yet do not provide working example code.

amaitland:
Thanks for the `PR`. Hopefully others will find it useful.

I'm not going to merge this into `master` as it would then need to be maintained as part of the normal repo, I myself don't have any experience with `D3D`.

I've created https://github.com/cefsharp/OnAcceleratedPaint.Example if you'd like to create a `PR` against that which references the `Nuget` pacakges then that would be welcomed.

> The CEF developers however deny any such issue exists, yet do not provide working example code.

If your referring to https://github.com/chromiumembedded/cef/issues/3755 then I'd strongly suggest you take a different approach to asking for help.

Scalibq:
I had made a perfectly clear and polite issue, but they were immediately attacking my level of knowledge and experience (which is preposterous, as I have decades of experience with D3D under my belt, and probably know a lot more about it than they do).
They have the attitude problem, I have been quite lenient with them, and tried to explain that I had already tried what they suggested, but it didn't work, yet they kept insisting that it does.
This code proves that it doesn't. Their claim that CEF is 100% perfect is wrong (and is a preposterous claim to make for any developer).

Issue #4958, opened 2024-10-20
SerializeV8SimpleObject does not support ExpandoObject


mitchcapper:
### Is there an existing issue for this?

- [X] I have searched both open/closed issues, no issue already exists.

### CefSharp Version

129

### Operating System

Windows 10

### Architecture

x64

### .Net Version

.net 8

### Implementation

WPF

### Reproduction Steps

Per https://github.com/cefsharp/CefSharp/discussions/4937 I found the 'simple':
```
var curExtSettings = (Dictionary<string,object>) context.GetPreference("extensions.settings");
context.SetPreference("extensions.settings", curExtSettings, out errSet);
```
fails.  The cefclient shows us it should be possible.  Digging into it the serialization used for SetPreference only handles generic IDictionaries but not typed ones like expandoObject.

### Expected behavior

getting and setting the settings works

### Actual behavior

settings store is corrupted visiting extension page will crash,  corrupt json will be written to disk,  restarting however does seem to do some cleanup.

### Regression?

no

### Known Workarounds

Force an object containing expandoobjects to be converted prior to being sent for serialization

### Does this problem also occur in the CEF Sample Application

No

### Other information

will submit PR #4959

PR #4959, opened 2024-10-20
SerializeV8SimpleObject support for ExpandoObjects

mitchcapper:
**Fixes:** #4958

**Summary:**
   This adds ExpandoObject support to the SerializeV8SimpleObject path.   We could add support for testing if it is assignable to a idictionary<string,object> which might allow a few more things to work but im not aware off hand what they would be.  Maybe we could add support for any templated IDictionary but that is beyond my CLR c++ knowledge. 
We could eliminate some code potentially merging this and the dictionary function into one.  If the item is an expandoObject we could use the IDictionary constructor of Dictionary to create a Dictionary<string,object> and then use the rest of the IDictionary support that is there.   Let me know if you want me to revise it to that.

**Changes:**
Added specific case for ExpandoObject so should have minimal impact
      
**How Has This Been Tested?**  
Ran it once, cow was still standing after, clearly bulletproof.

image

**Types of changes**
Default behavior created nodes with literal "Key": <expandoItemKey>, "Value":<expandoItemValue>, i doubt something replied on that.

**Checklist:**

- [x] Commented my code
- [ x] Changed the documentation(if applicable)
- [ x] New files have a license disclaimer
- [x ] The formatting is consistent with the project (project supports .editorconfig)

### DIFF
--- a/CefSharp.Core.Runtime/Internals/Serialization/V8Serialization.cpp
+++ b/CefSharp.Core.Runtime/Internals/Serialization/V8Serialization.cpp
@@ -140,6 +140,18 @@ namespace CefSharp
                         list->SetString(index, StringUtils::ToNative(Convert::ToString(obj)));
                     }
                 }
+                // ExpandoObject only implements IDictionary<string,object> not the non-generic System.Collection.IDictionary like Dictionary does
+                else if (type == System::Dynamic::ExpandoObject::typeid)
+                {
+                    auto subDict = CefDictionaryValue::Create();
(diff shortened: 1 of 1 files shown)


amaitland:
Nice one 😀 thanks!

Given the complexity I think handling the case individually is fine.

I'll try add a test on the weekend.

amaitland:
Basic test added in commit https://github.com/cefsharp/CefSharp/commit/c3339bd418b86a60f3c00b622510df7e88db999f

PR #5052, opened 2025-02-12
FolderSchemeHandlerFactory Handle URL encoded paths only in absolute path

jbohunovsky-IDG:
Related to pull request https://github.com/cefsharp/CefSharp/pull/2364

**Summary:**
The original fix can potentially lead to non-existing root path. This PR fixes the issue when the `rootFolder` has "escaped" URL characters even thought that's how the folder is named (e.g. `hello%2Fworld` folder would try to lookup `hello/world` folder which doesn't exist).

**Changes:**
- Moved `WebUtility.UrlDecode` from the full path to absolute path only.
      
**How Has This Been Tested?**  
1. I've used CefSharp.Wpf.Example.netcore project, where I've added new custom scheme with `FolderSchemeHandlerFactory`, with the following parameters:
   - rootFolder: `"escaped%2Fslash"`
   - schemeName: `"app"`
   - hostName: `"test"`
2. Then I've created folder `escaped%2Fslash` in the output directory and created two simple HTML files, one called `index.html` and the other `file with spaces.html`.
3. Run the project and navigate to `app://test/index.html` and `app://test/file%20with%20spaces.html`. Both files have been loaded correctly.

**Types of changes**
- [ ] Updated documentation

**Checklist:**
- [ ] Commented my code

### DIFF
--- a/CefSharp/SchemeHandler/FolderSchemeHandlerFactory.cs
+++ b/CefSharp/SchemeHandler/FolderSchemeHandlerFactory.cs
@@ -109,7 +109,7 @@ namespace CefSharp.SchemeHandler
                 asbolutePath = defaultPage;
             }
 
-            var filePath = WebUtility.UrlDecode(Path.GetFullPath(Path.Combine(rootFolder, asbolutePath)));
+            var filePath = Path.GetFullPath(Path.Combine(rootFolder, WebUtility.UrlDecode(asbolutePath)));
 
             //Check the file requested is within the specified path and that the file exists
             if (filePath.StartsWith(rootFolder, StringComparison.OrdinalIgnoreCase) && File.Exists(filePath))


amaitland:
Thanks for the PR!

Would this be a breaking change? Would there be an impact on upgrading users?

jbohunovsky-IDG:
I don't believe this is a breaking change, as the absolute path (URL address path) is untouched and the root folder (defined by the developer) already has a check that it exists, so if someone had URL encoded root folder, they would get an error everytime upon accessing any path within that folder (as decoding that path would lead to a different path than which was defined by the developer), which is what this PR is fixing.

amaitland:
> so if someone had URL encoded root folder, they would get an error everytime upon accessing any path within that folder

Is that not a breaking change? You'd only find this out by running the application and accessing the paths.

This will need to be documented so upgrading users are at least aware of the change.

jbohunovsky-IDG:
That's the current behaviour. If you set the root folder to `example%20folder`, it will pass the `Directory.Exists` check in the constructor, but then in the Create method the full path will be URL decoded, effectively changing the root folder to `example folder` which doesn't exist.

Users will not be affected if they don't use URL encoded root folder, but if they do, it's not working properly for them right now anyways.

PR #5073, opened 2025-03-23
Fixed duplicate calls to OnContextCreated

AgapovAlexsey:
Related issue in CEF:  [3867](https://github.com/chromiumembedded/cef/issues/3867)

**Summary:**
   - Removed multiple calls to IRenderProcessMessageHandler.OnContextCreated

**Types of changes**

- [ ] Updated documentation

**Checklist:**

- [ ] Commented my code

### DIFF
--- a/CefSharp.BrowserSubprocess.Core/CefAppUnmanagedWrapper.cpp
+++ b/CefSharp.BrowserSubprocess.Core/CefAppUnmanagedWrapper.cpp
@@ -131,6 +131,12 @@ namespace CefSharp
                 _handler->OnContextCreated(% browserWrapper, % frameWrapper, % contextWrapper);
             }
 
+            //Skip additional contexts (DevTools, extensions) to avoid
+            //double binding and duplicate calls to IRenderProcessMessageHandler.OnContextCreated
+            //https://github.com/chromiumembedded/cef/issues/3867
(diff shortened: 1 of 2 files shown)


amaitland:
Thanks for the `PR` 👍 

Please rebase on `master`.

amaitland:
Thanks 👍 Will cherry pick into 134 branch once builds are passing.

AgapovAlexsey:
@amaitland 
Should I add this fix to OnContextReleased as well? It seems like OnContextCreated and OnContextReleased should have consistent behavior.

amaitland:
If `OnContextReleased` is called for non main world context then yes we should make it consistent.

AgapovAlexsey:
@amaitland I checked it. OnContextReleased is called multiple times as well. But we can't use the same code to check it — frame->GetV8Context() will cause an access violation error. I also tried comparing frame IDs, but context->GetFrame() always returns an empty frame.

It looks like we’ll be able to do it after CEF adds word_id to the context.

amaitland:
> But we can't use the same code to check it — `frame->GetV8Context()` will cause an access violation error. I also tried comparing frame IDs, but `context->GetFrame()` always returns an empty frame.

Disappointing, but not really surprising. Thanks for taking the time to check 👍

PR #5091, opened 2025-04-27
JSB - Forward binding result to BindObjectAsync

AgapovAlexsey:
**Summary:**
   - Replaced kPromiseCreatorScript with native promise `CefV8Value::CreatePromise()`
   - Get binding reslut: `const result = await cefSharp.bindObjectAsync('boundAsync2');  const boundAsync2 = result.bindResult.boundAsync2;`

**Types of changes**

- [ ] Updated documentation

**Checklist:**

- [ ] Commented my code

## Summary by CodeRabbit

- **New Features**
  - Added enhanced validation in tests to ensure bind results include both lowercase and capitalized keys: count, message, success, and bindResult.

- **Refactor**
  - Simplified promise handling for asynchronous JavaScript method callbacks, using a unified promise object.
  - Streamlined the creation and binding of JavaScript objects, now returning V8 values directly from conversion methods.
  - Consolidated response object creation for binding operations, improving consistency and maintainability.

- **Style**
  - Updated test files to use single-quoted string literals throughout.

- **Tests**
  - Added a new test to verify the structure of bind results and improved assertions in existing tests.

### DIFF
--- a/CefSharp.Example/Resources/BindingTestAsync.js
+++ b/CefSharp.Example/Resources/BindingTestAsync.js
@@ -2,197 +2,197 @@ QUnit.module('BindingTestAsync', (hooks) =>
 {
     hooks.before(async () =>
     {
-        await CefSharp.BindObjectAsync("boundAsync");
+        await CefSharp.BindObjectAsync('boundAsync');
     });
 
-    QUnit.test("BindObjectAsync Second call with boundAsync param", async (assert) =>
+    QUnit.test('BindObjectAsync Second call with boundAsync param', async (assert) =>
     {
-        const res = await CefSharp.BindObjectAsync("boundAsync");
-        assert.equal(res.Success, false, "Second call to BindObjectAsync with already bound objects as params returned false.");
+        const res = await CefSharp.BindObjectAsync('boundAsync');
+        assert.equal(res.Success, false, 'Second call to BindObjectAsync with already bound objects as params returned false.');
     });
 
-    QUnit.test("Async call (Throw .Net Exception)", async (assert) =>
+    QUnit.test('Async call (Throw .Net Exception)', async (assert) =>
     {
         assert.rejects(boundAsync.error());
     });
 
-    QUnit.test("Async call (Divide 16 / 2):", async (assert) =>
+    QUnit.test('Async call (Divide 16 / 2):', async (assert) =>
     {
         const actualResult = await boundAsync.div(16, 2);
         const expectedResult = 8;
-        assert.equal(expectedResult, actualResult, "Divide 16 / 2");
+        assert.equal(expectedResult, actualResult, 'Divide 16 / 2');
     });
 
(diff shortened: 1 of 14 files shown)


CefSharp.Example/Resources/BindingTestAsync.js
```diff
@@ -2,209 +2,209 @@ QUnit.module('BindingTestAsync', (hooks) =>
 {
     hooks.before(async () =>
     {
-        await CefSharp.BindObjectAsync("boundAsync");
+        await CefSharp.BindObjectAsync('boundAsync');
```
amaitland, line 5: Please revert the styling change, it creates too much noise and doesn't actually provide a functional change.

Makes it very hard to review what's actually changed.

amaitland:
Firstly, thanks for the `PR` 👍 Will be a few days before I can review in detail.

> * Get binding reslut: `const result = await cefSharp.bindObjectAsync('boundAsync2');  const boundAsync2 = result.bindResult.boundAsync2;`

What's the use case for this?

CefSharp.Example/Resources/BindingTestAsync.js
AgapovAlexsey: Sorry. I did it out of habit.

AgapovAlexsey:
> What's the use case for this?

I don't have any special case. It's a request from colleagues. Because it seems more logical: you call bindObjectAsync and get the object you need, instead of looking for it somewhere in the window.

amaitland:
> Because it seems more logical: you call bindObjectAsync and get the object you need, instead of looking for it somewhere in the window.

I think a new method that creates a instance of the object without adding to the global scope would be preferable.

amaitland:
> I believe introducing a new method is unnecessary.

This is my preference, I believe this will make for a better user experience. Much easier to document and provide examples for.

> Instead, we can add a parameter to bindObjectAsync (similar to NotifyIfAlreadyBound).

For me a new `API` with an explicit behavior would be preferable. Extra parameters with different behavior I don't believe is desirable.

> if we decide to completely drop binding to window, minimal code changes will be required.

Unless `Chromium` makes changes that make this necessary I have no plans to do this. This would be an unnecessary breaking change.

amaitland:
> * Get binding reslut: `const result = await cefSharp.bindObjectAsync('boundAsync2');  const boundAsync2 = result.bindResult.boundAsync2;`

I'm not a fan of adding `bindResult` to the response object. I'd much prefer to see a new method added that explicitly creates a new object 

```js
const boundAsync2 = await cefSharp.createObjectAsync('boundAsync2'); 
```

I think doing that in a separate `PR` makes sense as there are some considerations regarding caching.

> * Replaced kPromiseCreatorScript with native promise `CefV8Value::CreatePromise()`

This is an excellent improvement and I'll cherry pick this change out of this `PR`. I'll try do this on the weekend.

PR #5120, opened 2025-06-16
Handle case where IPostDataElement.Bytes is null

Matheos96:
   - We noticed that we had to check the IPostDataElement.Bytes array for null ourselves before calling .GetBody(). This felt a bit weird, so I think it makes sense to handle this in CefSharp code like I did here.

   - Added null check for Bytes of IPostDataElement object
      
**How Has This Been Tested?**  
By calling .GetBody() on a IPostDataElement object with PostDataType.Empty and Bytes = null.

**Types of changes**

- [ ] Updated documentation

**Checklist:**

- [ ] Commented my code
- [x]  The formatting is consistent with the project (project supports .editorconfig)


amaitland:
Nice one! Thanks

Issue #4515, opened 2023-06-03
Ship d3dcompiler_47.dll on ARM64


amaitland:
https://github.com/chromiumembedded/cef/issues/3515

kpreisser:
I think this can be fixed by just reverting the changes from #3841, so that `d3dcompiler_47.dll` is no longer treated specially in the `DependencyChecker`.

amaitland:
Confirming the `nuget` package contains `d3dcompiler_47.dll`.

PR for the proposed change would be most welcome 👍

PR #5246, opened 2026-05-19
Update to support ARM64 build target

jamescharters:
**Fixes:** [issue-number] 

This change introduces full ARM64 support for CefSharp, covering native C++/CLI projects, managed .NET Framework projects, NuGet packaging, and the build pipeline. It enables CefSharp to be built and distributed for Windows ARM64 devices natively.

- Added ArchitectureHelper.cs — new helper class that resolves the current process architecture (x86/x64/arm64) for both .NET Framework and .NET Core
- Modified CefRuntime.cs — replaced Environment.Is64BitProcess checks with ArchitectureHelper.ProcessArchitecture so ARM64 processes resolve the correct arch-specific assembly path
- Modified DependencyChecker.cs — skips d3dcompiler_47.dll in optional dependency checks when running on ARM64 (not available on that platform)
- Modified CefSharp.Core.Runtime.vcxproj and CefSharp.BrowserSubprocess.Core.vcxproj — added Debug|arm64 and Release|arm64 project configurations
- Modified CefSharp.BrowserSubprocess.csproj — added arm64 platform/runtime identifier and updated the post-build editbin/sn script to select the correct vcvarsall arch for x86, x64, and arm64 (including cross-compile from AMD64 host)
- Modified CefSharp3.sln — registered Debug|arm64 and Release|arm64 solution-level configurations for all projects
- Modified CefSharp.Native.props — added arm64 locale content item group
- Modified example and test projects (OffScreen, WinForms, WPF, HwndHost, Test) — added arm64 to <Platforms>, <RuntimeIdentifiers>, and conditional <PlatformTarget>/<RuntimeIdentifier> properties
- Modified CefSharp.Common.nuspec — added chromiumembeddedframework.runtime.win-arm64 dependency and arm64 binary entries
- Modified CefSharp.Common.props — added CefSharpCommonBinariesArm64 and CefSharpBrowserProcessCoreArm64 item groups
- Modified CefSharp.Common.targets — added arm64 <None>/<Content> item groups and CefSharpCopyFilesAfterNugetRestoreArm64 copy target
- Modified build.ps1 — refactored BuildSolution to invoke msbuild.exe directly; added Get-HostArchitecture, Get-VCVarsAllArch, and Ensure-BuildEnvironment helpers to correctly bootstrap the VC environment per target arch (native or cross-compile); removed the blanket arm64 exclusion for NetFramework builds and replaced it with a guard that requires VS2022/v143 when targeting arm64

**How Has This Been Tested?**  

Built the solution targeting ARM64 using build.ps1 with -BuildArches arm64 -Target vs2022 on an AMD64 host (cross-compile). Verified arm64 binaries are produced in bin\arm64\Release for both CefSharp.Core.Runtime and CefSharp.BrowserSubprocess.

**Types of changes**

- [ ] Updated documentation

**Checklist:**

- [ ] Commented my code


build.ps1
```diff
@@ -116,38 +116,83 @@ function BuildSolution
         "/verbosity:normal"
     )
 
-    $StartInfo = New-Object System.Diagnostics.ProcessStartInfo
-    $StartInfo.FileName = "msbuild.exe"
-    $StartInfo.Arguments = $Arguments
```
amaitland, line 121: Is this strictly necessary? From memory there was a specific reason the EnvironmentVariables were cleared (I don't remember the specifics, it's been years since this was done).

Please revert any unnecessary changes to this file. Keeping it to the bare minimum change will also make it easy to review.

jamescharters:
@amaitland Hi Alex, this is currently a draft of changes to get this building for ARM64. My team has compiled this locally and is using it on real ARM64 hardware (M4 Macs) without issue. Was wanting to know how to move forward with this, as it looks like the CI/CD environment does not support this architecture.

CefSharp/DependencyChecker.cs
```diff
@@ -172,6 +172,19 @@ public static List<string> CheckDependencies(bool checkOptional, bool packLoadin
...
+            foreach (var dependency in CefOptionalDependencies)
+            {
+                if (ArchitectureHelper.IsArm64Process && string.Equals(dependency, "d3dcompiler_47.dll", StringComparison.OrdinalIgnoreCase))
+                {
+                    continue;
+                }
+
+                yield return dependency;
+            }
+        }
```
amaitland: Chromium started shipping the d33compiler some time back.

https://github.com/cefsharp/CefSharp/issues/4515

unless they've stopped shipping the dll this can be reverted.

CefSharp/ArchitectureHelper.cs
```diff
@@ -0,0 +1,47 @@
...
+        {
+            return MapArchitecture(RuntimeInformation.ProcessArchitecture.ToString());
+        }
+
+        private static string MapArchitecture(string architecture)
+        {
+            switch (architecture)
+            {
+                case "Arm64":
+                case "arm64":
```
amaitland: Could we not just turn the string to lowercase?

relying on hard coded case comparison seems less than ideal.

If we do need a comparison then using if statements with string comparison to ignore case I think makes more sense then multiple case statements.

CefSharp/CefSharp.csproj
```diff
@@ -22,6 +22,7 @@
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" />
+    <PackageReference Include="System.Runtime.InteropServices.RuntimeInformation" Version="4.3.0" />
```
amaitland, line 25: If we need this then it will need to be added to the nuspec file.

amaitland (changes requested):
See comments inline

amaitland:
> Was wanting to know how to move forward with this

Until it builds there really isn't much I can do. You'd need to build locally in the mean time.

As per my comment inline try reverting your changes to the build script to see if that helps.

CefSharp/CefSharp.csproj
jamescharters: It is used in ArchitectureHelper.cs, and since the project targets net462 we need this to backport the API (it was added to the BCL in net471)

CefSharp.BrowserSubprocess/CefSharp.BrowserSubprocess.csproj
```diff
@@ -41,6 +42,6 @@
...
...EnvDir)..\..\VC\Auxiliary\Build\vcvars32.bat&quot;&#xD;&#xA;  )&#xD;&#xA;  editbin /largeaddressaware /TSAWARE &quot;$(TargetPath)&quot;  &#xD;&#xA;  sn -R &quot;$(TargetPath)&quot; &quot;$(ProjectDir)..\CefSharp.snk&quot;&#xD;&#xA;) else if $(PlatformName) == x64 (&#xD;&#xA;  if /I &quot;%PROCESSOR_ARCHITECTURE%&quot; == &quot;ARM64&quot; (&#xD;&#xA;    call &quot;$(DevEnvDir)..\..\VC\Auxiliary\Build\vcvarsall.bat&quot; arm64_x64&#xD;&#xA;  ) else (&#xD;&#xA;    call &quot;$(DevEnvDir)..\..\VC\Auxiliary\Build\vcvars64.bat&quot;&#xD;&#xA;  )&#xD;&#xA;  editbin /TSAWARE /STACK:8388608 &quot;$(TargetPath)&quot;&#xD;&#xA;  sn -R &quot;$(TargetPath)&quot; &quot;$(ProjectDir)..\CefSharp.snk&quot;&#xD;&#xA;) else if $(PlatformName) == arm64 (&#xD;&#xA;  if /I &quot;%PROCESSOR_ARCHITECTURE%&quot; == &quot;ARM64&quot; (&#xD;&#xA;    call &quot;$(DevEnvDir)..\..\VC\Auxiliary\Build\vcvarsall.bat&quot; arm64&#xD;&#xA;  ) else (&#xD;&#xA;    call &quot;$(DevEnvDir)..\..\VC\Auxiliary\Build\vcvarsamd64_arm64.bat&quot;&#xD;&#xA;  )&#xD;&#xA;  editbin /TSAWARE /STACK:8388608 &quot;$(TargetPath)&quot;&#xD;&#xA;  sn -R &quot;$(TargetPath)&quot; &quot;$(ProjectDir)..\CefSharp.snk&quot;&#xD;&#xA;)" />
```
amaitland, line 45: Is this failing? Revert it for now to see if the rest builds?


I was expecting this to look like the netcore version with the addition of strong naming.

https://github.com/cefsharp/CefSharp/blob/master/CefSharp.BrowserSubprocess/CefSharp.BrowserSubprocess.netcore.csproj#L65-L67

CefSharp/DependencyChecker.cs
```diff
@@ -135,7 +135,6 @@ public static List<string> CheckDependencies(bool checkOptional, bool packLoadin
 
             if (checkOptional)
             {
-                missingDependencies.AddRange(CheckDependencyList(nativeLibPath, CefOptionalDependencies));
```
amaitland, line 138: Is this correct?

NuGet/CefSharp.Common.nuspec
```diff
@@ -16,6 +16,8 @@
       <group targetFramework=".NETFramework4.6.2">
         <dependency id="chromiumembeddedframework.runtime.win-x86" version="[$RedistVersion$]" />
 		<dependency id="chromiumembeddedframework.runtime.win-x64" version="[$RedistVersion$]" />
+		<dependency id="chromiumembeddedframework.runtime.win-arm64" version="[$RedistVersion$]" />
+		<dependency id="System.Runtime.InteropServices.RuntimeInformation" version="4.3.0" />
       </group>
```
amaitland, line 21: We can probably add a `.net 4.7.1` target that doesn't include the `RuntimeInformation` if it's included in the framework later on.

```
<group targetFramework=".NETFramework4.7.1">
        <dependency id="chromiumembeddedframework.runtime.win-x86" version="[$RedistVersion$]" />
		<dependency id="chromiumembeddedframework.runtime.win-x64" version="[$RedistVersion$]" />
		<dependency id="chromiumembeddedframework.runtime.win-arm64" version="[$RedistVersion$]" />
      </group>
```

CefSharp.BrowserSubprocess/CefSharp.BrowserSubprocess.csproj
amaitland: I have resolved this in #5267

amaitland:
I've squashed rebased your PR in my own branch and created a new PR (#5267)

I've stripped it back to the bare minimum and it's now compiling, only an unrelated test is failing.

PR #5257, opened 2026-05-31
Core - Support Chrome command execution.

SLT-World:
**Fixes:** -

**Summary:**
- Added support for executing Chrome commands by exposing `CanExecuteChromeCommand` and `ExecuteChromeCommand` from CEF to CefSharp. However, they only function in Chrome runtime style.
https://github.com/chromiumembedded/cef/commit/53ef570f571de4b5e33c70ee7e8d4423bd66aca5
https://cef-builds.spotifycdn.com/docs/148.0/classCefBrowserHost.html#a95d2a68db5a1b3dafa5874d92fa9ff2f

**Changes:** 
   - I have modified `IBrowserHost.cs`, `WebBrowserExtensions.cs`, `CefBrowserHostWrapper.h` and `CefBrowserHostWrapper.cpp`.
      - Added `CanExecuteChromeCommand` & `ExecuteChromeCommand`. With an additional `CanExecuteChromeCommandAsync` for non-CEF UI thread execution.
      - Added `CanExecuteChromeCommandAsync` & `ExecuteChromeCommand` to `WebBrowserExtensions` so `ChromiumWebBrowser.ExecuteChromeCommand(-----)` can be directly used without `GetBrowserHost()`.
   - I have added `CefIdMappers.cs`, `CefIdMappers.h` and `CefIdMappers.cpp`.
      - Added `CefIdForCommandIdName` as a command ID mapping utility (`"IDC_FULLSCREEN"` into `34030`) from https://github.com/chromiumembedded/cef/blob/master/include/cef_id_mappers.h
      
**How Has This Been Tested?**  
Operating System: Windows 11
Environment: Visual Studio 2022
Changes function after adding an temporary button (not included in pull request):
```cs
var commandId = CefIdMappers.CefIdForCommandIdName("IDC_SAVE_PAGE");
if (await Browser.CanExecuteChromeCommandAsync(commandId)))
   Browser.ExecuteChromeCommand(commandId, WindowOpenDisposition.CurrentTab);

MessageBox.Show((await Browser.CanExecuteChromeCommandAsync(34030)).ToString());//34030 = IDC_FULLSCREEN
Browser.ExecuteChromeCommand(35004, WindowOpenDisposition.CurrentTab);//35004 = IDC_SAVE_PAGE
```
within `MainWindow.xaml.cs` in CefSharp.Wpf.HwndHost.Example.

A list of supported chrome command IDs can be seen here: https://chromium.googlesource.com/chromium/src/+/master/chrome/app/chrome_command_ids.h

After some testing, I believe this feature can be quite beneficial in the following use cases within Chrome runtime style:
- `IDC_SAVE_PAGE`, opens a save page dialog for the current site.
Previously I've only been able to accomplish this with `Browser.StartDownload(Address)`, which did not work on custom local schemes.
- `IDC_TASK_MANAGER`, opens Chrome's task manager. Similar to WebView2's [OpenTaskManagerWindow](https://learn.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/icorewebview2_6?view=webview2-1.0.3912.50#opentaskmanagerwindow) API.
- `IDC_RELOAD_BYPASSING_CACHE` & `IDC_RELOAD_CLEARING_CACHE`, essentially Chrome's hard reload & empty cache without reload.
- `IDC_QRCODE_GENERATOR`, opens a QR code popup.
There's still a handful of commands I have not tested, but some commands appeared to have no visible effect.

image

**Types of changes**
- [x] Updated documentation

**Checklist:**
- [ ] Commented my code

## Release Notes

* **New Features**
  * Added Chrome command execution capabilities for invoking native browser commands with configurable window disposition settings
  * Added synchronous and asynchronous methods to check and validate whether specific Chrome commands are available and executable
  * Added comprehensive enumeration of all supported Chrome commands with command ID lookup functionality using human-readable command names


CefSharp.Core.Runtime/Internals/CefBrowserHostWrapper.cpp
```diff
@@ -639,6 +639,54 @@ void CefBrowserHostWrapper::SetAudioMuted(bool mute)
...
+
+    if (CefCurrentlyOn(TID_UI))
+    {
+        return _browserHost->CanExecuteChromeCommand(commandId);
+    }
+
+    throw gcnew InvalidOperationException("This method can only be called directly on the CEF UI Thread. Use CanExecuteChromeCommandAsync or use Cef.UIThreadTaskFactory to marshal the call onto the CEF UI Thread.");
+}
+
+Task<bool>^ CefBrowserHostWrapper::CanExecuteChromeCommandAsync(int commandId)
```
amaitland: I wonder if we'd be better off creating this as an extension method, would be simpler and not require the private method. I think `CefCurrentlyOn` is accessible via the api.

https://github.com/cefsharp/CefSharp/blob/e4991c2947e62d0ce794ee50a8c4a5edaabf3956/CefSharp.Core.Runtime/Cef.h#L710

CefSharp.Core.Runtime/Internals/CefBrowserHostWrapper.cpp
```diff
@@ -639,6 +639,54 @@ void CefBrowserHostWrapper::SetAudioMuted(bool mute)
...
+    if (_browserHost.get())
+    {
+        int commandId = (int)commandObject;
+        return _browserHost->CanExecuteChromeCommand(commandId);
+    }
+
+    return false;
+}
+
+void CefBrowserHostWrapper::ExecuteChromeCommand(int commandId, WindowOpenDisposition disposition)
```
amaitland, line 654: Typically we'd create an enum from the command Id's to make it more user friendly, obviously that has some downsides, needs to be kept up to date, though I wonder how often commands change.

I think it's more user friendly than exposing CEF's string to int conversion method.

CefSharp.Core.Runtime/CefIdMappers.cpp
```diff
@@ -0,0 +1,29 @@
...
+#include "include\cef_id_mappers.h"
+#include <msclr/marshal_cppstd.h>
+
+using namespace msclr::interop;
+
+namespace CefSharp
+{
+    namespace Core
+    {
+        int CefIdMappers::CefIdForCommandIdName(String^ name)
```
amaitland: Historically I've exposed all of the `CEF` static methods through the `Cef` class as static methods, makes them easier to find.

This I think makes sense to follow the same pattern.

amaitland (changes requested):
Thanks for the PR!

Few comments inline, happy to discuss

CefSharp.Core.Runtime/Internals/CefBrowserHostWrapper.cpp
SLT-World: I've actually considered manually mapping the command IDs to an enum initially before discovering `cef_id_for_command_id_name`. 

However, looking at the Chromium github mirror, these IDs are updated and changed almost every week or two.
https://github.com/chromium/chromium/commits/main/chrome/app/chrome_command_ids.h
Sometimes quite drastically, as seen in this commit:
https://github.com/chromium/chromium/commit/8d255c638cbd2a1bde3e2c5f60b8c72967b7e425
Using an enum would likely require updates with every CefSharp release just to keep up, which is probably why CEF included the `cef_id_for_command_id_name`.

An enum is definitely more user-friendly, so I've implemented `ChromeCommandIds` enum as requested, but it did take quite a considerable amount of time.
Given the sheer number of commands, the time required to map them, and their frequent changes in Chromium, I do not recommend going forward with this enum.

It can be used by
```cs
(int)ChromeCommandIds.IDC_RELOAD
```

SLT-World:
Alright, I've removed the private method, moved `CefIdMappers.CefIdForCommandIdName` to `Cef.GetCefIdForCommandIdName`, and relocated all `CanExecuteChromeCommandAsync` logic into `CefSharp/WebBrowserExtensionsEx.cs`, seeing as it's the only place I can access `Cef.CurrentlyOnThread` for method extensions.

I've also implemented `ChromeCommandIds` enum, but I really do not recommend going forward with it, more details on the reply to comment.

CefSharp.Core.Runtime/Cef.h
```diff
@@ -973,6 +975,18 @@ namespace CefSharp
             {
                 return IsWindows10OrGreater();
             }
+
+            static int GetCefIdForCommandIdName(String^ name)
```
amaitland: The CEF method name isn't very friendly, maybe we rename and just include the CEF name in the comments, 

Maybe MapChromeCommandNameToId or something (open to suggestions)

CefSharp.Core.Runtime/CefIdMappers.cpp
amaitland: Do we need this now?

CefSharp.Core.Runtime.RefAssembly/CefSharp.Core.Runtime.netcore.cs
```diff
@@ -1,4 +1,4 @@
-﻿//------------------------------------------------------------------------------
+//------------------------------------------------------------------------------
```
amaitland, line 1: This is a generated file, I don't expect to see formatting changes. 

How'd this get modified?

amaitland (changes requested):

amaitland:
> I've also implemented ChromeCommandIds enum, but I really do not recommend going forward with it, more details on the reply to comment.

I cannot easily find your response on my phone, so reply here.

We should generate the fine then, short term AI should be able to generate the enum with minimal effort 


Long term we have need to generate a few enums, so just a basic generator should work. 

The devtools client API is already generated, shouldn't be hard to extend that librsy.

SLT-World:
I've removed the `CefIdMappers` files and renamed `GetCefIdForCommandIdName` to `MapChromeCommandNameToId`, I can't think of a better name myself.

> Do we need this now?

Apologies, apparently I was mistaken by Visual Studio's "Remove" option, which only removes `CefIdMappers` reference from `.vcxproj` but not the file itself.

> This is a generated file, I don't expect to see formatting changes.
> 
> How'd this get modified?


Seems to be an odd GitHub quirk? I don't see any changes on that specific line within Visual Studio.

> I cannot easily find your response on my phone, so reply here.

My response in the comment:

---
I've actually considered manually mapping the command IDs to an enum initially before discovering `cef_id_for_command_id_name`. 

However, looking at the Chromium github mirror, these IDs are updated and changed almost every week or two.
https://github.com/chromium/chromium/commits/main/chrome/app/chrome_command_ids.h
Sometimes quite drastically, as seen in this commit:
https://github.com/chromium/chromium/commit/8d255c638cbd2a1bde3e2c5f60b8c72967b7e425
Using an enum would likely require updates with every CefSharp release just to keep up, which is probably why CEF included the `cef_id_for_command_id_name`.

An enum is definitely more user-friendly, so I've implemented `ChromeCommandIds` enum as requested, but it did take quite a considerable amount of time.
Given the sheer number of commands, the time required to map them, and their frequent changes in Chromium, I do not recommend going forward with this enum.

It can be used by `(int)ChromeCommandIds.IDC_RELOAD`

---
AI does indeed solve this issue and should be able to generate the enum, however I've got no clue on that part.

PR #5260, opened 2026-06-01
fix ObjectDisposedException in synchronous js binding

skillbert:
**Fixes:**
https://github.com/cefsharp/CefSharp/discussions/5232

**Summary:**
CEF 145 introduced several issues with synchronous js bindings. This PR addresses an issue with the WCF channel used for synchronous js bindings being broken after reloading the page.
It seems like the CEF update  made it so that `CefBrowser` ids can still be reused when the underlying CefBrowser has to be recreated ([A code comment](https://github.com/chromiumembedded/cef/blob/ffa86e361d9c210136b539953663526add56191e/include/cef_render_process_handler.h#L67) states that this happens on cross-origin navigation, but it seems to happen at any navigation). When CEF recreates the `CefBrowser` like this it fires the `OnBrowserCreated` event *before* firing the corresponding `OnBrowserDestroyed` of the old browser with same id.

**Changes:**
- Dispose the old browser in `OnBrowserCreated` if the given id already exists in _browserWrappers
      
**How Has This Been Tested?**  
I didn't see a way to enable WCF in the unit tests, so it has been tested manually.
- Run `CefSharp.WinForms.Example`
- Go to button click JS binding demo page (in nav bar: script->listen for button click->go to demo page)
- Inject button js code (navbar: script->listen for button click->inject javascript code)
- Click the "Click me!" button, observe native popup
- Reload the page by clicking "Go" in address bar
- Inject js code again in the navbar
- Click the "Click me!" button again. Previously this would fail due disposed WCF channel, it now works

**Types of changes**
- [ ] Updated documentation

**Checklist:**
- [x] Commented my code

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved browser lifecycle handling when browsers are recreated with identical identifiers to avoid stale wrappers.
  * Safer removal and validation of browser instances to ensure proper cleanup and prevent resource leaks.
  * Added a helper to better detect matching browser instances, improving stability during rapid create/destroy cycles.

### DIFF
--- a/CefSharp.BrowserSubprocess.Core/CefAppUnmanagedWrapper.cpp
+++ b/CefSharp.BrowserSubprocess.Core/CefAppUnmanagedWrapper.cpp
@@ -55,6 +55,24 @@ namespace CefSharp
             return this;
         };
 
+        ref struct CefBrowserWrapperUpdater
+        {
+            CefBrowserWrapper^ _capture;
+            CefAppUnmanagedWrapper& _wrapper;
+            CefBrowserWrapperUpdater(CefBrowserWrapper^ capture, CefAppUnmanagedWrapper& wrapper) :
+                _capture(capture),
+                _wrapper(wrapper) {
+            }
+
+        public:
+            CefBrowserWrapper^ update(int key, CefBrowserWrapper^ oldValue)
+            {
+                _wrapper._onBrowserDestroyed->Invoke(oldValue);
+                delete oldValue;
+                return _capture;
+            }
+        };
+
         // CefRenderProcessHandler
         void CefAppUnmanagedWrapper::OnBrowserCreated(CefRefPtr<CefBrowser> browser, CefRefPtr<CefDictionaryValue> extraInfo)
         {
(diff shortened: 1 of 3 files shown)


amaitland:
Thanks for the PR 👍 

> It seems like the CEF update made it so that `CefBrowser` ids can still be reused when the underlying CefBrowser has to be recreated ([A code comment](https://github.com/chromiumembedded/cef/blob/ffa86e361d9c210136b539953663526add56191e/include/cef_render_process_handler.h#L67) states that this happens on cross-origin navigation, but it seems to happen at any navigation).

Please create an upstream issue for this https://github.com/chromiumembedded/cef/issues

I'd rather it was fixed `upstream` than trying to hack around it. Hacking around CEF issues is often problematic, every version might be subtly different.

If we are going ahead with this, we'll need tests.

skillbert:
I've decided to just downgrade CefSharp for my current project. I don't have enough experience with C++/CLI and the Cefsharp codebase to properly address this issue. After more testing i found that my implementation didn't fix it entirely as the old browser is still active after the second `OnBrowserCreated` call (can run an `onbeforeunload` callback)

The key issue seems to be that the current CefSharp code assumes that there can only be one browser with a given `browserId`, however in CEF there is are documented situations where there can temporarily be two browsers with the same id.
For example when you open, reload and then close the winforms example you get these hook calls:

1 start application
- OnBrowserCreated - browser pointer a with id 1

2 page reload
- OnBrowserCreated - browser pointer b, also with id 1 - two different browsers objects now exist, both with id=1
- OnBrowserDestroyed - browser pointer a, with id 1
- at this point CefSharp will dispose everything it knows about browser id 1, however browser b with id=1 still exists in CEF

3 close application
- OnBrowserDestroyed - browser pointer b with id 1 - CefSharp attempts to delete non-existent browser 1 again

Again, this seems to be intended behavior in CEF as documented [here](https://github.com/chromiumembedded/cef/blob/ffa86e361d9c210136b539953663526add56191e/include/cef_render_process_handler.h#L67). It seems like this is intended behavior to support a handover sequence so that a browser can be recreated without fully purging its container.

It seems relatively simple to fix, but someone with more knowledge of the CefSharp codebase will have to figure out the implications of this.

amaitland:
> I've decided to just downgrade CefSharp for my current project

https://github.com/cefsharp/CefSharp/releases/tag/v144.0.270

M144 has just been updated to a newer Chromium version.

> Again, this seems to be intended behavior in CEF as documented [here](https://github.com/chromiumembedded/cef/blob/ffa86e361d9c210136b539953663526add56191e/include/cef_render_process_handler.h#L67). It seems like this is intended behavior to support a handover sequence so that a browser can be recreated without fully purging its container.

Cross origin navigations should under normal circumstance be handled by a different render process. 

> * OnBrowserCreated - browser pointer b, also with id 1 - two different browsers objects now exist, both with id=1
> * OnBrowserDestroyed - browser pointer a, with id 1

This would have historically have been the old browser destroyed before the new browser created and there weren't any issues until relatively recently.

If this is happening in a different sequence now then this is something that should be raised with CEF as I don't believe this makes sense, 

What I need from a CefSharp point of view is a test that reliable reproduces the problem you are seeing.

This can likely be used as a base.
https://github.com/cefsharp/CefSharp/blob/master/CefSharp.Test/JavascriptBinding/JavascriptBindingSimpleTest.cs

Issue #2948, opened 2019-10-31
Winforms - ISelectClientCertificateCallback.Select exception


chancelyg:
**What version of the product are you using?**
built from 3700 CEF branch.
Visual Studio 2015

**What architecture x86 or x64?**
x86 build running on an x64 machine

**On what operating system?**
Win10

**Are you using WinForms, WPF or OffScreen?**
WinForms

**What steps will reproduce the problem?**
Select a client certificate and then show exception

**Please provide any additional information below.**
See document [cefsharp.github.io - IRequestHandler.OnSelectClientCertificate Method](http://cefsharp.github.io/api/57.0.0/html/M_CefSharp_IRequestHandler_OnSelectClientCertificate.htm)

```
OnSelectClientCertificate
Return Value
Type: Boolean
Return true to continue the request and call ISelectClientCertificateCallback.Select() with the selected certificate for authentication. Return false to use the default behavior where the browser selects the first certificate from the list.
```

In the CefSharp.WinForms.Example project OnSelectClientCertificate function code is like this
```c#
protected override bool OnSelectClientCertificate(IWebBrowser chromiumWebBrowser, IBrowser browser, bool isProxy, string host, int port, X509Certificate2Collection certificates, ISelectClientCertificateCallback callback)
        {
            var control = (Control)chromiumWebBrowser;

            control.InvokeOnUiThreadIfRequired(delegate ()
            {
                var selectedCertificateCollection = X509Certificate2UI.SelectFromCollection(certificates, "Certificates Dialog", "Select Certificate for authentication", X509SelectionFlag.SingleSelection);

                //X509Certificate2UI.SelectFromCollection returns a collection, we've used SingleSelection, so just take the first
                //The underlying CEF implementation only accepts a single certificate
                callback.Select(selectedCertificateCollection[0]);
            });

            return true;
        }
```
Throw C++ exception by after in call "callback.Select(selectedCertificateCollection[0]);" 

The program running normally if not call "InvokeOnUiThreadIfRequired" function and just like this
```c#
 protected override bool OnSelectClientCertificate(IWebBrowser chromiumWebBrowser, IBrowser browser, bool isProxy, string host, int port, X509Certificate2Collection certificates, ISelectClientCertificateCallback callback)
        {
            var control = (Control)chromiumWebBrowser;


[...]

PR #5281, opened 2026-08-19
Core - OnSelectClientCertificate own the copied certificate vector

jrhodnik:
**Fixes:** #2948

**Summary:**
   - `ISelectClientCertificateCallback.Select` still throws when called after `OnSelectClientCertificate` has returned, so the deferred form documented by CEF ("Return true and call `CefSelectClientCertificateCallback::Select` either in this method or at a later time") cannot be used
   - #2948 was closed on the basis that [there had been no reports since a51cdd37](https://github.com/cefsharp/CefSharp/issues/2948#issuecomment-1848819486), with an invitation to reopen if it still reproduced. It does, on v144.0.270, and `CefCertificateCallbackWrapper.h` is unchanged on master
   - The [2021 diagnosis on that issue](https://github.com/cefsharp/CefSharp/issues/2948#issuecomment-757356471) — *"It sounds like the vector is being freed. We can possibly copy the vector to resolve this (assuming the CefRefPtr instances are valid)"* — was correct. a51cdd37 makes the copy, but `CefCertificateCallbackWrapper` stores it as `const X509CertificateList&` and the copy is a local in `ClientAdapter::OnSelectClientCertificate`, so it is destroyed when the handler returns, exactly as the original was
   - Answering inside the handler works, which is likely why no further reports followed. [@thracx's workaround on that issue](https://github.com/cefsharp/CefSharp/issues/2948#issuecomment-759496971), removing `InvokeOnUiThreadIfRequired` so the certificate is selected on the handler's own thread, is the same observation from the other direction

   - I have modified `CefCertificateCallbackWrapper`
      - Owns a heap copy of the certificate list instead of holding `const CefRequestHandler::X509CertificateList&`
      - Allocated in the constructor initializer and deleted in the finalizer, following the ownership pattern already used by `BrowserSettings`, `CefSettingsBase`, `RequestContextSettings` and `WindowInfo`
      - `Select` iterates via the pointer
   - Copying the vector copies the reference-counted `CefX509Certificate` pointers, which is what keeps the certificates alive. A ref class cannot hold a `std::vector` (or a smart pointer) by value, which is why the field was a reference to begin with
   - No signature change, and no behaviour change for callers that answer inside the handler

**How Has This Been Tested?**
Built the NETCore packages for x86 and x64 from the v144.0.270 tag with this change, on VS2022 / Windows 10, and loaded servers that require a client certificate and servers that merely request one.

Deferred path, the failing case: held the callback, returned `true` from `OnSelectClientCertificate`, then called `Select` afterwards from another thread.
   - Before: the process died inside the thumbprint comparison loop.
   - After: the chosen certificate is presented on the still-suspended request and the server accepts it.

Also re-checked the two paths that already worked, both unchanged: calling `Select` inside the handler, and calling `Select(nullptr)` to continue without a certificate.

**Types of changes**
- [ ] Updated documentation

**Checklist:**
- [x] Commented my code

One point worth a maintainer's opinion: with deferral working, `Select` becomes reachable from a thread other than the CEF UI thread. The header's "or at a later time" wording reads as intended, and it works in practice, but if `Select` is meant to be UI-thread-only then that is worth documenting.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved certificate selection reliability by preserving the available certificate list throughout the selection process.
  * Prevented potential issues caused by accessing certificate data after its original source was released.
  * Ensured certificate options remain available while the selection callback completes.

### DIFF
--- a/CefSharp.Core.Runtime/Internals/CefCertificateCallbackWrapper.h
+++ b/CefSharp.Core.Runtime/Internals/CefCertificateCallbackWrapper.h
@@ -19,11 +19,19 @@ namespace CefSharp
         {
         private:
             MCefRefPtr<CefSelectClientCertificateCallback> _callback;
-            const CefRequestHandler::X509CertificateList& _certificateList;
+            // Owned copy of the certificates Chromium offered, not a reference to the caller's.
+            // ClientAdapter::OnSelectClientCertificate builds that list as a stack local, so a
+            // reference to it dangles the moment the handler returns. CEF permits calling Select
+            // "either in this method or at a later time", so a wrapper that outlives the handler
+            // has to own the list it selects from, or a deferred Select reads freed memory.
+            // A ref class cannot hold a std::vector by value, hence the pointer. Copying the
+            // vector copies the reference-counted CefX509Certificate pointers, and those
+            // references are what keep the certificates themselves alive.
+            CefRequestHandler::X509CertificateList* _certificateList;
 
         public:
             CefCertificateCallbackWrapper(CefRefPtr<CefSelectClientCertificateCallback>& callback, const CefRequestHandler::X509CertificateList& certificates)
-                : _callback(callback), _certificateList(certificates)
+                : _callback(callback), _certificateList(new CefRequestHandler::X509CertificateList(certificates))
             {
 
             }
@@ -31,6 +39,9 @@ namespace CefSharp
             !CefCertificateCallbackWrapper()
             {
                 _callback = nullptr;
+
+                delete _certificateList;
+                _certificateList = nullptr;
             }
 
             ~CefCertificateCallbackWrapper()
@@ -53,8 +64,8 @@ namespace CefSharp
                     auto certThumbprint = cert->Thumbprint;
 
                     std::vector<CefRefPtr<CefX509Certificate>>::const_iterator it =
-                        _certificateList.begin();
-                    for (; it != _certificateList.end(); ++it)
+                        _certificateList->begin();
+                    for (; it != _certificateList->end(); ++it)
                     {
                         auto bytes((*it)->GetDEREncoded());
                         auto byteSize = bytes->GetSize();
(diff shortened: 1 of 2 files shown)


CefSharp.Core.Runtime/Internals/CefCertificateCallbackWrapper.h
```diff
@@ -19,18 +19,29 @@ namespace CefSharp
         {
         private:
             MCefRefPtr<CefSelectClientCertificateCallback> _callback;
-            const CefRequestHandler::X509CertificateList& _certificateList;
+            // Owned copy of the certificates Chromium offered, not a reference to the caller's.
+            // ClientAdapter::OnSelectClientCertificate builds that list as a stack local, so a
```
amaitland: On the surface fixing that instead seems like a better long term solution. 

At a minimum we should remove the extra copy

Thoughts?

amaitland (changes requested):
Thanks for the PR, see comment lineline.

CefSharp.Core.Runtime/Internals/CefCertificateCallbackWrapper.h
amaitland: Had a slightly more detailed look at the existing code, I think having the callback wrapper create the `CefRequestHandler::X509CertificateList` that it in tern owns seems reasonable, so we just need to cleanup `ClientAdapter::OnSelectClientCertificate`

jrhodnik: Makes sense — done. `ClientAdapter::OnSelectClientCertificate` no longer builds its own copy, it just passes `certificates` straight through, and the wrapper owns the copy as before.

Also updated the comment on `_certificateList`, since it described the caller's stack local that this removes.

amaitland: Perfect, thank you!

